PR #2, opened 2023-12-03
Re-Adding Dawn Wallet

jaymeh13:
Woops on the previous commit. Here's Dawn Wallet!


mozrt2:
@jaymeh13 - thanks for this, I reverted the commit after merging it initially because it looks like some data is inaccurate (e.g. the type being Safe, availability in browser, etc.) - have added Dawn to the list of wallets to check and add soon though :)

PR #4, opened 2023-12-04
Add Obvious wallet to the list

jebu:
changes to add Obvious wallet


mozrt2:
Merged, thanks for the contribution @jebu! I currently can only select one type of account in the UI so I picked 4337 as I believe this is the main type for Obvious

PR #11, opened 2023-12-11
Math wallet

iSpeakNerd:
## Math Wallet 
- [ ] website https://mathwallet.org/en-us/
- [ ] github https://github.com/mathwallet

## Trust Wallet
- [ ] website https://trustwallet.com/
- [ ] github https://github.com/trustwallet


mozrt2:
Thanks @iSpeakNerd - just merged. Can you confirm what is meant by "Add auto-lock security toggle"? Do you mean locking the first column of the table so it stays in view when swiping through the data? If that's it, I agree this should be added :)

iSpeakNerd:
That does sound nice, but no. I meant QoL security features like auto
locking the mobile app after an amount of time

mozrt2:
Ah I see, I wonder how easy it would be to populate the data, we'd first need a specific measure for this, e.g. "autolocks after at most 30sec of inaction" or "autolocks when a user switches to another app", etc. Might be difficult to find one measure that covers the different implementations of such a feature

iSpeakNerd:
```
{
"Does autolock": "y/n" 
"Time to autolock": "customizable/1 hour/1 min"
}
```

PR #16, opened 2023-12-12
Add Den Multisig

jierlich:
Added data for the Den multisig wallet. Let me know if any questions!

Some fields were ambiguous so I made a best effort. A description and example for each field in the future would be helpful to ensure contributors to the repo are on the same page.


mozrt2:
thanks for the contribution @jierlich ! The definitions can be found on walletbeat.fyi below the main table, I've now also linked this in this repo's readme, thanks for the heads up on this

PR #18, opened 2023-12-19
Add Gemwallet

0xh3rman:
We have plan to open source the app and list on [F-Droid](https://f-droid.org/en/), here is the open source core library: https://github.com/gemwalletcom/core.


0xh3rman:
@mozrt2 could you please share the error msg? not sure what went wrong with the deployment

mozrt2:
just checked and merged - thanks for the contribution @0xh3rman !

PR #25, opened 2024-11-26
Update MetaMask Entry

danfinlay:
Previous entry seemed to be a hastily cloned Mathwallet entry.

Quite a few of our more recent features were missed:
- Transaction scanning
- Our snaps extensibility system (which enables custom chains, name resolvers, and account backup systems including MPC)

Some of our very long term features were mis-represented:
- Mobile's in-app browser and provider injection.

We may have to come back and update some ENS entries again.


Montoya (approved):

danfinlay:
I can't see why Vercel failed. Opening for review since I got a sign-off from another member of my team.

src/data/wallets/mathwallet.ts
Hugoo, line 1: As already flagged, this PR is editing the wrong file: "mathwallet.ts" but the content suggests "MetaMask".

Hugoo (changes requested):

Issue #14, opened 2023-12-12
Proposal to change the data structure of supported chains and `hardwareWalletSupport`


kien-ngo:
I think they should be an array to make it extendable and more data can be provided that way.

For example, instead of:
```
hardwareWalletSupport: true,
```

We can have:
```
hardwareWalletSupport: ['ledger', 'keystone'],

// or

hardwareWalletSupport: [], // if empty then -> FALSE
```

Same goes for the list of supported chains, we can pull all the latest chains from the chainlist repo and do something like this:
```
supportedChains: ['ethereum', 'avalanche', ...]
```

mozrt2:
That's a good point, I will try to adjust this in the next release and this will help to also gradually show more information in the frontend. Thanks for the suggestion @kien-ngo !

Shakacro:
Guys TY very much... :) beer for all ;)


![shaka](https://github.com/user-attachments/assets/c39bce1a-7db9-47a4-8a0a-16c29d3153c3)

Montoya:
Hardware wallet support can get complicated beyond that, depending on whether it's USB, QR code, Bluetooth... on mobile you can use USB with Android but iOS can only do Bluetooth / QR codes, so the data model can get very complex :)

Issue #12, opened 2023-12-12
`accountType` should be an array because some wallets support multiple types of addresses. 


chendatony31:
TokenPocket mobile wallet supports `EOA`, `Safe`, and `4337`.

mozrt2:
Thanks for bringing this up @chendatony31 - will adjust this in the next release!

PR #36, opened 2024-12-04
Proposal for new wallet data model

polymutex:
This PR is a proposal for a richer data model for wallets. It is incomplete, and shown here for feedback purposes only.

The goals of this data model are as follows:
- Support arbitrary complexity of what we know about wallets, e.g. to support requested features like #12 and #14.
- Keep the main comparison table view meaningful and comparable despite this increased complexity.
- Deduplicate wallet feature data for wallets which don't differ across platforms (or which only support one platform).

The main idea to solve this problem is to decouple two things that are currently coupled in the current model: (a) what we know about wallets ("features"), and (b) how well wallets fare on certain criteria ("attributes"). In more detail:

* "**Features**" is a detailed breakdown of the particulars of a wallet. Examples in this PR include the license under which the wallet is released (as an example of a simple feature), and some personal data leakage information (as an example of a more complex feature).
* "**Attributes**" are a summarized view of wallet features. Examples in this PR include:
  - Whether the license of the wallet is Free and Open Source Software.
  - Whether the source code of the wallet is visible to the public.
  - Whether the wallet allows a third-party to learn associations between a user's Ethereum address and other personal information.

Note that these attributes all share the properties that (a) they are meaningful from a "desirable feature" perspective, and (b) they are summarizable as an almost-boolean choice (in practice it is a 4-state value: `No`, `Partial`, `Yes`, and `Unrated`). This is what makes it useful to have on the comparison table.

This PR contains an example wallet entry (`src/data/wallets/not-a-real-wallet.ts`) to show what a hypothetical wallet would look like under this data model. This corresponds to a `Wallet` object, for which the meat is in the `features` field called `WalletFeatures`. `WalletFeatures` is an arbitrary tree of values, which at each node supports either branching out per platform or not. The definitions for all the types involved are under `src/schema/wallet.ts`, `src/schema/variants.ts`, and `src/schema/features.ts`. Each feature has a separate file, e.g. `src/schema/features/license.ts` for the license and `src/schema/features/privacy/data-collection.ts` for the data collection. This matches the structure of the `WalletFeatures` tree.

Meanwhile, attributes have their own separate tree. The basic types are defined in `src/schema/attributes.ts`, and (similar to features), each attribute has its own file in a tree structure. The PR contains `src/schema/attributes/transparency/open-source.ts`, `src/schema/attributes/transparency/source-visibility.ts`, and  `src/schema/attributes/privacy/address-correlation.ts` as example attributes.
I haven't created the attribute tree type but it would be similar to the `ResolvedFeatures` tree.

In order to determine the attribute tree for a set of `WalletFeatures`, each attribute is evaluated on each `Variant` of the `WalletFeatures`. For each node of this tree, if the results (the `Evaluation`s) agree, then that is used as the node in the attribute tree. If they disagree, then the lowest-rating `Evaluation` is picked.

The intent of the attribute tree is to be used as an input to the comparison table. Each attribute corresponds to a slice of the pie charts that exist under each category. So for example, if there is a "transparency" category and a "privacy" category, each of them would have a pie chart with as many slices as there are attributes underneath that category, and each of the slices corresponds to one attribute and is colored according to its rating (`No` is red, `Partial` is orange/yellow, `Yes` is green, `Unrated` is greyed out). Hovering over it reveals the `display_name` of the attribute and its evaluation so that the user understands what is being evaluated (and so that per-platform differences on that attribute can also be explained). Clicking on this slice would go to a per-wallet page that explains the score of the wallet on this attribute. This same page can explain the attribute in more details, show the wallet features that were involved in determining this, and show other wallets that fare better on this attribute.

Let me know what you think!


mozrt2:
Hi @polymutex - I finally had time to look through this - first off, thanks for the thoughtful & detailed work here, I really like your approach.

Overall, I think the model makes sense.

How are you thinking about data collection for this new model? It looks like there will be much more data for each wallet. Maybe it makes sense to have a section on the website where submitters can add data and it gets auto-formatted into a ts file?

The list of criteria to consider for a first version of this new model will also be interesting to discuss. I think it will be important to find a good balance between comprehensiveness and ease of submitting data, so sufficient wallets are represented from the beginning.

It might also make sense to add some quantitative data, e.g. about usage, although this data might be tricky to collect. Defillama has some interesting stats, e.g. https://defillama.com/protocol/metamask?fees=true&devMetrics=true&devCommits=true&twitter=false -but maybe that's something to look at further down the line.

mozrt2:
Thanks for the response! Love the approach, so no additional questions from my side :)

Regarding the collection of live data vs having walletbeat on IPFS: I think we could still do this by having most of the data and frontend be static on IPFS while the live parts get loaded via API calls. So in the worst case scenario of APIs breaking, the live data would not be available but everything else would remain functional.

mozrt2:
hi @polymutex - thanks for the PR, just had a look

It overall looks really clean and easy to understand from a code perspective, so I merged it already and you should be able to find it at walletbeat.fyi/beta

Two quick comments about the content (I know you are actively shaping this so this might be planned already):

- Definitions & methodology: it'd be great to have a clear outline of the definitions and methodology for each aspect reviewed at the bottom of the page (or another visible area, like on hover) 

- This will also make it easier to give input on each of the reviewed aspects - for example at a first glance, it feels that privacy should cover a few more aspects: Daimo for example assigns a public name to every user, which makes it very easy to track down everybody's payments and connections (see [this Dune dashboard](https://dune.com/queries/3186413)), which I'd consider a much bigger privacy issue than IP tracking

PR #52, opened 2024-12-20
1️⃣: Implement skeleton for per-wallet page.

polymutex:
The sections' contents are stubbed out for now but the structure is in place.

The intention is to have a more complete explanation of the wallet ratings, why the attributes are what they are and why they matter, as well as a full view of the wallet information on record.

This is PR 1️⃣.


mozrt2:
hey @polymutex - this is the error Vercel is throwing:

```Type error: Type '{ params: WalletUrlParams; }' does not satisfy the constraint 'PageProps'.
  Types of property 'params' are incompatible.
    Type 'WalletUrlParams' is missing the following properties from type 'Promise<any>': then, catch, finally, [Symbol.toStringTag]

PR #99, opened 2025-02-01
Remove all dependencies on UI components from `data` and `schema`.

polymutex:
This reworks the way the `text.tsx` library (which is now `content.ts`, notice the change of extension from `.tsx` to `.ts`!) works such that it only produces plain JavaScript `object`s (type: `Content`), which the newly-introduced `RenderContent` and `RenderTypographicContent` UI components have the logic to render.

Running `grep -ri components src/beta/data src/beta/schema` now returns empty, so the `src/beta/data` and `src/beta/schema` directories no longer have dependencies on UI code.

This provides a stronger separation of concerns between content and UI.

cc @darrylyeo


darrylyeo (approved):
Nice one!

PR #119, opened 2025-03-11
Obvious wallet seems to be a scam

ai-slave:
Obvious wallet seems to be a scam
https://apps.apple.com/us/app/obvious-smart-wallet/id1643088398?see-all=reviews


mozrt2:
thank you - as it is unclear if it is a scam but want to be cautious, I have removed Obvious from displaying in the frontend while keeping the data in the repo

PR #120, opened 2025-03-13
Create Ctrl

sean-mc-grath:
Add Ctrl Wallet


polymutex:
Hi there. You also need to add the wallet to `src/data/data.ts`, otherwise it won't show up. Can you add it?

PR #124, opened 2025-03-18
Update Enkrypt data

Abyscuit:
Update Enkrypt data to include Autoswitch and L2 ENS compatibility


polymutex:
I have received permission to merge this into `main` from @mozrt2, so merging this now.

PR #129, opened 2025-04-08
Add implmenetaion for account portability evaluation of EIP-7702 wallets

JIOjosBG:
Since EIP-7702 accounts are just EOA account with extra functionality, the criteria for evaluating the portability of such wallets is virtually the same as EOA wallets. 

For that reason the implementation for EOAs is used for 7702

### DIFF
--- a/src/schema/attributes/self-sovereignty/account-portability.ts
+++ b/src/schema/attributes/self-sovereignty/account-portability.ts
@@ -590,7 +590,7 @@ function evaluateMultifactor(
 
 function evaluateEip7702(
 	accountSupport: AccountSupport,
-	_: ReferenceArray,
+	references: ReferenceArray,
 ): Evaluation<AccountPortabilityValue> {
 	if (!isSupported<AccountType7702>(accountSupport.eip7702)) {
 		throw new Error('EIP-7702 account type is not supported')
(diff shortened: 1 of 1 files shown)


JIOjosBG:
I don't think i understand what you mean.

> wallet A could use EIP-7702 features to lock up funds in such a way that some wallet B importing the same underlying EOA may not be able to unlock.

if a wallet B has the private key or seed for the EOA, then it can broadcast a normal EOA txns to move the funds, regardless if there is an 7702 authorization in place. Also it can change the authorization or simply remove it.  This is not the same as 4337 accounts, where the authorization logic is all it is to the account.

> per-app account derivation

do you refer to normal account derivation here? If yes, then this logic is already implemented in `evaluateEoa`

@polymutex

JIOjosBG:
To your first point. This is equivalent to prompting the user to stake or lock up their assets in a protocol, developed by the wallet provider. Imo this is not a characteristic of a wallet and should not fall into the scope of walletbeat. I do not consider this self-custody. 

To your second point. I actually completely missed something that can be added as extra property to the 7702 wallet variant. There is a [very interesting way to have 7702 wallet](https://blog.biconomy.io/prep-deep-dive/) with different security model, discovered by Biconomy research. The security model looks more like old Smart contract wallets and 4337 than EOA or 7702. 

TL;DR:
You can forge a signature, that makes a 7702 authorization to a smart account implementation and allow only your EOA to be a signer to that account. That authorization is valid only for the account that is derived from the signature. The difference between that and normal 7702 is that:
- you cannot drain the account with its base private key, as such does not exist (exactly like 4337 and legacy smart accounts)
- you can change the authorized EOA (exactly like 4337 and legacy smart account, unlike generic 7702 where the base pk is authorized)
- you basically create a new account (like 4337 and legacy SA)

i am not aware of any legit live implementations of consumer wallets

PR #136, opened 2025-05-10
Enable subresource integrity (SRI)

polymutex:
This adds sha256 integrity hashes to all scripts and stylesheets. This ensures that if a webpage's HTML code can be served with integrity, then the external resources it depends on also are.

As part of this change, delete Plausible analytics, since this is an external resource.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -54,6 +54,7 @@
 		"jiffyscan",
 		"jiojosbg",
 		"KEEPKEY",
+		"kindspells",
 		"Kleptography",
 		"LDJSON",
 		"lifi",
@@ -131,6 +132,7 @@
 		"public/**/*.svg",
 		"node_modules/**",
 		"src/legacy/**",
+		"src/generated/**",
 		"pnpm-lock.yaml"
 	]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Build output
 dist/
+src/generated/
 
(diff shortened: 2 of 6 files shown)


darrylyeo:
This [Astro-Shield](https://github.com/KindSpells/astro-shield) plugin is pretty cool. Though currently if I run the `pnpm build` script multiple times it still outputs different random-generated suffixes for the .js files and outputs this warning:

```
09:06:10 [WARN] [@kindspells/astro-shield] SRI hashes have changed for static resources that may be used in dynamic pages. You should run the build step again
```

Maybe some operations are out of order in the build pipeline? Will take a closer look.

darrylyeo:
Looks like this non-determinism actually stems from the JavaScript build pipeline outputting different variable names on each run. This cascades down to the hashes of the static .js files, and in turn the hashes of the static .html files importing them.

Screenshot 2025-05-11 at 09 40 54

Screenshot 2025-05-11 at 09 41 06

darrylyeo:
If we're not too worried about cache-busting we can try removing the hashes from file names altogether as a workaround for the un-deterministic JavaScript output:

```ts
	vite: {
		build: {
			rollupOptions: {
				output: {
					assetFileNames: 'assets/[name][extname]',
				},
			},
		},
	},
```

Not a show-stopper for enabling SRI though – that seems to be working great!

darrylyeo (approved):

PR #154, opened 2025-06-04
Sign automated commits made from CI workflow.

polymutex:
This cryptographically signs commits made from the Helios checkpoint workflow.

This PR is on top of #153. This is such that they can both be merged together without breaking my commit signature. To preview the diff, please only look at the latest commit in the branch.

PR #155, opened 2025-06-04
Avoid redirect on home page load.

polymutex:
This changes the `/` page to be a duplicate of the `/wallet/summary` page. However, on load, it instantly rewrites its URL to `/wallet/summary` (without causing a reload), so that is is functionally equivalent to a redirect.

This has the advantage of:
- Ensuring that linking to the website's homepage doesn't break OpenGraph data, as a JS-based redirect does
- Avoid a flash of white from the unstyled JS redirect page during its loading.
- Being much faster to load the homepage (even faster than an HTTP-protocol-level redirect), since the page's data is sent as part of the server's first response, whereas any form of redirect requires at least one more roundtrip.

This PR is on top of #154. This is such that they can both be merged together without breaking my commit signature. To preview the diff, please only look at the latest commit in the branch.

### DIFF
--- a/src/layouts/Layout.astro
+++ b/src/layouts/Layout.astro
@@ -8,9 +8,14 @@ import { ClientRouter } from 'astro:transitions';
 
 type Props = {
   metadata: ComponentProps<typeof Metadata>['metadata'];
+
+  /**
+   * If specified, rewrite URL to this as the page loads.
+   */
+  rewriteUrlTo?: string;
 };
 
-const { metadata } = Astro.props;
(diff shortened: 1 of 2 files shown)


darrylyeo:
Might want to consider Astro's built-in [redirect](https://docs.astro.build/en/guides/routing/#redirects) or [rewrite](https://docs.astro.build/en/guides/routing/#rewrites) feature instead.

polymutex:
These won't work for IPFS, since there's no Astro server in such a context. This solution avoids the network roundtrip and works on IPFS.

PR #158, opened 2025-06-05
BiomeJS

darrylyeo:
* Reduce tooling dependencies by replacing ESLint and Prettier with [Biome](https://biomejs.dev/)
* Smarter import sorting that is "group"-aware (preserves double line breaks) and places types first
* Catch some additional issues in CSS and React

Known caveat:
* Biome doesn't currently have a built-in way to flag unsafe TypeScript `as` clauses (i.e. the equivalent of `@typescript-eslint/no-unsafe-type-assertion`)


polymutex:
Well I like the extra speed, but not the loss of type safety... Can we keep ESLint, but only enforce it in CI and in `pnpm run check:all`?

Also, do Biome's annotations still show up in VSCode? `.vscode/extensions.json` doesn't seem to have an entry, did you mean to add `biomejs.biome`?

polymutex:
cc @lucemans, who mentioned that the spread of top-level files we have is unwieldy (which I agree with). This change gets rid of `.prettierrc`, `.editorconfig`, `eslint.config.mjs`, and replaces them with a single `biome.jsonc`. Seems like a win in terms of tooling simplicity and of top-level file count.

That said, I am very adamant that we keep enforcing `@typescript-eslint/no-unsafe-type-assertion`. But it may be OK to hide the config to enforce this out of the top-level directory and shove it inside `deploy` or something, where only CI would look at it. That way it wouldn't clutter up the top level. And it would be OK to use a non-default path for that, because it's not meant to be used by individual developers themselves. Just by CI, which can bear to have its own non-default configuration.

polymutex:
Closing as stale.

PR #159, opened 2025-06-05
Restore original formatting

darrylyeo:
Restore formatting preferences from before the mass mixed-formatting change introduced in https://github.com/walletbeat/walletbeat/commit/9ab215cebce440e137a98a6db290632648b1cf89.

* Tabs
* Semicolons only as needed
* Manually fix mixed indentation in markdown files, shell scripts and multi-line template strings


polymutex:
**To all: Please see discussion on PR #189 about merging this**. I don't have a strong opinion about the formatting we use, which is why I didn't push back on this initial style nor on the later change to spaces+semicolons. But I do have the strong opinion that we should _not_ keep swinging back and forth on this, as it is highly disruptive to everyone, and I hope our project can grow past the tabs-vs-spaces debate phase.

(Though if I may opine for a second on that debate: I do think the current state is not great, because of the mixed use of tabs and spaces; there's still a lot of tabs remaining in template strings, and it's easier to change the code back to tabs than it is to finish converting all the template strings to spaces. But that's a weakly held opinion.)

For this reason, I am supportive of this change, but **only if merged at a time when doing so is not disruptive to others**, and on the condition that the responsibility for finding the right time to merge falls on PR author @darrylyeo. This means ensuring there is no large ongoing work in-flight by getting sign-off from @lucemans, @Hugoo, and @PatrickAlphaC, with let's say a 4-day timeout at which point we can claim action bias such that this PR may proceed without waiting for explicit sign-off.

You can also treat this message as a sign-off from myself, as I do not mind rebasing formatting changes for my own in-flight work.

(As an additional precondition for merging this PR, CI tests should pass. Currently they are still complaining about semicolons.)

polymutex (approved):

polymutex:
(I am approving this PR as a way to explicitly signal sign-off from my end. This _technically_ gives you the ability to merge this PR into `beta`, but I trust you to not actually do so until the above conditions are met.)

polymutex:
It has now been a full week without objections. I think you can go ahead with this.

polymutex:
Sent follow-up fixes in #209

PR #161, opened 2025-06-06
Revamped signing criteria, and added cypherock and onekey

PatrickAlphaC:
I updated the transaction signing and message signing features to be explicitly different, focusing on how a user can verify/extract the calldata on a transaction/message.

I added decoding calldata as a feature as well, although, this could be subjective. I could make it more objective by saying like:

- decodes aave
- decodes safe
- decodes multi-send (uses packed encoding, much trickier to decode)

To make it more objective. But I think for now this is probably fine...

What do you think? 

I did most of this on-stream if you want to see the thought process: https://www.youtube.com/watch?v=VRsGumn_q8w


CLAUDE.md
```diff
@@ -0,0 +1,88 @@
...
+- Avoid MUI component modifications (project is migrating away from MUI)
+- Prefer Tailwind classes over MUI components
+- Use CSS variables from `global.css` for theming
+
+## Adding New Wallets
+
+1. Copy appropriate `unrated.tmpl.ts` from `data/[wallet-type]/`
+2. Rename to `[wallet-name].ts` and populate with wallet data
+3. Import in corresponding `data/[wallet-type].ts` file
+4. Submit pull request
```
polymutex, line 80: Not sure this section on adding new wallets is useful here; Claude can't fill in the wallet data on its own, as it requires manual research work. But maybe this section could be repurposed to describe how the wallet data is laid out in the `data` directory.

public/images/wallets/cypherock.svg
polymutex, line 1: Can you crop this file so that it has no blank edges around it?
Otherwise the actual icon part of this image will look very small when shrunk down to icon size.

public/images/wallets/onekey.svg
polymutex, line 1: Same request for cropping this file to remove blank edges.

src/schema/attributes/security/firmware.ts
```diff
@@ -26,6 +26,12 @@ function evaluateFirmware(features: FirmwareSupport): Rating {
     features.reproducibleBuilds,
     features.customFirmware,
   ];
+
+  // If any rating is null (unreviewed), return UNRATED
+  if (ratings.some(r => r === null)) {
+    return Rating.UNRATED;
```
polymutex, line 32: Can you add some handling for this case in the `evaluate` function?
Otherwise it will currently say `{{WALLET_NAME}} has unrated firmware.` which sounds a bit weird. I think `evaluate` should call the `unrated` helper function (from `'../common'`, already imported here) if `evaluateFirmware` returns `Rating.UNRATED`.

src/schema/features/security/hardware-wallet-dapp-signing.ts
```diff
@@ -1,27 +1,116 @@
...
+ * - from
+ * - to
+ * - chain
+ * - value
  */
-export enum DappSigningLevel {
+export enum ShowsTransactionDetails {
+  NONE = 'NONE',
+  SOME = 'SOME', // Some of the stuff above
+  FULL = 'FULL', // All of the stuff above
```
polymutex: I suggest creating a

```
interface DisplayedTransactionDetails {
  gas: boolean;
  nonce: boolean;
  from: boolean;
  ...
}
```

And then you can define simple helper `consts` like

```
const displaysAllTransactionDetails: DisplayedTransactionDetails = {
  gas: true,
  nonce: true,
  ...
}
const displaysNoTransactionDetails: DisplayedTransactionDetails = {
  gas: false,
  nonce: false,
  ...
}
```

This way, wallet feature data is still objective and complete, but wallets that display all or none of these details still have a short way to express that.

src/schema/features/security/hardware-wallet-dapp-signing.ts
```diff
@@ -1,27 +1,116 @@
...
+}
+
+/**
+ * Calldata decoding
+ *
+ * Does a wallet decode calldata natively?
+ *
+ * Note: There are "packed" operations that get a pass for now.
+ * However, in the future, packed operations like MultiSend, will be required for
+ * the "NESTED" rating.
```
polymutex: This type of decision belongs into attribute evaluation logic, not feature data.
Similar to my other suggestion, I suggest something like

```
type CalldataDecoding = Support<{
  packed: boolean
  eip712: boolean
}>
```

... or whatever type makes the most sense to represent the levels of support that a wallet has. (Check other feature data that uses the `Support` type, it's used as a wrapper for functionality that wallets may or may not support, and its type parameter contains data that's only required _if_ the wallet supports the feature.)

Then, the logic to turn this into a 3-state rating-ish value (`NONE` / `BASIC` / `NESTED`) would be done at attribute evaluation time.

src/schema/attributes/security/hardware-wallet-dapp-signing.ts
polymutex, line 1: These changes on this file look great!

polymutex (changes requested):
Epic PR. I'm curious if you encountered difficulties working on this codebase. Guess I can watch the stream to find out :)

CLAUDE.md
PatrickAlphaC: I like it tbh, if I give an AI all the wallet data, and say "Add a new wallet, here is all the data" - it can do this!

public/images/wallets/cypherock.svg
PatrickAlphaC: https://x.com/PatrickAlphaC/status/1932208463074976036

src/schema/features/security/hardware-wallet-dapp-signing.ts
PatrickAlphaC: Ok! Give me a bit to fix this...

src/schema/features/security/hardware-wallet-dapp-signing.ts
PatrickAlphaC: Gotcha... ok ok. Let me make this much more objective then... I'll add some contracts people can call/use to test...

PatrickAlphaC:
It took me a little to figure out the codebase, claude helped of course! But once I got the hang of it, I think I can make more PRs easier moving forward!

The comments are helpful! I'll fix those changes this week! Sorry on my slow speed. Trying to contribute in my spare time!

public/images/wallets/cypherock.svg
polymutex: Thanks for asking. Note that PNG is fine too, so long as it's cropped the same way.

polymutex:
> It took me a little to figure out the codebase, claude helped of course! But once I got the hang of it, I think I can make more PRs easier moving forward!
> 
> The comments are helpful!

Good to hear. I think the way to summarize them is: wallet feature data should be purely objective "Does the wallet do feature X or Y"-type information, whereas anything that involves a judgement call as to where to draw the line between what's good/bad goes in attribute evaluation logic. Enums that represent different "levels" of implementation quality inherently involve such judgement calls to decide where the boundary lies, so such logic goes in attributes.

I added some description to that effect in 0d39f07e11638b9792cf9e5760bc0c3faf8c940b, make sure they are in context and hopefully that can help.

> I'll fix those changes this week! Sorry on my slow speed. Trying to contribute in my spare time!

No worries, Walletbeat is still a spare-time project for all of us at the moment too :)

PatrickAlphaC:
@polymutex ready for a new review!

I think it's VERY objective now :D 

My only question... would this make sense to split out as it's own attribute instead of bundling it into security?

Maybe a new attribute: `transaction clarity`. And it could replace:
- Self-sovereignty
- maintenance
- risk by device

(idk, I could be wrong, idk what those are going to be)

polymutex:
> My only question... would this make sense to split out as it's own attribute instead of bundling it into security?
> 
> Maybe a new attribute: `transaction clarity`. And it could replace:
> 
> * Self-sovereignty
> * maintenance
> * risk by device
> 
> (idk, I could be wrong, idk what those are going to be)

I think there may be a bit of terminology confusion here. "Security", "self-sovereignty", and "Maintenance" are attribute *groups*, not attributes.
- An *attribute group* is a pie chart, and corresponds to a folder under `src/schema/attributes`.
  - Example: `src/schema/attributes/security`, which is the "Security" attribute group, and contains various security-related attributes.
- An *attribute* is a single slice of the pie, and corresponds to a single `.ts` file under `src/schema/attributes/${some_attribute_group}`.
  - Example: `src/schema/attributes/security/security-audits.ts`, which looks at a wallet's security audit history specifically.
- The summary table has one column per *attribute group*, which displays that attribute group's pie chart. Each slice of this pie chart corresponds to a single *attribute*.

I think this makes more sense if you look at the software wallet table, as it is more filled out. Click the pie charts on Ambire for example and it should be clear how these things map.

On the hardware wallet side, these are less well-defined, because we've historically more focused on the software side and the hardware side came later. So the current attribute groups are copy/pasted and mostly unrated right now, but in the long run we should aim to tighten them into a coherent hierarchy that makes sense for hardware wallets.

Each *attribute* is evaluated for all wallet types (i.e. software wallet and hardware wallet alike). However, some attributes don't make sense to apply to software wallets, and some attributes don't make sense to apply to hardware wallets. Such attributes can return the `EXEMPT` rating, which makes them hidden from the pie slices. For example, look at `src/schema/attributes/security/chain-verification.ts`, you'll see that it does not apply to hardware wallets:

```typescript
  evaluate: (features: ResolvedFeatures): Evaluation<ChainVerificationValue> => {
    if (features.variant === Variant.HARDWARE) {
      return exempt(
        chainVerification,
        sentence('This attribute is not applicable for hardware wallets.'),
        brand,
        null,
      );
    }
    // ... rest of the code goes here for rating software wallets...
  },
```

This way, the "Chain verification" attribute, which is part of the "Security" attribute group, will show up in the software wallet table, but will not show up for the hardware wallet table. Each of these tables is created dynamically based on whether any of the wallets in that table is non-`EXEMPT`. And if all attributes in an attribute group are `EXEMPT` for all wallets, then that attribute group will automatically disappear from the table (and the column will be gone).

We only recently added hardware wallets, and we didn't go through all the attributes to see if they made sense to also apply to hardware wallets. So if you'd like, you could do a pass and add code like the Chain Verification code above to return `EXEMPT` for hardware wallets, and that'll shrink the table to only the attributes that make sense for hardware wallets. But that's for another PR.

For this PR, I think let's go ahead and merge this. We can later refactor the attribute logic so that there is a new "Transaction clarity" attribute. I think that would make total sense to have as part of the "Security" attribute group.

Hope this makes sense!

src/schema/features/security/hardware-wallet-dapp-signing.ts
```diff
@@ -1,27 +1,139 @@
...
+ *
+ */
```
polymutex, line 55: I love the objectivity of this check :)

polymutex (approved):

PatrickAlphaC:
> > My only question... would this make sense to split out as it's own attribute instead of bundling it into security?
> > Maybe a new attribute: `transaction clarity`. And it could replace:
> > 
> > * Self-sovereignty
> > * maintenance
> > * risk by device
> > 
> > (idk, I could be wrong, idk what those are going to be)
> 
> I think there may be a bit of terminology confusion here. "Security", "self-sovereignty", and "Maintenance" are attribute _groups_, not attributes.
> 
> * An _attribute group_ is a pie chart, and corresponds to a folder under `src/schema/attributes`.
>   
>   * Example: `src/schema/attributes/security`, which is the "Security" attribute group, and contains various security-related attributes.
> * An _attribute_ is a single slice of the pie, and corresponds to a single `.ts` file under `src/schema/attributes/${some_attribute_group}`.
>   
>   * Example: `src/schema/attributes/security/security-audits.ts`, which looks at a wallet's security audit history specifically.
> * The summary table has one column per _attribute group_, which displays that attribute group's pie chart. Each slice of this pie chart corresponds to a single _attribute_.
> 
> I think this makes more sense if you look at the software wallet table, as it is more filled out. Click the pie charts on Ambire for example and it should be clear how these things map.
> 
> On the hardware wallet side, these are less well-defined, because we've historically more focused on the software side and the hardware side came later. So the current attribute groups are copy/pasted and mostly unrated right now, but in the long run we should aim to tighten them into a coherent hierarchy that makes sense for hardware wallets.
> 
> Each _attribute_ is evaluated for all wallet types (i.e. software wallet and hardware wallet alike). However, some attributes don't make sense to apply to software wallets, and some attributes don't make sense to apply to hardware wallets. Such attributes can return the `EXEMPT` rating, which makes them hidden from the pie slices. For example, look at `src/schema/attributes/security/chain-verification.ts`, you'll see that it does not apply to hardware wallets:
> 
> ```ts
>   evaluate: (features: ResolvedFeatures): Evaluation<ChainVerificationValue> => {
>     if (features.variant === Variant.HARDWARE) {
>       return exempt(
>         chainVerification,
>         sentence('This attribute is not applicable for hardware wallets.'),
>         brand,
>         null,
>       );
>     }
>     // ... rest of the code goes here for rating software wallets...
>   },
> ```
> 
> This way, the "Chain verification" attribute, which is part of the "Security" attribute group, will show up in the software wallet table, but will not show up for the hardware wallet table. Each of these tables is created dynamically based on whether any of the wallets in that table is non-`EXEMPT`. And if all attributes in an attribute group are `EXEMPT` for all wallets, then that attribute group will automatically disappear from the table (and the column will be gone).
> 
> We only recently added hardware wallets, and we didn't go through all the attributes to see if they made sense to also apply to hardware wallets. So if you'd like, you could do a pass and add code like the Chain Verification code above to return `EXEMPT` for hardware wallets, and that'll shrink the table to only the attributes that make sense for hardware wallets. But that's for another PR.
> 
> For this PR, I think let's go ahead and merge this. We can later refactor the attribute logic so that there is a new "Transaction clarity" attribute. I think that would make total sense to have as part of the "Security" attribute group.
> 
> Hope this makes sense!

Makes perfect sense. I'll make a ticket and assign it to me.

PatrickAlphaC:
Edit: I cannot assign it to me

src/schema/features/security/hardware-wallet-dapp-signing.ts
PatrickAlphaC: Thank you for your feedback to encourage me to make it so much more objective!

PR #172, opened 2025-06-14
Remove variant selector for hardware wallets.

polymutex:
Unlike the software wallet table, there is only one variant on the hardware wallet table, so it makes no sense to display it there.


polymutex:
cc @PatrickAlphaC

Hugoo (approved):
makes sense :)

PR #174, opened 2025-06-16
Implement chain abstraction attribute.

polymutex:
Also filled in the data for Daimo and Rabby wallet.


src/schema/attributes/ecosystem/chain-abstraction.ts
Hugoo, line 1: great file with great explanations !

Hugoo (approved):

src/schema/attributes/ecosystem/chain-abstraction.ts
polymutex: :saluting_face:

PR #183, opened 2025-06-19
Use SSH authentication when pushing Helios checkpoint

polymutex:
CI needs to be able to push updated Helios checkpoints every so often (about once a week at least). To avoid human intervention, commits should be fully automated and thereby exempt from PR review requirements. To that end, CI needs to auth itself over SSH as a GitHub user that has the bypass permission on the `beta` branch protection rule. This PR fixes that.

### DIFF
--- a/.github/workflows/helios-refresh.yaml
+++ b/.github/workflows/helios-refresh.yaml
@@ -19,6 +19,11 @@ jobs:
         env:
           GIT_USERNAME: walletbeat-bot
           GIT_EMAIL: walletbeat-bot@users.noreply.github.com
+          GIT_REMOTE_URL: git@github.com:walletbeat/walletbeat.git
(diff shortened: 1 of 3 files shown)


Hugoo (approved):

Hugoo:
idk if it is somewhere else in the code but it might be nice to document this in a .md file, just to explain flow, how to set the keys (GIT_SSH_AUTHENTICATION_KEY_PUBLIC )etc on the helios side. if someone else come, they can get the context

polymutex:
> idk if it is somewhere else in the code but it might be nice to document this in a .md file, just to explain flow, how to set the keys (GIT_SSH_AUTHENTICATION_KEY_PUBLIC )etc on the helios side. if someone else come, they can get the context

I agree. Filed https://github.com/walletbeat/walletbeat/issues/188

Issue #79, opened 2025-01-24
Main page: Revamped wallet comparison table


polymutex:
The current wallet comparison table is powered by [MUI's Data Grid component](https://mui.com/x/react-data-grid/column-definition/), which is very versatile but that we may eventually exceed the capabilities of.

As I see it, Walletbeat's main wallet comparison table can already do the following and these features should be preserved:

* Show basic wallet information: wallet name, icon, platforms
* Support wallet platform selection (mobile/desktop/web, perhaps there may be more in the future)
* Have interactive pie-chart-type ratings for all attribute groups that make sense; note that the pie charts may be different across wallets because some will be exempt from some attributes.
* Support having an "expanded view" for each row, which still fits enough data on mobile phone screens for at least two wallets and provides a useful wallet comparison view

Things the wallet table does not yet do but ought to:

* Support filtering wallets by criteria, such as by profile and platform (example: "I only want to see payments-focused wallets that run on mobile")
* Have room for a per-wallet "letter grade" rating that summarizes all the ratings it has
* Partial rendering (don't render the entire table, only what's in view and some rows before/after).
* If the wallet data takes a while to load, make it possible to show some feedback that data is loading in the table.
* Better appearance and readability on phone-sized screens. Ideally without horizontal scrolling somehow, but if there is horizontal scrolling, the wallet title should remain visible instead of scrolling away from view.

I'm not sure what the ultimate table looks like, so we probably need some mockups before moving on to implementation here.

nuconomy:
+1 for filter

PatrickAlphaC:
Nice. I was actually just thinking about how the pie charts don't really convey what is important. Like if a wallet has everything, but keeps getting hacked every other week, it's probably not a good project. That would be a _really_ important factor to keep in mind when choosing a wallet.

PR #189, opened 2025-06-22
Migrate React → Astro + Svelte

darrylyeo:
---

Follow-up to #99, #104, #137 and #141.

1. Migrate layout and navigation from React → Astro.
   * Features:
      * Collapsible nav groups with native HTML `<details>` / `<summary>` (keeps site navigable even if the user disables JavaScript)
      * "Table of contents" nav on wallet detail pages using native [CSS scroll markers](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_overflow/CSS_carousels) (progressively enhanced; Chromium browsers only for now)
      * Nested sticky containers with accumulating [sticky position](https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky) insets (enables hierarchical sticky headers, scroll-snapping, and screen-wide content scrolling regardless of pointer location)

2. Migrate wallet tables, pie charts, tooltips and wallet details pages from React → Svelte.
    * Custom interactive Svelte components built from scratch:
      * `Table.svelte` component with sortable rows and hierarchically-collapsible columns
      * `Filters.svelte` component defining and selecting mutually-exclusive or inclusive sets of filter functions to apply to an array of data
      * `Pie.svelte` component displaying selectable, hoverable and keyboard-focusable SVG pie slices in a hierarchy
      * `Tooltip.svelte` component using browser-native [HTML Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using) and [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning/Using)

3. Remove all React-based components and tooling (`@astrojs/react`, MUI, Radix UI, TanStack Table, `cmdk`, `classnames`, etc.).

4. Remove CSS-in-JS libraries and CSS preprocessing tools (`tailwindcss`, `emotion`, `sass-embedded`) in favor of common utility attributes in `global.css` and component-scoped vanilla CSS `<style>` tags built into Astro and Svelte.

5. Replace framework-specific icon wrapper packages with raw SVG counterparts:

    * `@mui/icons-material` → `@material-icons/svg`
    * `react-icons` → `lucide-static`

6. Add some new features:

    * navigation search with Cmd+K shortcut
    * software wallet variant / hardware wallet model selection on wallet pages
    * hierarchically expandable table columns
    * multi-level pie charts
    * hierarchical sticky inset / positioning
    * scroll-driven animations for headings and pies on wallet pages (Chromium browsers only / progressively enhanced)
    * CSS scroll marker-based Table of Contents on wallet pages (Chromium browsers only / progressively enhanced)

---

This PR includes parallel migrations across three iterations of the Walletbeat website, encompassing work from the following branches (in rough chronological / topological order):
* https://github.com/walletbeat/walletbeat/tree/beta-svelte-data-table (https://github.com/walletbeat/walletbeat/pull/117)
* https://github.com/walletbeat/walletbeat/tree/beta-svelte-data-table-nested-pie (https://github.com/walletbeat/walletbeat/pull/118)
* https://github.com/walletbeat/walletbeat/tree/restore-formatting (https://github.com/walletbeat/walletbeat/pull/159)
* https://github.com/walletbeat/walletbeat/tree/beta-svelte-wallet-page (https://github.com/walletbeat/walletbeat/pull/143)
* https://github.com/walletbeat/walletbeat/tree/beta-astro-layout-navigation
* https://github.com/walletbeat/walletbeat/tree/beta-svelte-data-table-2
* https://github.com/walletbeat/walletbeat/tree/beta-astro-layout-navigation-2
* https://github.com/walletbeat/walletbeat/tree/beta-astro-svelte (https://github.com/walletbeat/walletbeat/pull/189)

For the best code reviewing experience, see the https://github.com/walletbeat/walletbeat/compare/restore-formatting...beta-astro-svelte diff.


polymutex:
I've played with this, and gotta say that table is super slick. The tanstack table was already an improvement but this goes beyond that. The filtering is clearer, and it's both sortable by overall rating or attribute group column score. Impressive. It's also a clear demonstration of the value of maintaining our own table component rather than using an off-the-shelf one.

Also I like the new pie slice component. The technique you used to replace the emojis with silhouettes by default (and only re-colorize on hover) is smart.

A few things:

- Upon startup (in a clean checkout in which I did `pnpm install`), I noticed the following error when running `pnpm dev`:

```
$ pnpm dev
[...]
 astro  v5.9.3 ready in 470 ms

┃ Local    http://localhost:4321/
┃ Network  use --host to expose

13:04:47 watching for file changes...
13:04:47 [ERROR] [vite] Error: The following dependencies are imported but could not be resolved:

  react-icons/lu (imported by /tmp/walletbeat/src/ui/organisms/Eip7702Adoption.tsx)
  @tanstack/react-table (imported by /tmp/walletbeat/src/ui/organisms/Eip7702Table.tsx)
  react (imported by /tmp/walletbeat/src/ui/organisms/Eip7702Table.tsx)
  @/utils/cx (imported by /tmp/walletbeat/src/ui/organisms/Eip7702Table.tsx)

Are they installed?
    at file:///tmp/walletbeat/node_modules/.pnpm/vite@6.3.5_@types+node@22.15.17_jiti@2.4.2_lightningcss@1.29.2_sass-embedded@1.87.0_tsx@4.19.3_yaml@2.7.1/node_modules/vite/dist/node/chunks/dep-DBxKXgDP.js:14849:15
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///tmp/walletbeat/node_modules/.pnpm/vite@6.3.5_@types+node@22.15.17_jiti@2.4.2_lightningcss@1.29.2_sass-embedded@1.87.0_tsx@4.19.3_yaml@2.7.1/node_modules/vite/dist/node/chunks/dep-DBxKXgDP.js:46985:28
```

This is despite having run `pnpm install`. The local server still works anyway.

- The EIP-7702 tracker page is missing. Probably related to the above error :)
- The table tooltips show up at the correct position on Chromium, but show up at coordinates (0, 0) on Firefox.
- _(not for this PR)_ Switching between expanded and non-expanded row view feels clunky. I think this is due to (a) the lack of smooth animation, and (b) the need for an x-axis sideways re-jiggering on Firefox (doesn't happen on Chromium). I don't think this needs to be fixed prior to merging this, but I do think we should plan to fix this.
- The pie slice inner radius alignment issue remains. It's not as obvious because now most pie slices are 4 elements or more, but if you look at the 3-slice pie charts it's still apparent.
- _(not for this PR)_ The per-wallet page pie slice should make better use of the space. I'm not sure how yet, but feels like there's plenty of room to show text labels for each slice for example.
- The per-wallet page attribute descriptions have too little margin, especially around the "Sources" list when present. Also, the "Sources" thing is not vertically aligned with the rest of the attribute box's text, so it feels out of place.
- _(not for this PR)_ The expandable boxes on the per-wallet page should probably have opening/closing animations.
- _(not for this PR)_ The "By rating" pages are just a subset of the summary table, which looks rather awkward. I don't think they are that useful in their current format. We should lay out the table differently for them, similar to the current tanstack version. But that can wait for after this PR.
- The sidebar having up to _4_ levels (in the Wallets -> Software wallets -> By wallet -> Some wallet) feels too deep. I think the "Software wallets", "Hardware wallets", and "Embedded wallets" can all be top-level menu items, which would bring down the max depth to 3.
- The light/dark toggle is broken. I don't think we strictly need this to be fixed in this PR, but if it remains broken then the button for it should be removed until it is fixed.

Lastly, yes, I need to bring up tabs vs spaces and formatting here. As mentioned before, I don't have a strong opinion on which one we should use, and I understand that it's not your fault that the code formatting style got changed mid-way. But what I do have a strong opinion on is that we should aim to have this massive Astro/Svelte migration be orthogonal to code formatting changes. This keeps the diffs meaningful, and makes it easier to understand things like what packages in `package.json` have changed (whereas right now the whole file looks like it was rewritten). Otherwise, it feels like swapping a jet engine turbine in mid-air while also fixing the paint coating of the aircraft wing at the same time. The paint coating may be a good thing to do, but it still adds avoidable complexity to the process of swapping the turbine, and doing paint coating in mid-air seems like the wrong time to do it.

I know your preferred solution would be to change the `beta` branch's code formatting back to tabs and no semicolons, but others disagree. So, to come to a hopefully happy resolution for everyone, I volunteer to implement the following solution. I will write a script that iterates through all the commits of this branch, and creates a parallel branch `beta-astro-svelte-p` that, for each commit on the `beta-astro-svelte` branch, creates a parallel commit on `beta-astro-svelte-p` that applies the same diff but with the `beta` branch's current formatting re-applied on top (i.e. 2-space indentation + semicolons). This will recreate a parallel history on `beta-astro-svelte-p` that performs the same code edits as the `beta-astro-svelte` branch does, but remains diffable. Then you can run that script, so that all the commits remain signed by your key and preserve authorship information.

Then we can merge `beta-astro-svelte-p`, and then we can have a separate PR that reformats the codebase.

Thoughts?

polymutex:
One more decision from the call: Need to hide the scores. They can be kept around internally for sorting, but do not display them. If we decide to do scores, they should be tied more strongly around stages, rather than a weighted average of attributes. Also, hiding the scores will make the inner arc alignment issue no longer a concern, since there would no longer be an inner arc to begin with.

I also noticed that the mouse hover targets for the "orange slice" view are not static, they move with the hover target itself. This creates an awkward loop where the hover target moves and slips out of being hovered, causing it to un-hover back into place, which causes it to get hovered again, etc. causing flicker. To avoid this, the hover targets should be invisible static regions that map to something close to the maximum expanded size of each component of the pie chart. Could be a follow-up PR.

Another one: In the expanded row view, the text underneath each pie chart can change in height. This creates awkward vertical jumpiness when hovering over the slices of the pie while the row is expanded. The row should maintain the tallest possible height it needs to be able to render any of its possible cell contents, in order to prevent this jumpiness. Could also be a follow-up PR.

darrylyeo:
Thanks for the detailed feedback!

Re: tabs vs. spaces – I've already been maintaining a separate [Restore Formatting](https://github.com/walletbeat/walletbeat/pull/159) PR that tracks `beta` but with the original formatting applied. This branch now encompasses that branch, so I recommend looking at this https://github.com/walletbeat/walletbeat/compare/restore-formatting...beta-astro-svelte diff for purposes of code review.

At this point I'd rather not add yet another -zig to our zig-zag... hope you understand what I mean. I think it would be easiest if we just merged the Restore Formatting PR the moment everyone else is at a stopping point. Then we can merge this PR cleanly.

polymutex:
I understand what you mean. The downside of that solution is that this would mean this PR is blocked until everyone is at a good "stopping point". I guess this would be a requirement for the Svelte/Astro migration itself anyway, as it does touch a lot of files regardless, but formatting changes do also touch the data files and the `src/schema` files and such.

I am supportive of going with either method, but if you prefer to go with the formatting changes first before merging in this PR, then I think the responsibility of finding such a stopping point (for the purpose of formatting changes) would fall on you. This means coordinating with the others to make sure they don't have work in-flight. Right now might be a good time to do so, as I'm not aware of any large in-flight work pending other than this very migration. But it would be good to get explicit sign-off from @lucemans, @Hugoo, and @PatrickAlphaC. And if you don't hear back an answer within some number of days, then we can choose to have an action bias and go ahead.

(For the purpose of my own in-flight work, you can treat this message as a "I'm good whenever" in terms of "stopping point". Rebasing changes for reformatting isn't very difficult.)

[edit: I have also posted this to #159, and we should keep further discussions on the logistics of merging code format changes to that PR.]

PatrickAlphaC:
For the next couple weeks, I'll be slow to help out here, so I wouldn't want to be a blocker. Also, from my perspective stuff looks ok from a very very shallow look!

polymutex:
What is the status of this PR? I'd like to get this merged.

darrylyeo:
Addresses https://github.com/walletbeat/walletbeat/issues/79

PR #204, opened 2025-06-27
Add Enkrypt, MEW Portfolio and MEW Mobile wallets

Abyscuit:
Add wallets: Enkrypt, MEW Portfolio and MEW Mobile
Add wallet dev: MyEtherWallet
Add contributor: Abyscuit


polymutex:
Thanks for filling this in!

I think we should first clarify the wallet grouping at a high level, before going into each of the wallet feature data. The `beta` branch's wallet data type system is meant to be able to represent multi-platform wallets as a single wallet. For example, Rabby has a mobile, desktop, and browser extension, but all of them are grouped under the same `rabby.ts` software wallet entry. Many of the feature data have type `VariantFeature<T>`, which allows them to be set to either `null` (unknown), or `T`, or `{ desktop?: T, mobile?: T, browser?: T }`. This way, cross-platform wallets where the feature data is all the same (or where they share code across all platforms) can be directly `T`, whereas features where the implementation differ between platforms can use the `{ desktop?: T, mobile?: T, browser?: T }` format.

That said, I am not sure the product suite here should be classified either as 3 independent wallets, or as one single wallet. It seems that https://www.myetherwallet.com/ and https://www.enkrypt.com/ and https://www.mewwallet.com/ all link to each other, are owned by the same entity, and, in the case of MEW Portfolio and MEW Wallet, share part of their name and the same branding/icon, which seems like an argument to group them together.

However, looking at their functionality, I'm not convinced that MEW Portfolio would really be classified as a software wallet by Walletbeat. It looks more like a portfolio viewer (and its name suggests this), i.e. it is a read-only application, with optional transaction signing features if you connect an external browser extension or hardware wallet to it. That said, it does have a private key/seed phrase import option (marked as "not recommended"), so I guess it can be used as a wallet too...? and it does have "wallet" in its domain name. And the wallet access page promotes the use of the Enkrypt browser extension and the MEW wallet app as "Official", which is an argument for grouping them. This is difficult to classify :) Perhaps there's case to be made that MEW Portfolio should be seen as a part of the Enkrypt browser extension, and also as part of the MEW wallet app, rather than its own wallet?

The case for keeping Encrypt and MEW Wallet separate from each other is probably stronger; they don't seem to be as connected to each other as they are to MEW Portfolio. So not sure whether you'd prefer to keep them separate or not.

I realize this is a long and rather unclear comment... Mostly a stream of consciousness about classification for these wallets.

Could you describe the branding and the strategy for these 3 products? Do you plan to make them converge into a more unified interface, or do you see them as three separate products that Enkrypt publishes independently? Do they share some of their codebase?

Abyscuit:
> Thanks for filling this in!
> 
> I think we should first clarify the wallet grouping at a high level, before going into each of the wallet feature data. The `beta` branch's wallet data type system is meant to be able to represent multi-platform wallets as a single wallet. For example, Rabby has a mobile, desktop, and browser extension, but all of them are grouped under the same `rabby.ts` software wallet entry. Many of the feature data have type `VariantFeature<T>`, which allows them to be set to either `null` (unknown), or `T`, or `{ desktop?: T, mobile?: T, browser?: T }`. This way, cross-platform wallets where the feature data is all the same (or where they share code across all platforms) can be directly `T`, whereas features where the implementation differ between platforms can use the `{ desktop?: T, mobile?: T, browser?: T }` format.
> 
> That said, I am not sure the product suite here should be classified either as 3 independent wallets, or as one single wallet. It seems that https://www.myetherwallet.com/ and https://www.enkrypt.com/ and https://www.mewwallet.com/ all link to each other, are owned by the same entity, and, in the case of MEW Portfolio and MEW Wallet, share part of their name and the same branding/icon, which seems like an argument to group them together.
> 
> However, looking at their functionality, I'm not convinced that MEW Portfolio would really be classified as a software wallet by Walletbeat. It looks more like a portfolio viewer (and its name suggests this), i.e. it is a read-only application, with optional transaction signing features if you connect an external browser extension or hardware wallet to it. That said, it does have a private key/seed phrase import option (marked as "not recommended"), so I guess it can be used as a wallet too...? and it does have "wallet" in its domain name. And the wallet access page promotes the use of the Enkrypt browser extension and the MEW wallet app as "Official", which is an argument for grouping them. This is difficult to classify :) Perhaps there's case to be made that MEW Portfolio should be seen as a part of the Enkrypt browser extension, and also as part of the MEW wallet app, rather than its own wallet?
> 
> The case for keeping Encrypt and MEW Wallet separate from each other is probably stronger; they don't seem to be as connected to each other as they are to MEW Portfolio. So not sure whether you'd prefer to keep them separate or not.
> 
> I realize this is a long and rather unclear comment... Mostly a stream of consciousness about classification for these wallets.
> 
> Could you describe the branding and the strategy for these 3 products? Do you plan to make them converge into a more unified interface, or do you see them as three separate products that Enkrypt publishes independently? Do they share some of their codebase?

@polymutex While MyEtherWallet is the company, each of the product lines is inherently unique serving different user bases and needs.

data/entities/myetherwallet.ts
```diff
@@ -0,0 +1,26 @@
...
+		chainDataProvider: false,
+		corporate: true,
+		dataBroker: false,
+		exchange: false,
+		offchainDataProvider: false,
+		securityAuditor: false,
+		transactionBroadcastProvider: false,
+		walletDeveloper: true,
+	},
+	crunchbase: { type: 'NO_CRUNCHBASE_URL' },
```
polymutex, line 17: https://www.crunchbase.com/organization/myetherwallet

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+			dataCollection: {
+				[Variant.BROWSER]: {
+					collectedByEntities: [
+						{
+							// Enkrypt is owned by MyEtherWallet.
+							entity: myetherwalletEntity,
+							leaks: {
+								cexAccount: Leak.NEVER,
+								endpoint: RegularEndpoint,
+								ipAddress: Leak.OPT_IN,
```
polymutex, line 141: Is it really `OPT_IN`? That would imply that, by default, MEW cannot learn the user's IP address, but that the user could "opt into" having MEW learn it.
The only way I could see this being the case is if all requests to MEW servers go through a mixnet by default, and that there is a setting users can choose to disable the use of a mixnet.
If that's not the case, then most likely this should be `ipAddress: Leak.ALWAYS`.

Edit: After doing a network capture of Chromium running enKrypt, it doesn't appear that a mixnet is in use. Therefore, MEW learns the user's IP address, so this should be `ipAddress: Leak.ALWAYS`.

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+								mempoolTransactions: Leak.ALWAYS,
```
polymutex, line 142: The `chainConfigurability` attribute above says that it is possible to customize the chain RPC endpoints. Can MEW still learn of the user's mempool transactions even when these RPC endpoint settings are changed?
- If yes (for example, if MEW is still involved for other parts of the transaction submission process, such as for transaction simulation), then `Leak.ALWAYS` is appropriate.
- If not (because the only entity that learns of these transactions is the chain's RPC endpoint), then this should be `Leak.BY_DEFAULT`.

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+								multiAddress: {
+									type: MultiAddressPolicy.ACTIVE_ADDRESS_ONLY,
+								},
+								ref: [
+									{
+										explanation: 'Enkrypt uses Amplitude Analytics to track user actions.',
```
polymutex, line 148: I've installed enKrypt and it made a request to `https://sr-client-cfg.amplitude.com/config?api_key=apikey&config_keys=analyticsSDK`, which isn't a MEW-owned domain as far as I know. So this config needs to have two separate blocks: one for the data that is leaked to `amplitude.com`, and one for the data that is leaked to MEW-owned domains (`*.mewwallet.dev`).

This allows specifying the subset of data that is leaked to each. For example, as far as I can tell, `amplitude.com` does not learn of the user's Ethereum address, whereas MEW does.

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+										url: 'https://github.com/search?q=repo%3Aenkryptcom%2FenKrypt+amplitude&type=commits',
+									},
+								],
+								walletAddress: Leak.ALWAYS,
+							},
+						},
+					],
```
polymutex, line 155: I've tried creating a wallet and it made a request to `https://backupstore.enkrypt.com/backups/<long hex string>?signature=<another long hex string>`. Can you describe what this request is for? What payload do the long hex strings contain? Does this allow MEW to figure out the user's Ethereum address, public key, private key, or seed phrase? (or is some of this encrypted by a secret which doesn't leave the browser?)

polymutex, line 155: It seems there is also a request to `https://raw.githubusercontent.com/enkryptcom/dynamic-data/main/configs/versions.json`, presumably to find the latest version. I also found a request for `https://raw.githubusercontent.com/enkryptcom/dynamic-data/main/swaplists/ETH.json` and `https://raw.githubusercontent.com/enkryptcom/dynamic-data/main/swaplists/top-tokens.json`. Either way, this reveals the user's IP to `github.com`, but no other information. Please add it.

polymutex, line 155: Icons for cryptoassets are requested from `https://coin-images.coingecko.com`. This leaks the user's IP and the specific combination of assets in the user's wallet (`walletAssets`) to `coingecko.com`. Please add it.

polymutex, line 155: Looking at the "Activity" tab makes a request to `https://eth.blockscout.com/api?module=account&action=txlist&address=<address>&sort=desc`. This leaks the user's IP address and Ethereum address to blockscout. Please add it.

polymutex, line 155: Looking at the "Activity" tab makes a request to `https://api.rango.exchange/basic/meta?apiKey=<some key>&excludeNonPopulars=true&transactionTypes=EVM,SOLANA`. This leaks the user's IP to `rango.exchange`. Please add it.

polymutex, line 155: I've seen requests to `spaceapi.prd.space.id` when initiating a swap transaction. Is this for name resolution? What data is sent there? Please add it.

polymutex, line 155: Initiating a swap transaction also makes a request to `public-node.rsk.co`. Please add.

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+						{
+							// Enkrypt is owned by MyEtherWallet.
+							entity: myetherwalletEntity,
+							leaks: {
+								cexAccount: Leak.NEVER,
+								endpoint: RegularEndpoint,
+								ipAddress: Leak.OPT_IN,
+								mempoolTransactions: Leak.ALWAYS,
+								multiAddress: {
+									type: MultiAddressPolicy.ACTIVE_ADDRESS_ONLY,
```
polymutex, line 144: While I've found this to be true when trying to add other accounts, it's also the case that requests made to `https://analytics-enkrypt.mewwallet.dev/record` carry identifier strings in their JSON data (specifically `user_id` and `device_id`) which do not change even when the user changes their active address. These strings could potentially be used by MEW to infer cross-address correlation, by looking at the timing of requests for two successive Ethereum addresses from the same IP, for which there are also `/record` requests from the same IP that carry these same identifiers. I don't think there's a good way to express this correlatability risk in this data structure right now, so no changes needed here for now, but I filed #227 to remember to add this here later.

data/software-wallets/enkrypt.ts
```diff
@@ -0,0 +1,246 @@
...
+			publicSecurityAudits: null,
+			scamAlerts: {
+				contractTransactionWarning: supported({
+					contractRegistry: true,
+					leaksContractAddress: true,
+					leaksUserAddress: false,
+					leaksUserIp: false,
+					previousContractInteractionWarning: false,
+					recentContractWarning: false,
+					ref: null,
```
polymutex, line 215: Can you add a ref here to describe which `contractRegistry` enKrypt uses?

public/images/wallets/enkrypt.svg
polymutex, line 1: Can you crop this SVG such that there are no transparent margins around the logo?

polymutex (changes requested):
Thanks, I agree with your assessment then that these should be treated as separate wallets.
I've started reviewing Enkrypt (haven't looked at MEW Portfolio or MEW mobile yet) and left comments on that which I think may also apply to the other two. A bunch of them are around the data collection information. I recommend running the wallet in a browser profile and log all network traffic to determine this; that's what I did to double-check this, and then I tried to clicking around in the wallet UI and to initiate a swap, then looked at the network logs in the Chromium network inspector. For MEW Mobile, it may be possible to get similar data using a tool like `mitmproxy`, though I haven't tried to set that up myself.

Issue #200, opened 2025-06-23
Software wallet integration for transaction legibility on supported hardware wallets


polymutex:
As per our call on 2025-06-22, one of the things software wallets are currently not rated for is whether or not their integration with hardware wallets enables these hardware wallets to understand the transaction they are signing. As I understand it, the current (sad) state of affairs for hardware wallet transaction legibility requires work on both the software side and the hardware wallet side.

This issue tracks the addition of wallet feature data to track the **software side** of this problem. The goal is that such feature data accurately describes the level of integration that software wallets have for enabling transaction legibility on the hardware wallets they support.

To make this determination, we should probably have a canonical set of "benchmark" transactions that we use to test with. @PatrickAlphaC has already done the work of identifying such transactions in the context of #161 for calldata decoding support. We should probably use the very same set of transactions here.

nconsigny:
https://github.com/walletbeat/walletbeat/pull/205

This PR refactors the hardware wallet integration criterion for software wallets

polymutex:
I believe this is done.

PR #205, opened 2025-06-28
Issue 200 | Software wallet integration for transaction legibility

nconsigny:
#200 

refactored the criterion for transaction legibility 

Update explanations and URLs for supported wallets, ensuring clarity on integration capabilities. Refactor hardware wallet support evaluation logic for better handling of unrated statuses.

### DIFF
--- a/data/software-wallets/zerion.ts
+++ b/data/software-wallets/zerion.ts
@@ -1,7 +1,10 @@
 import { lucemans } from '@/data/contributors/lucemans';
 import { AccountType } from '@/schema/features/account-support';
 import { WalletProfile } from '@/schema/features/profile';
-import { HardwareWalletType } from '@/schema/features/security/hardware-wallet-support';
+import {
+  HardwareWalletConnection,
+  HardwareWalletType,
+} from '@/schema/features/security/hardware-wallet-support';
 import { PasskeyVerificationLibrary } from '@/schema/features/security/passkey-verification';
 import { TransactionSubmissionL2Type } from '@/schema/features/self-sovereignty/transaction-submission';
 import { notSupported, supported } from '@/schema/features/support';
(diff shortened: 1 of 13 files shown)


data/software-wallets/zerion.ts
```diff
@@ -88,12 +88,39 @@ export const zerion: SoftwareWallet = {
...
+              url: 'https://www.ledger.com/zerion',
+            },
+          }),
+          [HardwareWalletType.TREZOR]: supported({
+            ref: {
+              explanation:
+                'Trezor can be used with Zerion via WalletConnect and external-wallet routes, with one extra hop.',
+              url: 'https://www.ledger.com/zerion',
+            },
+          }),
```
polymutex, line 116: These explanations all mention the mechanism used to connect the hardware wallet to the software wallet. How about making that an enum map of supported connection types, instead of freeform text?

Something like this in `src/schema/features/security/hardware-wallet-support.ts`:

```typescript
/**
 * Types of hardware wallets that can be supported
 */
export enum HardwareWalletType {
  // ...
}

/**
 * Connection method between software and hardware wallet.
 */
export enum HardwareWalletConnection {
  USB = 'USB',
  WALLET_CONNECT = 'WALLET_CONNECT',
}

/**
 * Set of connection types that are supported for a single hardware wallet.
 */
export type SingleHardwareWalletSupport = NonEmptyRecord<HardwareWalletConnection, Support>;

/**
 * A record of hardware wallet types and their support status
 */
export type HardwareWalletSupport = WithRef<{
  /**
   * Record of hardware wallet types and their support status
   */
  supportedWallets: Partial<Record<HardwareWalletType, Support<SingleHardwareWalletSupport>>>;
}>;
```

This would allow the attribute evaluation logic to understand the distinction between WalletConnect-only hardware wallet support vs those that are integrated directly.

data/software-wallets/zerion.ts
```diff
@@ -88,12 +88,39 @@ export const zerion: SoftwareWallet = {
...
+          [HardwareWalletType.LEDGER]: supported({
+            ref: {
+              explanation: 'Ledger has native first-class support in Zerion.',
+              url: 'https://www.ledger.com/zerion',
+            },
+          }),
+          [HardwareWalletType.TREZOR]: supported({
+            ref: {
+              explanation:
+                'Trezor can be used with Zerion via WalletConnect and external-wallet routes, with one extra hop.',
```
polymutex: I'm not sure we should treat hardware wallets that only work over WalletConnect as "supported", because the use of WalletConnect implies that the user needs to rely on _another_ software wallet (in addition to the one being rated here) that actually integrates the hardware wallet, in order to use that wallet in the wallet described here.

(As an aside, the use of WalletConnect in general is probably not something we want to encourage long term too, as it creates a centralizing force where a single entity (walletconnect) is able to censor transactions and gather orderflow before others.)

data/software-wallets/zerion.ts
nconsigny: Good idea ! will do

data/software-wallets/zerion.ts
nconsigny: Absolutely, I wanted  to keep the bar low for hw support and the bar high for hw integration. But  objectively supporting through walletconnect should be equivalent to no support. We should prioritize direct webHID / webUSB connectors

nconsigny:
Ok did the changes what do you think @polymutex

src/schema/attributes/security/software-hw-integration.ts
```diff
@@ -276,34 +299,10 @@ export const softwareHWIntegration: Attribute<SoftwareHWIntegrationValue> = {
...
-
-    // Determine integration level based on hardware wallet support count
-    if (supportedHWBrands >= 2) {
-      result = excellentHardwareWalletIntegration(supportedHardwareWallets);
-    } else if (supportedHWBrands >= 1) {
-      result = goodHardwareWalletIntegration(supportedHardwareWallets, []);
-    }
-
-    // Return result with references if any
+    // If wallet supports any hardware wallet, rate as unrated/grey for now
```
polymutex, line 302: Planning to implement this in a future PR I assume?

polymutex (approved):

JIOjosBG:
This PR adds keystone as supported hw wallet for ambire, but this is not the case. Ambire never had support for keystone, not in the legacy web interface, not in the legacy mobile app, not in the current browser extension. I will prepare a PR to fix this
@polymutex @nconsigny

Issue #192, opened 2025-06-23
Wallet attribute: Private transfers (beyond stealth addresses)


polymutex:
The current "Private transfers" attribute should support more privacy-preserving token transfer options than only stealth addresses.

Suggested on the 2025-06-22 call:

- Tornado Cash
- Privacy Pools
- Railgun
- Aztec

We don't have to support all of them for launch, but we should support more than just stealth addresses to reflect the fact that the attribute is credibly about private transfers in general, rather than "picking winners" by appearing to designate stealth addresses as only viable private token transfer tech.

polymutex:
Still needed: Railgun, maybe Aztec.

polymutex:
Marking this done for now. The only one we're missing is Aztec, which we should add once its mainnet is fully launched.

PR #207, opened 2025-06-30
Recognize Tornado Cash Nova as a private transfer technology.

polymutex:
This expands the wallet feature data and the Private Transfers attribute to recognize Tornado Cash Nova as a private transfer technology.

Updates issue #192.


data/software-wallets/metamask.ts
```diff
@@ -89,7 +90,8 @@ export const metamask: SoftwareWallet = {
       privacyPolicy: 'https://consensys.io/privacy-notice',
       transactionPrivacy: {
         defaultFungibleTokenTransferMode: 'PUBLIC',
-        stealthAddresses: notSupported,
+        [PrivateTransferTechnology.STEALTH_ADDRESSES]: notSupported,
```
PatrickAlphaC: I don't understand. Are we expecting wallets to have this contract interaction baked into the wallet itself?

polymutex: Stealth addresses don't involve a contract (other than the meta-address registry).
But for Tornado Cash Nova and other privacy-preserving transfer technologies, yes, absolutely. [Vitalik's blog post on wallets](https://vitalik.eth.limo/general/2024/12/03/wallets.html) (and the EF's wallet prototype) are both going into this direction.

polymutex: Note that the attribute doesn't require every privacy transfer technology to be baked into the wallet, just at least one of them (any of them). That lines up with Walletbeat's general "don't want to pick winners" approach. I plan to add more possible integrations like Railgun and Aztec after this one.

PatrickAlphaC: Got it

PR #208, opened 2025-06-30
fix: Remove gapAngle so pie chart slices are better aligned

usame-algan:
## What it solves

Small but annoying detail 😅 

- Sets the `gapAngle` to 0 so the pie slices are visually aligned

## Screenshots

Before:
Screenshot 2025-06-30 at 15 33 54

After:
Screenshot 2025-06-30 at 15 34 06

### DIFF
--- a/src/ui/molecules/RatingDetailModal.tsx
+++ b/src/ui/molecules/RatingDetailModal.tsx
@@ -113,7 +113,7 @@ export function RatingDetailContent<Vs extends ValueSet>({
     const centerX = 150;
     const centerY = 150;
     const radius = 120;
-    const gapAngle = 2; // Gap in degrees
+    const gapAngle = 0; // Gap in degrees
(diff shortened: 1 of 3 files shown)


src/ui/molecules/RatingDetailModal.tsx
```diff
@@ -113,7 +113,7 @@ export function RatingDetailContent<Vs extends ValueSet>({
     const centerX = 150;
     const centerY = 150;
     const radius = 120;
-    const gapAngle = 2; // Gap in degrees
+    const gapAngle = 0; // Gap in degrees
```
usame-algan, line 116: Only wanted to make minimal changes so I didn't remove the variable. Unless there are plans to make use of it in the future this could be removed though.

polymutex:
Thanks for fixing this :) It was bugging me too... (Note however that this pie chart is being replaced in #118)

polymutex (approved):

PR #209, opened 2025-07-01
Fix some post-reformatting errors from #159

polymutex:


src/schema/attributes.ts
```diff
@@ -361,44 +361,44 @@ export interface Attribute<V extends Value> {
...
+				/**
+				 * The type of display used to render the rating scale.
+				 * "simple" means to render a simple renderable block of text, useful for
+				 * simple yes/no-type attributes.
+				 */
+				display: 'simple'
+
+				/** The content to display to explain the rating scale. */
+				content: TypographicContent
+		  }
```
darrylyeo, line 373: Adding spaces here feels like a really odd choice...

darrylyeo (approved):

src/schema/attributes.ts
polymutex: That's not me doing that, that's `prettier --write`... I think the intention is that the curl braces for this type of construction line up with each other (where `S` is a space):

```
\t\tSS}
\t\t|S{
```

PR #212, opened 2025-07-04
feat/getclave wallet

PatrickAlphaC:
Added the [clave](https://getclave.io/) wallet.

So this one required a lot of... odd stuff. Clave uses ZKsync Era native account abstraction. The main difference between this an 4337, is that native account abstraction transactions get sent directly to the normal mempool, whereas 4337 transactions are sent to an alt-mempool.

I didn't make a separate type of smart account because it's "basically" the same, with the added advantage of the AA wallet also being able to act like an EOA. However, it made me do some "funky" stuff... 

Some big changes:
- Changed `rawErc4337` -> `erc4337` (It might make sense to just give native AA wallet their own category instead...)
- made a couple new contract types

Let me know what you think!


data/entities/aiprise.ts
```diff
@@ -0,0 +1,26 @@
...
+	},
+	crunchbase: { type: 'NO_CRUNCHBASE_URL' },
+	farcaster: { type: 'NO_FARCASTER_PROFILE' },
+	icon: 'NO_ICON',
+	jurisdiction: 'Unknown',
+	linkedin: { type: 'NO_LINKEDIN_URL' },
+	privacyPolicy: 'https://aiprise.io/privacy-policy',
+	repoUrl: 'https://github.com/aiprise',
+	twitter: { type: 'NO_TWITTER_URL' },
+	url: 'https://aiprise.io',
```
polymutex: I think this should be .com, not .io. Same for the privacy policy URL.

data/entities/aiprise.ts
```diff
@@ -0,0 +1,26 @@
...
+		chainDataProvider: false,
+		corporate: true,
+		dataBroker: false,
+		exchange: false,
+		offchainDataProvider: true,
+		securityAuditor: false,
+		transactionBroadcastProvider: false,
+		walletDeveloper: false,
+	},
+	crunchbase: { type: 'NO_CRUNCHBASE_URL' },
```
polymutex: Crunchbase: https://www.crunchbase.com/organization/aiprise

LinkedIn: https://www.linkedin.com/company/aiprise-inc

data/entities/nethermind.ts
polymutex, line 1: Huh, didn't know Nethermind also did security audits...

src/schema/contracts.ts
```diff
@@ -34,3 +34,47 @@ export interface SmartWalletContract {
 		validateUserOp: Support
 	}
 }
+
+/**
+ * A native account abstraction contract deployer, for deploying smart accounts that use the
+ * chain's native account abstraction. Which mean, the wallet can send
+ * transactions to the public mempool.
+ */
```
polymutex, line 47: I think this docstring was intended for AADeployer, not NativeAccountAbstraction.

src/schema/contracts.ts
```diff
@@ -34,3 +34,47 @@ export interface SmartWalletContract {
...
+
+	/** Is the source code for this contract available? */
+	sourceCode:
+		| { available: false }
+		| WithRef<{
+				available: true
+		  }>
+}
+
+export interface AADeployer {
```
polymutex: I'd suggest the full unabbreviated `AccountAbstractionDeployer` here, as the other type earlier in this file (`NativeAccountAbstraction`) also doesn't use the "AA" abbreviation. More generally, the walletbeat type names don't use abbreviations, and I find that not using abbreviations makes it easier for LLMs to understand what the types mean without needing to read the file they're declared in.

src/schema/contracts.ts
```diff
@@ -34,3 +34,47 @@ export interface SmartWalletContract {
 		validateUserOp: Support
 	}
 }
+
```
polymutex, line 42: > I didn't make a separate type of smart account because it's "basically" the same, with the added advantage of the AA wallet also being able to act like an EOA. However, it made me do some "funky" stuff...

Yeah, I think this part needs more work. These types (`SmartWalletContract` and `NativeAccountAbstraction`) are now somewhat redundant with each other. On a semantic level, I think that's because they have some but not full overlap; they represent different things that now make sense to break out. Specifically:

- The _type_ of account abstraction the wallet uses: Native vs ERC-7702 vs ERC-4337
- For non-native account abstraction types, the EVM contract in question
- For the EVM contract, references to its source code and whether it implements standard operations.

It made sense to conflate these all in a `SmartWalletContract` single type prior to the introduction of this wallet, but now there's the need to support the notion of native account abstraction. Breaking them out also removes the need to have fields that would likely never be populated (e.g. `sourceCode` in `NativeAccountAbstraction` doesn't make much sense: what would that be the source code of?). So I think it's a good time to break out all of this.

Here's a suggestion on how this can be expressed:

```typescript
export type AccountAbstraction = ContractBasedAccountAbstraction | NativeAccountAbstraction

export interface ContractBasedAccountAbstraction {
  type: 'eip7702' | 'erc4337'
  contract: EVMContract
  methods: { ... }
}

export interface NativeAccountAbstraction {
  type: 'native' // Useful for type predicate, see below.
  chain: string
  deployer: EVMContract
}

export interface EVMContract {
  name: string
  address: EVMAddress
  sourceCode: { ... }
}
```

This also makes it easy to write TypeScript type predicate functions for these types, so that attributes can distinguish between them easily (whereas right now, the code needs to do things like `if ('chain' in contract)` which is brittle):

```typescript
export function isNativeAccountAbstraction(aa: AccountAbstraction): aa is NativeAccountAbstraction {
  return aa.type === 'native'
}
```

This function, thanks to the "`aa is NativeAccountAbstraction`" in the type signature, informs the compiler that `aa` now has all the fields of `NativeAccountAbstraction`, so no longer need to manually check individual fields.

src/ui/organisms/Eip7702Table.tsx
```diff
@@ -174,6 +174,11 @@ export default function Eip7702Table(): React.JSX.Element {
 				return <span style={{ opacity: 1 }}>Unknown - Send PR!</span>
 			}
 
+			// Handle NativeAccountAbstraction which doesn't have an address property
+			if ('chain' in contract) {
```
polymutex, line 178: As per my other comment, this is probably better done with type predicate functions.

src/ui/organisms/Eip7702Table.tsx
```diff
@@ -191,6 +196,11 @@ export default function Eip7702Table(): React.JSX.Element {
 				return null
 			}
 
+			// Handle NativeAccountAbstraction which doesn't have sourceCode property
+			if ('chain' in contract) {
```
polymutex, line 200: Same here.

data/software-wallets/clave.ts
```diff
@@ -0,0 +1,269 @@
...
+									explanation:
+										'AiPrise is used for KYC verification when users access Rain Card or Onramp services through Clave.',
+									url: 'https://aiprise.io/privacy-policy',
+								},
+							],
+						},
+					},
+				],
+				onchain: {
+					pseudonym: Leak.OPT_IN,
```
polymutex: Is this `OPT_IN`, or is this `PROMPTED`? All wallets I've seen that have support for some type of "usernames" typically actively ask the user for a username during wallet setup, which is represented by `PROMPTED` ("wallet actively asks"). By contrast, `OPT_IN` would mean the user is the one who has to actively dig in the settings (past the wallet setup process) to register their username.

data/software-wallets/clave.ts
```diff
@@ -0,0 +1,269 @@
...
+									url: 'https://www.getclave.com/privacy-policy',
+								},
+							],
+						},
+					},
+					{
+						entity: aiprise,
+						leaks: {
+							endpoint: RegularEndpoint,
+							ipAddress: Leak.OPT_IN,
```
polymutex, line 142: If this is used for KYC, I think this should have much more than just IP address. Probably also need to add `legalName`, `phone`, `physicalAddress`, maybe `face` if it includes facial recognition, maybe `governmentId` if it asks for government ID.

data/software-wallets/clave.ts
```diff
@@ -0,0 +1,269 @@
...
+				keyDerivation: {
+					type: 'NONSTANDARD',
+				},
+			}),
+			erc4337: supported({
+				contract: claveAccount,
+				controllingSharesInSelfCustodyByDefault: 'YES',
+				keyRotationTransactionGeneration:
+					TransactionGenerationCapability.USING_OPEN_SOURCE_STANDALONE_APP,
+				// Question: Technically, I can always send a transaction because I'm a developer... But I will often use their proprietary API to generate transactions... What should we put?
```
polymutex: I'm assuming that you, as a developer, would need to write your own code to generate such a transaction. As soon as you have to write code, you are no longer using the wallet software by itself to achieve your goal. The `tokenTransferTransactionGeneration` field (and the fields around here in general) refers specifically to what the wallet software by itself (or separate/add-on software developed and distributed by the same entity as the wallet itself) lets you do as a regular user. Feel free to adjust the docstrings to make sure this is clear.

src/schema/features/account-support.ts
```diff
@@ -33,13 +33,13 @@ export enum AccountType {
 	 * Raw ERC-4337 account, i.e. an account for which the address matches the
 	 * smart contract code.
 	 */
-	rawErc4337 = 'rawErc4337',
+	erc4337 = 'erc4337',
```
polymutex, line 36: Good change, I've been meaning to do this as well.

polymutex (changes requested):

polymutex:
Thanks, overall great PR and the `erc4337` enum change is good.

As part of the changes above, please also rebase (to avoid the merge conflict on `.cspell.json`), and use more descriptive commit names. This avoids the commit log having a large (refactoring + new wallet) commit that touches many files for which the commit message is "wip" :) Descriptive commit = better reflection of contribution.

src/schema/contracts.ts
polymutex: Your idea of adding a new AccountType (next to `erc4337`) is also good. Pro: better type distinction, no need for type predicate functions. Con: Some code duplication for attributes that should treat smart-contract-based AA wallets similarly to native AA wallets. Feel free to do either one.

data/entities/nethermind.ts
PatrickAlphaC: They sure do!

src/schema/contracts.ts
PatrickAlphaC: hmm... I think `ContractBased` vs `Native` isn't quite right, because even native account abstraction uses a contract, the main difference really is where the mempool is... I'll try to make the distinction there.

PatrickAlphaC: Hmm... going deeper I'm a bit more intrigued on what we should do. Right now we treat EIP-7702 smart accounts as the same as ERC-4337 accounts, when they are very different. 

I think we should have all three be different.

It might make sense to do like:

```javascript
export interface SmartWalletAccount {
  name: string
  erc4337?: ERC_4337
  eip7702: EIP_7702
  nativeAccountAbstraction: NativeAccountAbstraction
}
```

Since smart wallets can support multiple kinds of account abstraction... Or am I now over complicating it...

data/software-wallets/clave.ts
PatrickAlphaC: Ah this is prompted. My mistake.

data/software-wallets/clave.ts
PatrickAlphaC: Make sense. I guess if the wallet has support for both wallet connect (third party) and their own proprietary app, what would you like to see?

PatrickAlphaC:
@polymutex Oh whoops. Are we not squashing commits? Or are we doing merge commits?

And ok, I need to think a bit more, specifically on this one: https://github.com/walletbeat/walletbeat/pull/212#discussion_r2186168762

data/wallet-contracts/ambire-account.ts
```diff
@@ -17,3 +18,15 @@ export const ambireAccountContract: SmartWalletContract = {
 		},
 	},
 }
+
+const chains = ['ethereum', 'polygon', 'avalanche', 'bnb', 'fantom', 'arbitrum', 'gnosis', 'optimism', 'base', 'scroll', 'metis'] as const
```
polymutex, line 22: We should probably have a shared enum for this.

src/schema/features/account-support.ts
```diff
@@ -51,7 +51,11 @@ export enum TransactionGenerationCapability {
 	/** The process to generate such a transaction requires the use of a standalone proprietary application. */
 	USING_PROPRIETARY_STANDALONE_APP = 'USING_PROPRIETARY_STANDALONE_APP',
 
-	/** The process to generate such a transaction requires the use of an open-source standalone application. */
+	/** The process to generate such a transaction requires the use of an open-source standalone application.
+	 *
+	 * Note: This also means that a developer could also generate such a transaction
+	 * without relying on this open-sourced application.
```
polymutex, line 57: This phrasing makes it sound like it is _sufficient_ for a developer to be able to generate a such a transaction, in order to qualify for the `USING_OPEN_SOURCE_STANDALONE_APP` level. I think it's the opposite: the fact that a developer is able to generate such  transaction is _not_ sufficient to qualify for this level. `USING_OPEN_SOURCE_STANDALONE_APP` requires that non-developers (aka regular wallet users) are able to generate such transactions, without the help of a developer.

Example clarification comment if you want to use it:

```
/**
 * The process to generate such a transaction is doable using an application that
 * is open source, standalone (has no external third-party network dependencies),
 * and does not require developer expertise to use. An average non-developer wallet
 * user should be able to generate this transaction.
 */
```

src/schema/contracts.ts
polymutex: > Right now we treat EIP-7702 smart accounts as the same as ERC-4337 accounts, when they are very different

Yes and no... Stepping back a bit, the main intent behind collecting this data about smart wallets is to determine whether the account is truly "portable", i.e. exportable/importable into other wallets. This is a function of:

1. The standardization and interoperability of the type of AA the wallet uses
2. The permissions and dependencies of the export/import process (i.e. the generation and inclusion of token transfers and/or key rotation transactions)
3. The immutability of the onchain contracts that govern the above.

From that perspective, 7702 and 4337 are not that different: they both allow arbitrary contract logic to be used for transaction authorization purposes, they both require there being an EVM contract to implement that logic (following some of the same-ish standards (e.g. `validateUserOp`). So the data we collect about either should be mostly the same.

The one part that's different from the account portability perspective is the one around key rotation. With 7702, it's not possible to "gate" the account in the way that it is possible in 4337, because you can always broadcast a 7702 redelegation in the public mempool using the same EOA key, and the previous 7702-delegated smart contract logic can't block that redelegation from going through. That's why `AccountType4337` extends `AccountTypeMutableMultifactor` (which includes fields relating to key generation and key rotation transactions), whereas `AccountType7702` doesn't (because key generation and key rotation are the same as EOAs, so there are no 7702-specific concerns for account portability).

> Since smart wallets can support multiple kinds of account abstraction

That multiple account kind support is best represented at a higher level in the wallet feature type; specifically in `src/schema/features/account-support.ts`. The `AccountSupport` type has fields for each "kind" of account abstraction, so we could probably add a new one there. Each field in `AccountSupport` has its own account-type-specific type, e.g. the `eip7702` account type has `AccountType7702`, and the `erc4337` account type has `AccountType4337`. So a non-4337, non-7702 account type could have its own `AccountTypeNativeSmartAccount`, which can contain whatever makes sense to encompass the above. The intent of these `AccountType[Something]` types is to contain the account-type-specific considerations, whereas the intent of `SmartWalletContract` is to contain pure EVM contract information.

So I think you could have something like:

```patch
 export type AccountSupport = Exclude<
 	{
 		/**
 		 * Support for raw EOA accounts.
 		 * Leave as NOT_SUPPORTED if the wallet only supports EIP-7702-type EOAs.
 		 */
 		eoa: AccountTypeSupport<AccountTypeEoa>
 
 		/** Support for MPC-based (sharded key) accounts. */
 		mpc: AccountTypeSupport<AccountTypeMpc>
 
 		/**
 		 * Support for EIP-7702 EOA accounts.
 		 * This usually also implies `rawEoa` support.
 		 */
 		eip7702: AccountTypeSupport<AccountType7702>
 
 		/**
 		 * Support for smart accounts (pure ERC-4337 accounts for which the
 		 * address matches the contract code).
 		 */
 		erc4337: AccountTypeSupport<AccountType4337>

+		/**
+		 * Support for native AA smart accounts.
+		 */
+		nativeSmartAccount: AccountTypeSupport<AccountTypeNativeSmartAccount>
 	},
 	// At least one account type must be supported.
 	Record<AccountType, NotSupported>
 > & { defaultAccountType: AccountType }
```

Then you can split the `SmartWalletContract` type to be able to specialize it for the purpose of the deployer contracts:

```typescript
export interface SmartWalletContract {
	name: string
	address: EVMAddress // TODO: Maybe replace with Map<Chain, EVMAddress>?
	sourceCode:
		| { available: false }
		| WithRef<{
				available: true
		  }>
	// TODO: btw, maybe we need a `immutable` field here?
}

export interface SmartWalletERC4337Contract extends SmartWalletContract {
	type: '4337' // Useful for being able to write type predicate functions
	methods: {
		isValidSignature: Support
		validateUserOp: Support
	}
}

export interface SmartWalletEIP7702Contract extends SmartWalletContract {
	type: '7702' // Useful for being able to write type predicate functions
	methods: {
		isValidSignature: Support
		validateUserOp: Support
	}
}

export interface SmartWalletDeployerContract extends SmartWalletContract {
	type: 'deployer' // Useful for being able to write type predicate functions
	// Can add more fields here that are specific to deployer contracts
}
```

And with that, it becomes possible to express the right type of contract for each type of supported smart account:

```patch
 /** A wallet backed by a smart contract. */
-export interface SmartAccountType {
-	contract: 'UNKNOWN' | SmartWalletContract
+export interface SmartAccountType<SC extends SmartWalletContract> {
+	contract: 'UNKNOWN' | SC
 }
 
 /** Support information for ERC-4337 accounts. */
-export type AccountType4337 = AccountTypeMutableMultifactor & SmartAccountType
+export type AccountType4337 = AccountTypeMutableMultifactor & SmartAccountType<SmartWalletERC4337Contract>
 
 /**
  * Support information for EIP-7702 accounts.
  */
-export type AccountType7702 = SmartAccountType
+export type AccountType7702 = SmartAccountType<SmartWalletEIP7702Contract>
 
+/**
+ * Support information for native AA.
+ */
+export type AccountTypeNativeSmartAccount = SmartAccountType<SmartWalletDeployerContract>
```

Can also add more fields to `AccountTypeNativeSmartAccount` in case more details are needed to determine account portability.

How does that sound?

> I think `ContractBased` vs `Native` isn't quite right, because even native account abstraction uses a contract, the main difference really is where the mempool is

That's a great point, which the current type system doesn't capture well. You're right that the mempool the user may be _required_ to use has implications for account portability. I'm assuming that in the case of this "native" AA with the public mempool, that's _better_ (because presumably the public mempool has higher censorship resistance), whereas the use of "alt" mempools means there's an additional potential censorship vector for transactions.

So for the purpose of this PR, I'd suggest not worrying about this dimension yet, but to add as a follow-on PR for smart account types that require an alt mempool (since presumably, these native AA accounts don't have this problem).

polymutex:
> @polymutex Oh whoops. Are we not squashing commits? Or are we doing merge commits?

If you'd like to do squash commits, that's possible too, but please squash it on your own branch then. I've been doing merge commits when pulling in PRs, because squashing commits breaks their cryptographic signatures. If you squash it on your own branch, then you can sign the squashed commit :)

(btw, as an aside, you have write access to the walletbeat repo, so feel free to push to feature branches on this repo directly, rather than pushing to your fork. Only the `beta` branch is protected.)

PR #222, opened 2025-07-08
feat: added dapp connection support for hardware wallets, and signing…

### LINKED ISSUE #213: Rename: hardware-dapp-signing -> Signature Intent Verification
I feel like this is a more clear descriptor of what we are trying to... describe.

This would be good for both software and hardware wallets, it should be easy to verify the intent of your transaction in all wallets.

What do you think?

PatrickAlphaC:
This ended up being a bigger change, because as I dug deeper, more unearthed.
- It made sense to me to separate `signingIntentClarity` from a new attribute of `dappConnectionSupport` specifically for hardware wallets (we have a feature of what HW wallets a software wallet supports, why not the other way around?)
- Adding `signingIntentClarity` to all wallets I think is important, with the feature rating being _only_ what calldata/transaction data is being shown on the device itself (for example, if ledger live showed TX data, but not the device, we would mark that the ledger device does not show TX data)
- I added an image of coinbase wallet failing to show calldata/tx data, I wasn't sure where to put it as a reference
- softened the calldata decoding requirements if a wallet allows the user to extract that calldata
- it probably makes sense to put the `dappConnectionSupport` into the `ecosystem` bucket over security

Let me know what you think

Fixes: https://github.com/walletbeat/walletbeat/issues/213#event-18478799408


src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+export enum DappConnectionMethod {
+	/**
+	 * The wallet connects to dApps through its own proprietary closed-source application
+	 */
+	PROPRIETARY_CLOSED_SOURCE_APP = 'PROPRIETARY_CLOSED_SOURCE_APP',
+
+	/**
+	 * The wallet connects to dApps through its own open-source application
+	 */
+	PROPRIETARY_OPEN_SOURCE_APP = 'PROPRIETARY_OPEN_SOURCE_APP',
```
polymutex: "Proprietary" is often contrasted with "open-source", so "proprietary closed-source" seems like a tautology, and "proprietary open-source" like a contradiction. I assume that this is using "proprietary" in the "belongs the same entity as the hardware wallet manufacturer" sense. In that case, I suggest using the term "first-party" (`FIRST_PARTY_CLOSED_SOURCE_APP`, `FIRST_PARTY_OPEN_SOURCE_APP`) or "vendor" (`VENDOR_CLOSED_SOURCE_APP`, `VENDOR_OPEN_SOURCE_APP`) to be less ambiguous.

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+	 * Can the hardware wallet connect to dApps at all?
+	 */
+	canConnectToDapps: boolean
+
+	/**
+	 * Connection details if the wallet can connect to dApps
+	 * If canConnectToDapps is false, this should be null
+	 */
+	connectionDetails: Support<DappConnectionMethodDetails> | null
+}>
```
polymutex: I think this whole type can simply be

```typescript
/**
 * A record of hardware wallet dApp connection support
 */
export type DappConnectionSupport = Support<WithRef<DappConnectionMethodDetails>>
```

Some general tips:

- `Support<T>` mean "either a thing is supported (in which case you must provide `T`), either the feature is unsupported (and you don't need to provide anything)".
- `Support<WithRef<T>>` means "either a thing is supported (in which case you must provide `T` and optionally a reference), either the feature is unsupported (and you don't need to provide anything)".
- If you want the `ref` to be able to be provided even in the "unsupported" case (for example in case you want to be able to provide proof that it is unsupported), then this can be flipped into `WithRef<Support<DappConnectionMethodDetails>>`. Helper functions like `refs()` and `isSupported()` support it either way.

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Shorthand for a wallet that cannot connect to dApps
+ */
+export const cannotConnectToDapps: DappConnectionSupport = {
+	canConnectToDapps: false,
+	connectionDetails: null,
+}
```
polymutex: Then this `const` can be replaced with `unsupported`.

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+export interface DappConnectionMethodDetails {
+	/**
+	 * Which methodologies are supported?
+	 */ 
+	supportedConnectionMethods: Partial<Record<DappConnectionMethod, boolean>>
+
+	/**
+	 * If they support any software wallets: which software wallets are supported
+	 */
+	supportedSoftwareWallets?: Partial<Record<SoftwareWalletType, boolean>>
```
polymutex: I think this can be combined in a single record:

```typescript
supportedConnections: Partial<Record<DappConnectionMethod | SoftwareWalletType, boolean>>
```

Actually, you can instead use `NonEmptySet` to guarantee that at least one of them is set to `true` at the typesystem level (i.e. it is not possible to specify `supportedConnections` without at least one `true` element); this ensure that `DappConnectionSupport` below cannot be `Supported` without actually at least one connection type.

```typescript
supportedConnections: NonEmptySet<DappConnectionMethod | SoftwareWalletType>
```

There are a few helper functions for `NonEmptySet` in `src/types/utils/non-empty.ts`, but it is still also just a normal `Partial<Record<K, boolean>>` when specifying it, so you shouldn't need to change how it is specified in the `/data` files.

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+	/**
+	 * The wallet connects to dApps through its own open-source application
+	 */
+	PROPRIETARY_OPEN_SOURCE_APP = 'PROPRIETARY_OPEN_SOURCE_APP',
+
+	// If we decide to make this an feature for all wallets
+	/**
+	 * The wallet is injected into the window of the dapp (ie: most browser extension wallets)
+	 */
+	// INJECTED = 'INJECTED'
```
polymutex: No, this data is already tracked in the `WalletSoftwareFeatures.integration.browser` field for software wallets (and has more specifics about which browser integration EIP it implement there).

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+ */
+export const cannotConnectToDapps: DappConnectionSupport = {
+	canConnectToDapps: false,
+	connectionDetails: null,
+}
+
+/**
+ * Returns whether the wallet supports any connection method at all
+ */
+export function supportsAnyConnectionMethod(connectionDetails: Support<DappConnectionMethodDetails> | null): boolean {
```
polymutex: This function becomes equivalent to simply calling `isSupported`, if using `NonEmptySet` inside `DappConnectionMethodDetails`.

src/schema/features/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,156 @@
...
+		.filter(supported => supported === true).length
+	
+	// Count each software wallet as an individual connection method
+	return connectionMethodCount + softwareWalletCount
+}
+
+/**
+ * Returns all supported software wallet types
+ */
+export function getSupportedSoftwareWallets(connectionDetails: Support<DappConnectionMethodDetails> | null): SoftwareWalletType[] {
```
polymutex: For this one, I suggest creating a new `isSoftwareWalletType` function like

```typescript
function isSoftwareWalletType(maybeWalletType: DappConnectionMethod | SoftwareWalletType): maybeWalletType is SoftwareWalletType {
  switch(maybeWalletType) {
    case DappConnectionMethod.PROPRIETARY_CLOSED_SOURCE_APP:
      return false;
    case DappConnectionMethod.PROPRIETARY_OPEN_SOURCE_APP:
      return false;
    default:
      return true;
  }
}
```

Then this function can become

```typescript
export function getSupportedSoftwareWallets(connectionSupport: DappConnectionSupport): SoftwareWalletType[] {
  if (!isSupported(connectionSupport)) {
    return []
  }
  return setItems(connectionSupport.supportedConnections).filter(isSoftwareWalletType).length
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -88,32 +64,28 @@ function noDappSigning(
 			__brand: brand,
 		},
 		details: paragraph(
-			'{{WALLET_NAME}} supports hardware wallets but does not implement effective dApp signing. dApp signing is important for security as it allows users to verify transaction details on their hardware wallet screen before signing.',
+			'{{WALLET_NAME}} implements either zero or very little signing intent clarity. Signature intent clarity is important for security as it allows users to verify transaction details on their wallet screen before signing. Without this, users are at the mercy of the dApp they are interacting with sending them a bad transactions, either because they have a bug, were hacked, or are malicious. Without a signer being able to verify if their transaction is correct, user should not send such a transaction.\nNote: For hardware wallets, the signature/transaction information *must* be visable on the hardware wallet itself. Any data shown on a software wallet component is ignored for hardware wallet ratings.',
```
polymutex: The "Without this, users are at the mercy of the dApp they are interacting with sending them a bad transactions, either because they have a bug, were hacked, or are malicious." part should be explained in the attribute-wide "Why is this important?" section (the `why` field below on line 209), so no need to re-explain it here. This `details` section should only explain the rating the wallet is getting, not why the attribute matters.

polymutex: Instead of `\n`, please use literal newlines (and wrap the string in backticks). Also note that you can use markdown if you change `details: paragraph(...)` to `details: mdParagraph(...)`, which can render proper paragraphs. This will render the `*must*` properly, rather than showing literal `*` characters. Check other attributes for examples.

polymutex: Typo: visable -> visible

polymutex: The "Note: For hardware wallets, the signature/transaction information *must* be visable on the hardware wallet itself. Any data shown on a software wallet component is ignored for hardware wallet ratings." part should be part of the `methodology` field of the attribute, not part of the `details` of the wallet rating itself.

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -123,32 +95,28 @@ function basicDappSigning(
 			__brand: brand,
 		},
 		details: paragraph(
-			`{{WALLET_NAME}} supports hardware wallets with basic dApp signing${supportedWalletsText}, but the implementation does not provide full transparency for all transaction details. dApp signing allows users to verify transaction details on their hardware wallet screen before signing, which is crucial for security.`,
+			`{{WALLET_NAME}} supports basic signing intent clarity, but the implementation does not provide full transparency for all transaction details. Signature intent clarity is important for security as it allows users to verify transaction details on their wallet screen before signing. Without this, users are at the mercy of the dApp they are interacting with sending them a bad transactions, either because they have a bug, were hacked, or are malicious. Without a signer being able to verify if their transaction is correct, user should not send such a transaction.`,
```
polymutex: Similar comments here

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -229,58 +194,58 @@ function singleCalldataDecodingType(calldataDecoding: CalldataDecoding): Calldat
...
 		midSentenceName: null,
-		howIsEvaluated: "How is a wallet's hardware wallet dApp signing support evaluated?",
+		howIsEvaluated: "How is a wallet's signing intent clarity support evaluated?",
 		whatCanWalletDoAboutIts: sentence(
-			'What can {{WALLET_NAME}} do to improve hardware wallet dApp signing support?',
+			'What can {{WALLET_NAME}} do to improve signing intent clarity support?',
 		),
 	},
-	question: sentence('Does the wallet support secure dApp signing with hardware wallets?'),
+	question: sentence('Does the wallet have signing intent clarity?'),
```
polymutex: Suggestion: `When asking the user to sign a transaction, does the wallet make its intent clear?`

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -229,58 +194,58 @@ function singleCalldataDecodingType(calldataDecoding: CalldataDecoding): Calldat
...
-		Without dApp signing, users must trust that the software wallet is displaying the correct
-		transaction details and not manipulating them. With dApp signing, the hardware wallet shows
+		Without signing intent clarity, users must trust that the software wallet is displaying the correct
+		transaction details and not manipulating them. With signing intent clarity, the wallet shows
 		the actual transaction details that will be signed, providing an independent verification
 		mechanism that significantly enhances security.
 		
-		Full dApp signing implementations ensure that all relevant transaction details (recipient
-		address, amount, fees, etc.) are clearly displayed on the hardware wallet screen, EIP-712 message hashes,
+		Full  signing intent clarity implementations ensure that all relevant transaction details (recipient
```
polymutex: Typo: double space between "Full" and "signing"

src/schema/attribute-groups.ts
```diff
@@ -121,7 +125,8 @@ type SecurityValues = Dict<{
 	securityAudits: SecurityAuditsValue
 	scamPrevention: ScamPreventionValue
 	chainVerification: ChainVerificationValue
-	hardwareWalletDappSigning: HardwareWalletDappSigningValue
+	signingIntentClarity: SigningIntentClarityValue
+	dappConnectionSupport: DappConnectionSupportValue
```
polymutex: I agree with your assessment on the PR comments that `dappConnectionSupport` should probably be an ecosystem thing, not a security thing (as it does not impact the security of the wallet).

data/evidence-reference/coinbase/limited-signature-intent.jpg
polymutex, line 1: Hm, I agree we should have this image somewhere, but we don't have good support for displaying this on the site as a reference. Consider filing a new issue to track this. For now, checking the image here in the repo seems fine to me.

data/hardware-wallets/ledger.ts
```diff
@@ -100,14 +102,30 @@ export const ledgerWallet: HardwareWallet = {
...
 			firmware: null,
-			hardwareWalletDappSigning: {
+			signingIntentClarity: {
 				messageSigning: {
 					calldataDecoding: noCalldataDecoding,
 					details:
 						'Ledger provides basic message signing details when using hardware wallets, and complex signatures can be verified by comparing the EIP-712 hashes to their expected outcomes.',
 					messageExtraction: {
 						[DataExtraction.EYES]: true,
+						[DataExtraction.COPY]: false, // Fantastic
```
polymutex: Did you intend to copy/paste the `// Fantastic` comment here? :)

src/schema/attributes/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,391 @@
...
+import { WalletType } from '@/schema/wallet-types'
+import { markdown, mdParagraph, paragraph, sentence } from '@/types/content'
+
+import { exempt, pickWorstRating, unrated } from '../common'
+
+const brand = 'attributes.security.dapp_connection_support'
+
+export type DappConnectionSupportValue = Value & {
+	canConnectToDapps: boolean
+	connectionDetails: Support<DappConnectionMethodDetails> | null
```
polymutex: I think you can replace these two fields with just `connectionDetails: Support<DappConnectionMethodDetails>`. In the case of returning unrated or exempt rating, you can use `unsupported` as value for `connectionDetails`; it won't be shown anyway.

src/schema/attributes/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,391 @@
...
+			supportedSoftwareWallets,
+			__brand: brand,
+		},
+		details: paragraph(
+			hasOnlyClosedSource
+				? '{{WALLET_NAME}} can connect to dApps, but only through its own proprietary closed-source application. This creates vendor lock-in and requires users to trust the wallet provider\'s software without the ability to verify its security. Users cannot use their preferred software wallet or standard protocols like WalletConnect.'
+				: '{{WALLET_NAME}} supports connecting to dApps but with limited options. While functional, the restricted connection methods may limit user choice and flexibility in how they interact with Web3 applications.',
+		),
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should expand its dApp connection options by supporting standard protocols like WalletConnect and enabling connections through popular software wallets. If using a proprietary app, consider open-sourcing it for transparency.',
```
polymutex: I'm not sure we should promote WalletConnect too much here, since it adds another intermediary and censorship vector for transactions. That said, it's true that it would be an improvement in the case of a wallet that otherwise only supports its own app.

src/schema/attributes/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,391 @@
...
+		This attribute only applies to hardware wallets. Software wallets and smart contract 
+		wallets are exempt as they inherently support dApp connections.
+	`),
+	ratingScale: {
+		display: 'pass-fail',
+		exhaustive: true,
+		pass: [
+			exampleRating(
+				paragraph(`
+					The wallet provides excellent dApp connectivity with support for WalletConnect,
```
polymutex: The wallet tracking data doesn't actually check whether WalletConnect is supported, so I don't think we should say this here. Can just say `The wallet provides excellent dApp connectivity, supporting multiple software wallet integrations and its own open-source application.`

src/schema/attributes/security/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,391 @@
...
+		fail: exampleRating(
+			paragraph(`
+				The wallet cannot connect to dApps, severely limiting its functionality in the 
+				Web3 ecosystem.
+			`),
+			noDappConnectionSupport().value,
+		),
+	},
+	evaluate: (features: ResolvedFeatures): Evaluation<DappConnectionSupportValue> => {
+		// Check for ERC-4337 smart wallet
```
polymutex, line 248: Why does this matter for this attribute?

polymutex (changes requested):
Another epic PR.

polymutex:
> This ended up being a bigger change, because as I dug deeper, more unearthed.

Sorry for the deluge of comments as a result :)

> * It made sense to me to separate `signingIntentClarity` from a new attribute of `dappConnectionSupport` specifically for hardware wallets (we have a feature of what HW wallets a software wallet supports, why not the other way around?)

For sure.

> * Adding `signingIntentClarity` to all wallets I think is important, with the feature rating being _only_ what calldata/transaction data is being shown on the device itself (for example, if ledger live showed TX data, but not the device, we would mark that the ledger device does not show TX data)

Agreed.

> * softened the calldata decoding requirements if a wallet allows the user to extract that calldata

Makes sense.

> * it probably makes sense to put the `dappConnectionSupport` into the `ecosystem` bucket over security

Seems right, since it doesn't affect security and it's more about ecosystem integration.

src/schema/features/security/hw-dapp-connection-support.ts
PatrickAlphaC: makes sense!

src/schema/features/security/hw-dapp-connection-support.ts
PatrickAlphaC: makes sense

src/schema/features/security/hw-dapp-connection-support.ts
PatrickAlphaC: much cleaner!

data/hardware-wallets/ledger.ts
PatrickAlphaC: hehe no

src/schema/attributes/security/hw-dapp-connection-support.ts
PatrickAlphaC: maeks sense

src/schema/attributes/security/hw-dapp-connection-support.ts
PatrickAlphaC: Hardware wallets are not `ERC-4337` wallets? Maybe I'm confused?

PatrickAlphaC:
The comments are great! 

Doing the reviews showed me that there are some issues with the current attributes group. https://github.com/walletbeat/walletbeat/issues/223

PatrickAlphaC:
Ok! We've moved the dapp support to the ecosystem column

Screenshot 2025-07-08 at 3 20 07 PM

But it made me think the way we currently are organizing might need a hand: https://github.com/walletbeat/walletbeat/issues/224

https://github.com/walletbeat/walletbeat/issues/223

And signing intent clarity is now live for both software and hardware wallets!

Screenshot 2025-07-08 at 3 24 20 PM

src/schema/features.ts
```diff
@@ -308,9 +312,13 @@ export function resolveFeatures(features: WalletBaseFeatures, variant: Variant):
...
 			hardwareWalletSupport: softwareFeat(features => features.security.hardwareWalletSupport),
-			hardwareWalletDappSigning: hardwareFeat(
-				features => features.security.hardwareWalletDappSigning,
-			),
+			signingIntentClarity: (() => {
+				if (isWalletSoftwareFeatures(features) && features.security.signingIntentClarity) {
+					return softwareFeat(features => features.security.signingIntentClarity)
+				}
+
+				return hardwareFeat(features => features.security.signingIntentClarity)
```
polymutex, line 320: Maybe `signingIntentClarity` should be part of `WalletBaseFeatures`, rather than being repeated in both `WalletSoftwareFeatures` and `WalletHardwareFeatures`? This would also avoid this weird switching here, because you could just use `baseFeat(features => features.security.signingIntentClarity)`.

src/schema/features/ecosystem/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * A record of hardware wallet dApp connection support
+ */
+export type DappConnectionSupport = Support<WithRef<DappConnectionMethodDetails>>
+
+/**
+ * Shorthand for a wallet that cannot connect to dApps
+ */
+export const cannotConnectToDapps: DappConnectionSupport = notSupported
```
polymutex, line 57: Actually, is this used anywhere?
If not, might as well delete it, I think whatever needs to use it could just use `notSupported` directly.

src/schema/attributes/ecosystem/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,347 @@
...
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should implement at least one method of connecting to dApps such as support through popular software wallets.',
+		),
+	}
+}
+
+function limitedDappConnectionSupport(
+	connectionDetails: Supported<DappConnectionMethodDetails>,
+): Evaluation<DappConnectionSupportValue> {
+	const hasOnlyClosedSource =
```
polymutex, line 55: This boolean determines almost all the fields of the evaluation. It's already already computed in the `evaluate` function below. So I suggest having two separate evaluation functions instead, one for the `hasOnlyClosedSource` case and one for the other case, and update `evaluate` to call the appropriate one. No need to recompute it here.

This removes one level of indentation, and makes the wording clearer in `howToImprove` (as the "If using a proprietary app, consider open-sourcing it for transparency." part would only be shown in the case where it's appropriate).

```typescript
function closedSourceOnlyConnectionSupport(connectionDetails: Supported<DappConnectionMethodDetails>): Evaluation<DappConnectionSupportValue> {
  return { id: 'closed_source_only', rating: Rating.PARTIAL, ... };
}

function limitedDappConnectionSupport(connectionDetails: Supported<DappConnectionMethodDetails>): Evaluation<DappConnectionSupportValue> {
  return { id: 'limited_dapp_connection', rating: Rating.PARTIAL, ... };
}

// [...]

// ... in `evaluate`:
if (hasOnlyClosedSource) {
	return closedSourceOnlyConnectionSupport(dappSupport)
}
// [...]
if (totalMethodCount <= 2 && !hasOpenSource) {
	return limitedDappConnectionSupport(dappSupport)
}
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+			calldataExtraction,
+			calldataDecoding,
+			displayedTransactionDetails,
+			__brand: brand,
+		},
+		details: mdParagraph(
+			'{{WALLET_NAME}} full signing intent clarity. All transaction details are clearly displayed on the wallet screen/window for verification before signing, providing maximum security and transparency for users.',
+		),
+		// Include references if provided
+		references: refs.length > 0 ? refs : undefined,
```
polymutex, line 156: This is equivalent to just `references: refs`. But actually I don't think you need this at all (and you can also remove the `refs` parameter of the `fullSigningIntentClarity` function). This is because the `evaluate` function already injects `references` into the returned value in the last few lines of this file:
```typescript
		// Return result with references
		return {
			...result,
			references,
		}
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+				calldataDecoding: null,
+				calldataExtraction: null,
+				displayedTransactionDetails: null,
+			})
+		}
+
+		// Extract references from the wallet signing intent clarity feature
+		const references = features.security.signingIntentClarity?.ref
+			? refs(features.security.signingIntentClarity)
+			: []
```
polymutex, line 319: Please pass the `references` to all possible returned values of this function.
Two ways:
- Either add `references` as an argument to all of the functions that return an `Evaluation<SigningIntentClarityValue>`
- Either augment it here in this function using something like `return { ...evaluation, references }`

src/schema/attributes/security/hw-dapp-connection-support.ts
polymutex: The code already (more comprehensively) filters out hardware wallets just below this check. So checking for ERC-4337 doesn't really add value here.

polymutex (approved):
Looks good, feel free to merge this into `beta` on your own after the comments are resolved!

(Also after making CI happy)

src/schema/attributes/security/signing-intent-clarity.ts
polymutex: Actually this is already taken care of, whoops. Please ignore this comment.

maykelxyz:
Continuation here #352

polymutex:
Closing this PR, as it was continued and eventually merged by @maykelxyz in #352.

PR #238, opened 2025-07-25
Update ambire data

JIOjosBG:
# Ambire data update

## Updates:
- multi-address data collection
- monetization info

## Remove
- jiffyscan from entities that collect user IPs
- keystone HW support (incorrectly added in https://github.com/walletbeat/walletbeat/pull/205)

## Add:
- chain abstraction info

### DIFF
--- a/data/software-wallets/ambire.ts
+++ b/data/software-wallets/ambire.ts
@@ -398,7 +398,7 @@ Payouts are handled by the Ambire team directly and are denominated in USD. Howe
 							},
 						}),
 					}),
-					[HardwareWalletType.KEYSTONE]: notSupported
+					[HardwareWalletType.KEYSTONE]: notSupported,
 				},
 			},
 			lightClient: {
--- a/data/software-wallets/ambire.ts
+++ b/data/software-wallets/ambire.ts
@@ -193,7 +193,8 @@ export const ambire: SoftwareWallet = {
 					perChainBalanceViewAcrossMultipleChains: featureSupported,
 					crossChainSumView: notSupported,
 					ref: {
-						url: 'https://www.ambire.com/',
+						label: 'Implementation of token filtering by name',
+						url: 'https://github.com/AmbireTech/extension/blob/main/src/common/modules/dashboard/components/Tokens/Tokens.tsx#L89-L106',
 						explanation: 'Ambire supports filtering by token name.',
 					},
 				}),
@@ -206,7 +207,8 @@ export const ambire: SoftwareWallet = {
 					},
 				}),
 				ref: {
-					url: 'https://ambire.notion.site/Ambire-Wallet-Whitepaper-d502e54caf584fe7a67f9b0a018cd10f',
+					label: 'Implementation of token filtering by name',
+					url: 'https://github.com/AmbireTech/extension/blob/main/src/common/modules/dashboard/components/Tokens/Tokens.tsx#L89-L106',
 					explanation:
 						'Ambire supports filtering by token name and chain, as well as displaying the total balance from the resulting tokens',
 				},
--- a/data/software-wallets/ambire.ts
+++ b/data/software-wallets/ambire.ts
@@ -28,7 +28,6 @@ import { ambireEntity } from '../entities/ambire'
 import { biconomy } from '../entities/biconomy'
 import { github } from '../entities/github'
 import { hunterSecurity } from '../entities/hunter-security'
-import { jiffylabs } from '../entities/jiffyscan'
 import { lifi } from '../entities/lifi'
 import { pashov } from '../entities/pashov-audit-group'
 import { pimlico } from '../entities/pimlico'
@@ -287,18 +286,7 @@ export const ambire: SoftwareWallet = {
 							ref: dataLeakReferences.github,
 							walletAddress: Leak.NEVER,
 						},
-					},
-					{
-						entity: jiffylabs,
-						leaks: {
-							cexAccount: Leak.NEVER,
-							endpoint: RegularEndpoint,
-							ipAddress: Leak.ALWAYS,
-							mempoolTransactions: Leak.NEVER,
-							ref: dataLeakReferences.jiffylabs,
-							walletAddress: Leak.NEVER,
-						},
-					},
+					}
 				],
 				onchain: {},
 			},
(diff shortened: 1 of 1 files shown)


data/software-wallets/ambire.ts
```diff
@@ -28,12 +28,12 @@ import { ambireEntity } from '../entities/ambire'
 import { biconomy } from '../entities/biconomy'
 import { github } from '../entities/github'
 import { hunterSecurity } from '../entities/hunter-security'
-import { jiffylabs } from '../entities/jiffyscan'
 import { lifi } from '../entities/lifi'
 import { pashov } from '../entities/pashov-audit-group'
 import { pimlico } from '../entities/pimlico'
 import { ambireAccountContract } from '../wallet-contracts/ambire-account'
 import { ambireDelegatorContract } from '../wallet-contracts/ambire-delegator'
+import { unsupportedProp } from '@mui/material'
```
polymutex: Please remove this import, as it is not used. (Run `pnpm check:all` to make sure the PR passes CI checks)

data/software-wallets/ambire.ts
```diff
@@ -359,15 +388,6 @@ Payouts are handled by the Ambire team directly and are denominated in USD. Howe
...
 					}),
-					[HardwareWalletType.KEYSTONE]: supported({
-						[HardwareWalletConnection.QR]: supported({
-							ref: {
-								explanation:
-									'Ambire supports Keystone hardware wallets through connector integration (non-native).',
-								url: 'https://www.ambire.com/',
-							},
-						}),
-					}),
```
polymutex, line 370: Rather than removing it, please replace it with `[HardwareWalletType.KEYSTONE]: unsupported` so that it's explicit (if missing, it's unknown whether the hardware wallet is supported)

data/software-wallets/ambire.ts
```diff
@@ -288,17 +328,6 @@ export const ambire: SoftwareWallet = {
...
 						},
 					},
-					{
-						entity: jiffylabs,
-						leaks: {
-							cexAccount: Leak.NEVER,
-							endpoint: RegularEndpoint,
-							ipAddress: Leak.ALWAYS,
-							mempoolTransactions: Leak.NEVER,
-							ref: dataLeakReferences.jiffylabs,
```
polymutex, line 298: Please delete `jiffylabs` from the `dataLeakReferences` object earlier in this file, as it is no longer used.

data/software-wallets/ambire.ts
```diff
@@ -232,7 +272,7 @@ export const ambire: SoftwareWallet = {
 							ipAddress: Leak.ALWAYS,
 							mempoolTransactions: Leak.ALWAYS,
 							multiAddress: {
-								type: MultiAddressPolicy.SINGLE_REQUEST_WITH_MULTIPLE_ADDRESSES,
+								type: MultiAddressPolicy.ACTIVE_ADDRESS_ONLY,
```
polymutex, line 285: Can you provide evidence of this change?

data/software-wallets/ambire.ts
```diff
@@ -175,7 +175,47 @@ export const ambire: SoftwareWallet = {
...
+				usdc: supported({
+					perChainBalanceViewAcrossMultipleChains: featureSupported,
+					crossChainSumView: notSupported,
+					ref: {
+						url: 'https://www.ambire.com/',
+						explanation: 'Ambire supports filtering by token name.',
+					},
+				}),
+				ref: {
+					url: 'https://ambire.notion.site/Ambire-Wallet-Whitepaper-d502e54caf584fe7a67f9b0a018cd10f',
```
polymutex: Is there a better URL here, perhaps a code pointer to the chain balance display UI widget?

(Same for other references; linking to `ambire.com` is OK for features that are shown directly on the Ambire homepage, otherwise please link to the code)

polymutex (changes requested):
Thanks for updating!

data/software-wallets/ambire.ts
JIOjosBG: is link to the codebase sufficient?

polymutex: Short answer: Yes.

Long answer: What I'm wondering about is whether (a) it used to be the case that Ambire bundled multiple addresses in a single RPC but no longer does, or (b) Ambire has always operated under the one-address-at-a-time model and this is just updating it to reflect the truth.

- If (a), then I imagine there's a issue link or PR that could be used as reference to show that this was an active effort in Ambire to ensure that the wallet now operates on a one-address-at-a-time basis.
- If (b), then the best would be a code reference where multiple addresses are involved and which shows that the RPC being made about them is about one address only. There's not always an easy single code reference for this. For Rabby for example, the reference is https://github.com/RabbyHub/Rabby/blob/356ed60957d61d508a89d71c63a33b7474d6b311/src/background/controller/wallet.ts#L1622 which shows that the balance refresh function used in the wallet inherently only takes one address at a time as a string parameter, so it can't contain multiple addresses at a time. It's not conclusive proof (the wallet could bundle addresses into a single RPC as part of other operations than balance refreshes), but it's a signal. (I used `mitmproxy` to analyze the wallet traffic outside of balance refreshes to verify the rest, but that kind of work isn't really convertible into a small reference URL 😅)

polymutex:
(btw, I am adding more EIP-7702 feature data in #239)

data/software-wallets/ambire.ts
JIOjosBG: Afaik this has always been the case, but I was not aware it was covered everywhere in the codebase when i submitted the initial Ambire PR here.

JIOjosBG:
>(btw, I am adding more EIP-7702 feature data in https://github.com/walletbeat/walletbeat/pull/239)

dope lmk if i can help

data/software-wallets/ambire.ts
JIOjosBG: yeah, added more appropriate url

polymutex (approved):

polymutex:
Thanks for this PR!

> > (btw, I am adding more EIP-7702 feature data in #239)
> 
> dope lmk if i can help

Yes, you can check if the data I wrote about Ambire there is accurate :)

JIOjosBG:
> Yes, you can check if the data I wrote about Ambire there is accurate :)

seems accurate

PR #253, opened 2025-08-26
Add Mt Pelerin/Bridge Wallet

Sigri44:
Add Mt Pelerin Company, with Bridge Wallet product


data/entities/mtpelerin.ts
```diff
@@ -0,0 +1,28 @@
...
+	crunchbase: 'https://www.crunchbase.com/organization/mt-pelerin',
+	farcaster: { type: 'NO_FARCASTER_PROFILE' },
+	icon: {
+		extension: 'svg',
+	},
+	jurisdiction: 'Switzerland',
+	privacyPolicy: 'https://www.mtpelerin.com/privacy-policy',
+	repoUrl: 'https://github.com/mtpelerin',
+	linkedin: 'https://www.linkedin.com/company/mt-pelerin/',
+	twitter: 'https://x.com/intent/user?screen_name=mtpelerin',
```
polymutex: I think this should be `https://x.com/mtpelerin`

polymutex: I think this should be `https://x.com/mtpelerin`

data/software-wallets/mtpelerin.ts
```diff
@@ -0,0 +1,127 @@
...
+		iconExtension: 'svg',
+		lastUpdated: '2025-08-26',
+		repoUrl: 'https://github.com/mtpelerin',
+		url: 'https://www.mtpelerin.com/',
+	},
+	features: {
+		accountSupport: {
+			defaultAccountType: AccountType.eoa,
+			eip7702: notSupported,
+			// BIP support is not verified
```
polymutex, line 31: In that case, please either:
- Evaluate it now and update the PR accordingly, or
- Leave it as `notSupported` (with a comment saying it still needs to be evaluated), or
- Set the `accountSupport` field to `null`

data/software-wallets/mtpelerin.ts
```diff
@@ -0,0 +1,127 @@
...
+				'1193': null,
+				'2700': null,
+				'6963': null,
+				ref: null,
+			},
+			walletCall: null,
+		},
+		license: License.PROPRIETARY,
+		monetization: {
+			ref: true,
```
polymutex: This isn't a valid `ref`. Please use a URL that explains the monetization of the wallet or, if there's no such URL, leave this `null`.

polymutex (changes requested):
Thanks for the addition!
Please fix the errors (run `pnpm check:all`). There's a few missing imports and some type errors that need addressing. Let me know if any of them is difficult to fix.

Sigri44:
Hey @polymutex ! Thx for return, think now all was fixed.

data/software-wallets/mtpelerin.ts
polymutex: Please address this comment thread.

polymutex (changes requested):
Hi there, still one unaddressed comment on BIP support. Also, some of the fields/imports aren't sorted; you can run `pnpm fix` to automatically fix this problem. You can also run `pnpm check:all` to verify whether any such problems remain.

data/software-wallets/mtpelerin.ts
Sigri44: > In that case, please either:
> 
> * Evaluate it now and update the PR accordingly, or
> * Leave it as `notSupported` (with a comment saying it still needs to be evaluated), or
> * Set the `accountSupport` field to `null`

eip7702 is notSupported for our wallet.
Wallet support BIP32 derivation

polymutex:
Thanks, I will address the lint issues in a follow-up PR.

polymutex (approved):

Issue #93, opened 2025-02-01
Wallet research: MetaMask (browser version)


polymutex:
This issue tracks the work of fully researching data about the MetaMask browser extension under the new wallet schema.

PR #256, opened 2025-09-01
Update MetaMask Data

blaiseup:
Hi Team,

We’d like to update the MetaMask data and raise two points for consideration:

- **Source visibility**: Our codebases are public(visible) but not open source. [[extension](https://github.com/MetaMask/metamask-extension/blob/main/LICENSE)] [[mobile app](https://github.com/MetaMask/metamask-mobile/blob/main/LICENSE)]
If we set the license to `PROPRIETARY`, the _Source visibility_ section fails. We’ve kept it as `UNLICENSED_VISIBLE` to keep the metric passing, but it feels like there should be an intermediate option here.
- **Multi-address privacy**: All of our 3rd-party data partners are `OPT_IN`, but the _Multi-address privacy_ section remains unrated. This may be a gap in the evaluation logic.

Please let us know if any further adjustments are needed on our side. Really appreciate your input!


nkkurt:
@polymutex Could you help us navigate the process to get this PR reviewed/approved? Do we need to do anything other than submitting the PR? Many thanks 🙏

data/software-wallets/metamask.ts
```diff
@@ -55,45 +70,196 @@ export const metamask: SoftwareWallet = {
...
+							[LeakedWalletInfo.WALLET_ASSETS]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_ACTIONS]: Leak.NEVER,
+							[LeakedWalletInfo.MEMPOOL_TRANSACTIONS]: Leak.NEVER,
+							endpoint: RegularEndpoint,
+							multiAddress: {
+								type: MultiAddressPolicy.ACTIVE_ADDRESS_ONLY,
+							},
+						},
+					},
+				],
```
polymutex: This list doesn't seem complete; please list all entities MetaMask contacts here. For example, installing the extension and onboarding, I see requests to:

- `metamask.github.io` (`https://metamask.github.io/ledger-iframe-bridge/9.0.1/` and a bunch of ones to URLs under `/phishing-warning`)
- `api.web3auth.io`: `https://api.web3auth.io/fnd-service/node-details?network=sapphire_mainnet&verifier=auth-connection-id&verifierId=user-id&keyType=secp256k1&sigType=ecdsa-secp256k1`
- `defiadapters.api.cx.metamask.io`

Opening the main browser extension popup, I see requests to:
- `infura.io`: `wss://solana-mainnet.infura.io`, `https://mainnet.infura.io/v3/...`
- `*.metamask.io`: `client-config.api.cx.metamask.io`, `accounts.api.cx.metamask.io`, `defiadapters.api.cx.metamask.io`, `token.api.cx.metamask.io`, `authentication.api.cx.metamask.io`, `oidc.api.cx.metamask.io`, `user-storage.api.cx.metamask.io`. No need to break all of these explicitly, but if there is a specific subset for which certain data is leaked that isn't leaked for other subsets, then those subsets may benefit from being broken down. For example, the user's IP and wallet address is leaked to `accounts.api.cx.metamask.io` but not to `client-config.api.cx.metamask.io` as far as I can tell.

Also note that if I understand correctly, MetaMask defaults the "Batch account balance requests" to enabled by default, so multi-address handling would be `SINGLE_REQUEST_WITH_MULTIPLE_ADDRESSES`.

data/software-wallets/metamask.ts
```diff
@@ -55,45 +70,196 @@ export const metamask: SoftwareWallet = {
...
 		privacy: {
-			dataCollection: null,
+			dataCollection: {
+				collectedByEntities: [
+					{
+						entity: etherscan,
+						leaks: {
+							[LeakedPersonalInfo.IP_ADDRESS]: Leak.OPT_IN,
+							[LeakedWalletInfo.WALLET_ADDRESS]: Leak.OPT_IN,
+							[LeakedWalletInfo.WALLET_BALANCE]: Leak.NEVER,
```
polymutex: Note that `Leak.NEVER` entries don't need to be explicitly listed.

data/software-wallets/metamask.ts
```diff
@@ -55,45 +70,196 @@ export const metamask: SoftwareWallet = {
...
+							endpoint: RegularEndpoint,
+							multiAddress: {
+								type: MultiAddressPolicy.ACTIVE_ADDRESS_ONLY,
+							},
+						},
+					},
+					{
+						entity: segment,
+						leaks: {
+							[LeakedPersonalInfo.IP_ADDRESS]: Leak.OPT_IN,
```
polymutex: I'm not sure if segment is what's behind the "Help us improve MetaMask" choice during onboarding, but if it is, then this should be `Leak.PROMPTED`, not `Leak.OPT_IN`, since the user is asked to make an explicit choice as part of being able to use the wallet.

(`Leak.OPT_IN` means that the user is not actively asked about this, and that the wallet does not make any request to this by default; requests are only made if the user goes out of their way to do some specific action or enables an explicit non-default setting. Let me know if the documentation is not clear...)

data/software-wallets/metamask.ts
```diff
@@ -55,45 +70,196 @@ export const metamask: SoftwareWallet = {
...
+							[LeakedWalletInfo.WALLET_ADDRESS]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_BALANCE]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_ASSETS]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_ACTIONS]: Leak.NEVER,
```
polymutex: If not wallet action data, then what information is sent to Segment?

data/software-wallets/metamask.ts
```diff
@@ -55,45 +70,196 @@ export const metamask: SoftwareWallet = {
...
+						},
+					},
+					{
+						entity: sentry,
+						leaks: {
+							[LeakedPersonalInfo.IP_ADDRESS]: Leak.OPT_IN,
+							[LeakedWalletInfo.WALLET_ADDRESS]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_BALANCE]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_ASSETS]: Leak.NEVER,
+							[LeakedWalletInfo.WALLET_ACTIONS]: Leak.NEVER,
```
polymutex: I'd also assume that Sentry logs would contain information on the wallet actions that the user has performed.

data/software-wallets/metamask.ts
```diff
@@ -185,13 +350,60 @@ export const metamask: SoftwareWallet = {
...
+					contractRegistry: true,
+					leaksContractAddress: false,
+					leaksUserAddress: false,
+					leaksUserIp: false,
+					previousContractInteractionWarning: true,
+					recentContractWarning: true,
+					ref: [
+						{
+							explanation:
+								'MetaMask uses Blockaid integration to provide privacy-preserving transaction simulation and malicious contract detection without sharing transaction details with third parties.',
```
polymutex: The blog post doesn't go into the details, but I'd assume that transaction data is still sent to Infura (but not down to Blockaid), right? Walletbeat uses "third party" to mean "any party that the user doesn't have control over", so Infura is a third party from that perspective. So I suggest the following wording: `MetaMask uses Blockaid integration to provide transaction simulation and malicious contract detection. Transaction details are shared with Infura, but no other party.`
(And then ensure that the Infura entry in `privacy.dataCollection` has `MEMPOOL_TRANSACTIONS` set to `Leak.BY_DEFAULT`, assuming this feature is enabled by default.)

data/entities/sentry.ts
```diff
@@ -0,0 +1,26 @@
+import type { CorporateEntity, OffchainDataProvider } from '@/schema/entity'
+
+export const sentry: CorporateEntity & OffchainDataProvider = {
```
polymutex: Not sure this is really an `OffchainDataProvider`; that name is used to mean things like offchain ENS resolvers, offchain price data oracles, etc. which the wallet queries for information. Here, Sentry is an analytics company, so they do not provide any data to the user of the wallet.

data/entities/sentry.ts
```diff
@@ -0,0 +1,26 @@
...
+	id: 'sentry',
+	name: 'Sentry',
+	legalName: { name: 'Functional Software, Inc.', soundsDifferent: true },
+	type: {
+		chainDataProvider: false,
+		corporate: true,
+		dataBroker: false,
+		exchange: false,
+		offchainDataProvider: true,
```
polymutex: Set to `false`

polymutex (changes requested):
Thanks! I had actually started to review MetaMask a week ago, happy to have some of the work already done :)

There's some CI error that you can fix by running `pnpm fix`.

I also want to specifically say thanks for providing really good reference URLs throughout; other wallets don't have as much to back up their feature claim.
Also, despite the comments below on privacy and data collection, it's good that MetaMask is still ahead of most wallets on this with things like explicitly asking for analytics and letting users customize their RPC endpoints before requests are made. Most other wallets don't bother with that.

Noted on the source visibility thing; I will add a way to be able to express the use of a proprietary source-visible license.

nkkurt:
Thank you for the quick review and the feedback @polymutex  - we'll get back to you with an update shortly.

polymutex:
Updated the MetaMask license in #257, please rebase and remove the use of `License.UNLICENSED_VISIBLE` in `metamask.ts`.

data/software-wallets/metamask.ts
polymutex: FYI, b8f9e37a60e2178bdf030fe90d6330f641600886 refactors this structure a bit to make it clear what data is collected for what purpose, and as part of which flow. Check `rabby.ts`, `daimo.ts`, and `ambire.ts` for examples

data/software-wallets/metamask.ts
blaiseup: @polymutex Infura is part of Consensys, so I believe we treat it as first-party.

polymutex: I understand that perspective, but it's not the one Walletbeat takes. Walletbeat the perspective that anything that isn't running locally on the user's machine is a third party, which includes Infura and Consensys. It is a bit like a web browser: Google Chrome may be made by Google, but you wouldn't expect it to contact `google.com` every time you visit a website to verify whether what you're browsing is a scam site.

Walletbeat attempts to provide users with a similar degree of assurance over the safekeeping of their personal data and browsing habits. This is following guidance in this [Vitalik blog post](https://vitalik.eth.limo/general/2024/12/03/wallets.html) (see "Secure chain access" section) and [parts 5/6/7 of the L1 privacy roadmap](https://ethereum-magicians.org/t/a-maximally-simple-l1-privacy-roadmap/23459), suggesting the transition for wallets to rely on a either TEEs/PIR/mixnets to access external services in order to avoid creating a data trail of the user's activity in any external database. I realize much of the infrastructure for this isn't widespread yet (with some exceptions like 1RPC for accessing chain data in a TEE, and Brume wallet to access chain data over Tor), so most wallets score poorly on this, but just wanted to illuminate the rationale behind this.

Perhaps this is just semantics and the wording Walletbeat uses should be changed to "non-local" or "external" to make this clearer :)

polymutex: FYI, changing the wording from "third party" to "external" in #269.
Also [wrote a guide on how to collect this data with `mitmproxy`](https://github.com/walletbeat/walletbeat/blob/beta/docs/mitmproxy-guide/mitmproxy-guide.md) (though feel free to use any other method that works with similar comprehensiveness)

blaiseup:
Thanks @polymutex for the rewording and for the guide you shared! I’ll go through it and hopefully we can follow up with a more complete PR soon.

Would it be fine to merge the current PR as is, and then submit a second one once we’ve finalized the data collection assessment?

data/software-wallets/metamask.ts
```diff
@@ -203,13 +269,60 @@ export const metamask: SoftwareWallet = {
...
+						},
+						{
+							explanation:
+								'Privacy preserving registry of malicious URLs implemented through eth-phishing-detect.',
+							url: 'https://github.com/MetaMask/eth-phishing-detect',
+						},
+					],
+				}),
+				sendTransactionWarning: supported({
+					leaksRecipient: true,
```
polymutex, line 312: I think this should be `leaksRecipient: false`. If there are no external network lookups then there is no leak.

data/software-wallets/metamask.ts
```diff
@@ -203,13 +269,60 @@ export const metamask: SoftwareWallet = {
 					variantsScope: 'ALL_VARIANTS',
 				},
 			],
-			scamAlerts: null,
+			scamAlerts: {
+				contractTransactionWarning: supported({
+					contractRegistry: true,
+					leaksContractAddress: false,
+					leaksUserAddress: false,
+					leaksUserIp: false,
```
polymutex: As per earlier comment, "third party" here refers to any external provider, which includes Consensys. So I think all the above (`leaksContractAddress`, `leaksUserAddress`, `leaksUserIp`) should be set to `true`.

That said, I acknowledge that this would cause the rating to fail to recognize the efforts you've gone to in order to ensure that the data isn't shared beyond Consensys itself, and that there's a meaningful difference between the typical Blockaid-using wallet and how MetaMask does this. For this reason, it's appropriate to use an `override` here to add a note explaining the distinction. You can add this at the bottom of the file (you can adjust the wording):

```typescript
	overrides: {
		attributes: {
			security: {
				scamPrevention: {
					note: mdParagraph(`
						While MetaMask integrates with Blockaid for transaction simulation and
						scam detection, it does so in a
						[unique way that does not share transaction data beyond Consensys](https://metamask.io/news/metamask-security-alerts-by-blockaid-the-new-normal-for-a-safer-transaction).
					`),
				},
			},
		},
	},
```

`mdParagraph` will cause the note to be rendered as markdown in the relevant attribute (in this case, the Scam Prevention attribute) with leading whitespace trimmed. Check `daimo.ts` for an example of how to use `overrides`. You can use `pnpm dev` to preview the website locally and go to `/metamask` to see how the note gets rendered.

btw, I believe the only way to offer full privacy-preserving transaction simulation support would be to either:

- Only ever offer simulation when it's using a user's own node (which I realize would be a large UX and security regression in practice for most users)
- Run the simulation engine in a TEE/secure enclave that does not leak the transaction data beyond the confines of the enclave, and where the client (wallet) verifies that the enclave is running such code.

So I don't think any wallet would qualify today.

data/software-wallets/metamask.ts
```diff
@@ -203,13 +269,60 @@ export const metamask: SoftwareWallet = {
...
+					previousContractInteractionWarning: true,
+					recentContractWarning: true,
+					ref: [
+						{
+							explanation:
+								'MetaMask uses Blockaid integration to provide privacy-preserving transaction simulation and malicious contract detection without sharing transaction details with external parties.',
```
polymutex: "with external parties" here should be changed to "with parties other than Consensys" or "beyond Consensys", since Consensys *is* an external party.

data/software-wallets/metamask.ts
```diff
@@ -94,30 +160,31 @@ export const metamask: SoftwareWallet = {
...
+		multiAddress: featureSupported,
 		privacy: {
 			dataCollection: null,
-			privacyPolicy: 'https://consensys.io/privacy-notice',
-			transactionPrivacy: {
-				defaultFungibleTokenTransferMode: 'PUBLIC',
-				[PrivateTransferTechnology.STEALTH_ADDRESSES]: notSupported,
-				[PrivateTransferTechnology.TORNADO_CASH_NOVA]: notSupported,
-				[PrivateTransferTechnology.PRIVACY_POOLS]: notSupported,
-			},
```
polymutex, line 120: I believe this `transactionPrivacy` data was correct; any particular reason to change it back to `null`?

data/software-wallets/metamask.ts
```diff
@@ -56,25 +60,87 @@ export const metamask: SoftwareWallet = {
...
+					{
+						explanation:
+							'MetaMask browser extension supports standard Ethereum Provider APIs. Mobile app does not inject into web pages.',
+						url: 'https://support.metamask.io/hc/en-us/articles/360015489471',
+					},
+				],
 			},
-			walletCall: null,
+			walletCall: supported({
+				atomicMultiTransactions: featureSupported,
```
polymutex, line 143: Nice :)

data/software-wallets/metamask.ts
```diff
@@ -56,25 +60,87 @@ export const metamask: SoftwareWallet = {
...
+					risksExplained: 'NOT_IN_UI',
+				}),
+				suggestedBridging: notSupported,
+			},
+			crossChainBalances: {
+				ether: {
+					crossChainSumView: notSupported,
+					perChainBalanceViewAcrossMultipleChains: featureSupported,
+				},
+				// MetaMask supports aggregated balance across popular networks, but custom networks are not included
```
polymutex, line 100: (This is OK)

data/software-wallets/metamask.ts
```diff
@@ -56,25 +60,87 @@ export const metamask: SoftwareWallet = {
...
 			chainSpecificAddressing: {
-				erc7828: null,
-				erc7831: null,
+				erc7828: notSupported,
+				erc7831: notSupported,
+			},
+			nonChainSpecificEnsResolution: {
+				medium: 'CHAIN_CLIENT',
+				support: 'SUPPORTED',
 			},
```
polymutex: Please replace with:

```typescript
nonChainSpecificEnsResolution: supported({
    medium: 'CHAIN_CLIENT',
})
```

data/software-wallets/metamask.ts
```diff
@@ -26,13 +30,13 @@ export const metamask: SoftwareWallet = {
...
 		blurb: paragraph(`
-			MetaMask is a popular Ethereum wallet created by Consensys and that has
+			MetaMask is a popular Multichain wallet created by Consensys and that has
 			been around for a long time. It is a jack-of-all-trades wallet that can
 			be extended through MetaMask Snaps.
 		`),
 		contributors: [polymutex, nconsigny],
 		iconExtension: 'svg',
-		lastUpdated: '2025-02-08',
+		lastUpdated: '2025-09-01',
```
polymutex: Feel free to update this to a more recent date now :)

data/software-wallets/metamask.ts
```diff
@@ -26,13 +30,13 @@ export const metamask: SoftwareWallet = {
 		displayName: 'MetaMask',
 		tableName: 'MetaMask',
 		blurb: paragraph(`
-			MetaMask is a popular Ethereum wallet created by Consensys and that has
+			MetaMask is a popular Multichain wallet created by Consensys and that has
```
polymutex: Uppercase-M "Multichain" appears to be a [project named such](https://multichain.org/), so I suggest using lowercase-m "multichain" here instead.

You will probably also need to change the `Multichain` entry to lowercase `multichain` in `.cspell.json`, as uppercase words in `.cspell.json` are interpreted as required casing for proper nouns.

polymutex:
> Would it be fine to merge the current PR as is, and then submit a second one once we’ve finalized the data collection assessment?

Yes, for sure. It's possible to leave the `features.privacy.dataCollection` field as `null` (which stands for "we don't have this data yet" across all feature fields), leave the partial data for it below and `/* commented out like this */`, and come back to it in a later PR. It looks like you've already done that, so looks all good.

polymutex (changes requested):
Sorry, there are CI issues but they are my fault. I will fix them today.

This PR is looking very good, just minor comments left.

polymutex:
> Sorry, there are CI issues but they are my fault. I will fix them today.

CI issues should be fixed if you rebase on the `beta` branch now.

blaiseup:
Thanks for the detailed feedback @polymutex! I’ve just updated the PR and re-requested your review.

polymutex (approved):
Looks great, thank you so much!

polymutex:
This is merged and visible on https://beta.walletbeat.eth.limo now. Thanks again!

Let me know if you plan to follow up with the rest of the data, otherwise I'm happy to research this on my own (was already planning to do this as part of #93)

blaiseup:
Thanks @polymutex! Our timeline is uncertain, if it's already in your plan, we'd appreciate you taking the lead on it.

PR #277, opened 2025-10-09
Minor stylistic fixes on and fixed URL on "About" page

sideris:
- Fixed Farcaster URL that was pointing to GH
- Added underline to all links to be consistent
- Fixed light theme external link underline (was not visible)
- Changed navigation icons. All icons from "react-icons/lu" so they have the same size

### DIFF
--- a/src/components/navigation.tsx
+++ b/src/components/navigation.tsx
@@ -1,21 +1,25 @@
-import ForumIcon from '@mui/icons-material/Forum'
-import HelpCenterIcon from '@mui/icons-material/HelpCenter'
-import HomeIcon from '@mui/icons-material/Home'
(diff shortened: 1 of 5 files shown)


polymutex:
Thanks for the contribution!

Looks like CI is failing with:

```
src/ui/atoms/ExternalLink.tsx:24:9 - warning ts(6133): 'hovered' is declared but its value is never read.

24  const [hovered, setHovered] = useState(false)
           ~~~~~~~
```

Please fix, I think this is because the link style no longer depends on hover state. You can run `pnpm check:all` to run these checks locally.

sideris:
Great thanks! I believe I addressed the issue (I can't see pipeline status).

polymutex (approved):

polymutex:
Thanks!

PR #312, opened 2025-10-30
Replace use of `undefined` with `null` in `Score`.

polymutex:
Forces explicit handling in functions like `defaultRatingScore`.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,9 +1,10 @@
 {
 	"recommendations": [
-		"astro-build.astro-vscode",
-		"streetsidesoftware.code-spell-checker",
 		"dbaeumer.vscode-eslint",
-		"rvest.vs-code-prettier-eslint"
+		"astro-build.astro-vscode",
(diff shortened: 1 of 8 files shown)


darrylyeo:
What was wrong with `undefined` / `?.`?

polymutex:
> What was wrong with `undefined` / `?.`?

@darrylyeo Maybe that's just me, but I saw this:

```typescript
export function defaultRatingScore(rating: Rating): Score {
	switch (rating) {
		case Rating.FAIL:
			return 0.0
		case Rating.PARTIAL:
			return 0.5
		case Rating.PASS:
			return 1.0
	}
}
```

and I was like "there's no way this typechecks", but it does when `undefined` is part of the `Score` type definition, because it's what implicitly returned when no `return` statement is reached. `null` doesn't have this problem (`return` must be reached or the function won't compile), so it forces being explicit about all the cases, and I like not accidentally relying on implicit behavior.

That said I agree this ended up having larger implications across the rest of the codebase than I anticipated...

PR #335, opened 2025-11-02
Restore import grouping in Svelte files.

polymutex:


darrylyeo:
Can we also have two empty lines between groups? (Had this convention in Astro files too)

src/views/WalletPage.svelte
```diff
@@ -1,25 +1,7 @@
...
-	import { allHardwareModels } from '@/data/hardware-wallets'
-	import { allRatedWallets, type WalletName } from '@/data/wallets'
-	import {
-		attributeTree,
-		calculateAttributeGroupScore,
-		calculateOverallScore,
-	} from '@/schema/attribute-groups'
-	import {
+import { allRatedWallets, type WalletName } from '@/data/wallets'
+import {
```
darrylyeo: Need indents here

darrylyeo (approved):

Issue #336, opened 2025-11-03
Issue: Using same references lead to hydration error


maykelxyz:
Using multiple references with same links lead to hydration issue because of duplicate keys

<img width="873" height="358" alt="Image" src="https://github.com/user-attachments/assets/12f4d554-1d3c-4957-87be-26acf11f0c73" />

<img width="918" height="638" alt="Image" src="https://github.com/user-attachments/assets/9298ac30-3354-497d-af37-43b33d9e8d1c" />

PR #337, opened 2025-11-03
Fix: Resolve duplicate key using index

maykelxyz:
Fixes `each_key_duplicate` error when same URL appears in multiple references. Changed key from URL-based to index-based


maykelxyz:
Fixes #336

polymutex (approved):

Issue #230, opened 2025-07-16
Wallet attribute: Bug Bounty (revamp)


polymutex:
This issue is about revamping the bug bounty feature data and corresponding attribute. It is currently based on a 4-state enum where the values are pretty subjective.

```typescript
/**
 * Types of bug bounty programs that can be implemented
 */
export enum BugBountyProgramType {
	COMPREHENSIVE = 'COMPREHENSIVE',
	BASIC = 'BASIC',
	DISCLOSURE_ONLY = 'DISCLOSURE_ONLY',
	NONE = 'NONE',
}
```

Goal: Make it more objective and define where the thresholds are more concretely.

polymutex:
Relevant links:

- https://github.com/walletbeat/walletbeat/blob/fbc836ee8c439819c2908ec8ab46d5b4d88734ae/src/schema/features/security/bug-bounty-program.ts
- https://github.com/walletbeat/walletbeat/blob/fbc836ee8c439819c2908ec8ab46d5b4d88734ae/data/hardware-wallets/trezor.ts#L83
- https://github.com/walletbeat/walletbeat/blob/fbc836ee8c439819c2908ec8ab46d5b4d88734ae/src/schema/attributes/security/bug-bounty-program.ts

polymutex:
Ideas of objective things we can look for in a bug bounty program:

- When was the program established? (Allows for measuring duration and consistency of how seriously the wallet developer takes security)
- Scope: Do they explicitly list it? If so, does it cover all versions of the wallet?
- Disclosure process: does it have a fixed number of days, or is it up to the discretion of the wallet developer?
- Disclosure process: does it allow security researchers to eventually publish their findings?
- What range of money is the wallet developer offering as bounties?
- Does it have Safe Harbor legal language to protect security researchers from being sued by disclosing? If so, require reference. ([Example](https://hackerone.com/anthropic-vdp/safe_harbor))
- Is it hosted on a popular security bounty platform (HackerOne/Bugcrowd/Intigriti/etc.)? If so, require a link to them.

We could potentially include response metrics (like HackerOne's average time-to-response, time-to-triage, time-to-resolve) but I worry this may be difficult to keep up-to-date without additional automation, and would be difficult for new wallet development teams to get off zero since they are new, so we can't really penalize them. Perhaps we could still collect a set of `{ vulnerabilitiesDisclosed: number; date: Date }` snapshots, and only start enforcing this as an attribute after the program has been running for over a year or so.

[...]

PR #338, opened 2025-11-04
Refactor: Bug bounty program

maykelxyz:
# Refactor: Bug Bounty Program Schema

## Overview
Refactored the bug bounty program type definitions to provide granular, objective criteria for evaluating wallet security programs.

### Replaced subjective categories with measurable attributes
**Before:** Used broad, subjective classifications (`COMPREHENSIVE`, `BASIC`, `DISCLOSURE_ONLY`, `NONE`)  
**After:** Introduced specific, verifiable fields that capture concrete program details

### New Type Definitions

**`BugBountyPlatform`** - Tracks which platform hosts the program (Self-hosted, HackerOne, Bugcrowd, etc.)

**`LegalProtectionType`** - Distinguishes between Safe Harbor clauses and other legal assurances for researchers

**`BugBountyProgramAvailability`** - Indicates current program status (`ACTIVE`, `INACTIVE`)

**`CoverageBreadth`** - Specifies what's covered: app, firmware, hardware, or full scope

**`BugBountyProgramSupport`** - Comprehensive program details including:
- Program start date
- Current availability status
- Coverage scope (with type-safe "at least one" coverage requirement)
- Reward ranges with currency
- Platform information
- Disclosure timeline
- Upgrade path availability
- Legal protections for researchers

## Benefits

- **Objective evaluation**: Enables consistent, measurable comparisons across wallets
- **Better transparency**: Makes security program details explicit and verifiable
- **Reference support**: Uses `WithRef` and `MustRef` types for proper documentation linkage

## Migration Notes

Existing wallets using the old enum-based system will need data migration to populate the new structured fields.

### Wallet Updates for new structured fields
- Bitbox
- Cypherock
- Firefly
- Gridplus
- Keystone
- Ledger
- Ngrave
- Onekey
- Trezor
### Wallet Updates for new structured fields
- Ambire
  - Updated Ambire's bug bounty program to be inactive. [Reference](https://blog.ambire.com/ambire-x-immunefy-bug-bounty-audit-our-code-and-earn-rewards/) 
- ImToken
- MtPelerin

Issue #230


src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,12 +1,67 @@
...
+export interface LegalProtection {
+	/**
+	 * The type of legal protection provided
+	 */
+	type: LegalProtectionType
+
+	/**
+	 * URL or reference to the legal protection language
+	 */
+	reference?: string
```
polymutex: Instead of URL fields, please use the `WithRef` wrapper type instead in general (check how it's used in other feature types). `WithRef` standardizes the way reference URLs are expressed across the wallet data and makes it easier to display them consistently in the UI later.

In this case however, if you implement `legalProtection: Support<LegalProtection>` as per my other comment suggestion, then any remaining value of `LegalProtection.type` must come with a URL. In such a case, you can use `MustRef` instead of `WithRef`, which forces specifying a URL (whereas it is optional with `WithRef`.

So something like this:

```typescript

export enum LegalProtectionType {
	SAFE_HARBOR = 'SAFE_HARBOR',
	LEGAL_ASSURANCE = 'LEGAL_ASSURANCE',
}

export type LegalProtection = MustRef<{
  type: LegalProtectionType
  // ... other fields here...
}>

// [...]

export interface BugBountyProgramSupport {
  // [... other fields...]

  legalProtections: Support<LegalProtection>
}
```

This will force either `legalProtections: unsupported`, or `legalProtections: supported({ ... })` and require a `ref` URL to be specified inside the `supported({ ... })` block.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,12 +1,67 @@
...
+
+	/**
+	 * Specific details or excerpt of the legal protection language
+	 */
+	details?: string
+
+	/**
+	 * Whether they participate in #legalbugbounty or similar standardization
+	 */
+	standardizedLanguage?: boolean
```
polymutex: Is this referring to [this project](https://github.com/EdOverflow/legal-bug-bounty)? It doesn't look super standard. IMO we should remove this, and if we find a wallet with a bug bounty that uses this project we can add it at that time (possibly as a different `LegalProtectionType` enum value)

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,12 +1,67 @@
...
+ * The coverage breadth of the bug bounty program
+ */
+export enum CoverageBreadth {
+	FULL = 'FULL',
+	PARTIAL = 'PARTIAL',
+	APP_ONLY = 'APP_ONLY',
+	FIRMWARE_ONLY = 'FIRMWARE_ONLY',
+	HARDWARE_ONLY = 'HARDWARE_ONLY',
 	NONE = 'NONE',
 }
```
polymutex, line 61: As with other feature data, I recommend keeping this set small to the set of options we're sure are needed, in order not to make the attribute logic more complicated than it needs to be. It's easy to add new feature data later, harder to remove them from consideration.

In this case for example, the `NONE` option seems useless: what kind of bug bounty program covers nothing at all?
It's also not clear what the difference is between "PARTIAL" and "APP_ONLY" (aren't "APP_ONLY"/"FIRMWARE_ONLY"/"HARDWARE_ONLY" all by definition also partial?).

IMO this enum can either be deleted (and the implicit value is that everything is covered by the bug bounty program), or could be replaced by something like `coverageBreadth: 'ALL_VARIANTS' | AtLeastOneVariant`, which allows selecting by Variant instead. Look at the `SecurityAudit.variantsScope` field in `src/schema/features/security/security-audits.ts` as an example.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
  */
 export interface BugBountyProgramSupport {
 	/**
-	 * The type of bug bounty program implemented
+	 * The date the bug bounty program started
+	 */
+	dateStarted?: Date
```
polymutex: All feature data in general must not be `undefined` (therefore, field names should never be followed by `?`). This is a general style we've followed (there's actually a test for it at `deploy/misc-tests/no_undefined_in_wallet_features.sh` though it only actually checks `src/schema/features.ts` so it didn't catch this). The rationale is that allowing `undefined` makes it difficult to look at a wallet data file and figure out what's missing and still needs to be filled out. So every field is required. For fields where we don't have any data yet, we can put `null`. This way, searching for `null` in a wallet file gives a comprehensive view of all the fields we still need data for.

(Yes I know there's an exception to this in `src/schema/features/security/security-audits.ts` right now; maybe for another PR to clean that up...)

Anyway, I don't think this field should be `dateStarted: Date | null` either. IMO, data entry for the bug bounty program should be all-or-nothing for simplicity. In other words, either we have full feature data for a wallet's `BugBountyProgramSupport`, or we have nothing. The `VariantFeature` wrapper type actually already takes care of this (look at `WalletBaseFeatures.security.bugBountyProgram`) as it already makes it `null`-able.

So for the purpose of `BugBountyProgramSupport`, every field should be non-null, non-undefined. I think that should keep things simpler to handle in the attribute evaluation logic.

Also, specifically for `dateStarted`, please use the `CalendarDate` type (from `src/types/date.ts`) as it is easier to input than JavaScript dates.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
+	/**
+	 * The maximum reward for a bug bounty program
+	 */
+	maximumReward?: number
+
+	/**
+	 * The minimum reward for a bug bounty program
 	 */
-	type: BugBountyProgramType
+	minimumReward?: number
```
polymutex: I suggest replacing this with:

```typescript
rewards: Support<{
     minimum: number
     maximum: number
     currency: string
   }>
```

The `Support` wrapper type is useful for data that only makes sense to populate in case that a wallet supports something (in this case, if its bug bounty program supports rewards). This way it can be written like this in wallet data files:

```typescript
rewards: supported({
  minimum: 1000,
  maximum: 10000,
  currency: 'USD',
})
```

or:

```typescript
rewards: notSupported
```

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
 
 	/**
 	 * URL to the bug bounty program details
 	 */
 	url?: string
```
polymutex: Remove field and wrap in `WithRef` instead

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
 
+	/**
+	 * The platform of the bug bounty program
+	 */
+	platform?: BugBountyPlatform
+
+	/**
+	 * The URL to the bug bounty program
+	 */
+	platformUrl?: string
```
polymutex: I think this is probably going to be the same as the other URL already part of this type, so no need for another one here.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
+
 	/**
 	 * Additional details about the bug bounty program implementation
 	 */
 	details?: string
```
polymutex: No longer needed; the new fields being added in this PR should remove the need to have a freeform `details` field.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
 
+	/**
+	 * The disclosure process of the bug bounty program;
+	 * number of days, privacy of reports, etc.
+	 */
+	disclosureProcess?: string
```
polymutex: The goal of this PR is to make this not freeform, so let's either remove this, or keep it but structured, like:

```typescript
disclosure: Support<{ numberOfDays: number }>
```

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,12 +1,67 @@
...
+}
+
+/**
+ * Types of legal protection provided to security researchers
+ */
+export enum LegalProtectionType {
+	SAFE_HARBOR = 'SAFE_HARBOR', // Explicit Safe Harbor language
+	LEGAL_ASSURANCE = 'LEGAL_ASSURANCE', // Pledges not to sue, similar protections
+	NONE = 'NONE', // No legal protections mentioned
+}
```
polymutex, line 26: Please put these comments as `/** JSDoc comments */` above each enum value instead, so that they show up in type documentation tooltips in IDEs.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -15,24 +70,65 @@ export enum BugBountyProgramType {
...
+
 	/**
 	 * Whether the wallet offers an upgrade path for security issues
 	 */
 	upgradePathAvailable: boolean
+
+	/**
+	 * The legal protections offered to security researchers
+	 */
+	legalProtections?: LegalProtection
```
polymutex: `legalProtections: Support<LegalProtection>`

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,12 +1,67 @@
...
+	INTIGRITI = 'Intigriti',
+}
+
+/**
+ * Types of legal protection provided to security researchers
+ */
+export enum LegalProtectionType {
+	SAFE_HARBOR = 'SAFE_HARBOR', // Explicit Safe Harbor language
+	LEGAL_ASSURANCE = 'LEGAL_ASSURANCE', // Pledges not to sue, similar protections
+	NONE = 'NONE', // No legal protections mentioned
```
polymutex: Can remove NONE here if using `legalProtection: Support<LegalProtection>` as per my other comment suggestion.

polymutex (changes requested):
Great direction so far. I only reviewed the feature data (not the attribute logic yet). Overall this is the right approach, just some stylistic/type changes to be more in line with the way other wallet feature data is usually expressed.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,39 +1,114 @@
...
+}>
+
+/**
+ * The availability of the bug bounty program
+ */
+export enum BugBountyProgramAvailability {
+	ACTIVE = 'ACTIVE', // Running now, accepting reports
+	INACTIVE = 'INACTIVE', // Temporarily paused
+	NEVER = 'NEVER', // Never existed
+}
```
polymutex, line 52: Put comments as JSDoc

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,39 +1,114 @@
...
  * Information about the bug bounty program implementation
  */
-export interface BugBountyProgramSupport {
+export type BugBountyProgramSupport = WithRef<{
 	/**
-	 * The type of bug bounty program implemented
+	 * The date the bug bounty program started
 	 */
-	type: BugBountyProgramType
+	dateStarted: Date
```
polymutex: Use CalendarDate rather than Date, it's easier to enter.

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,39 +1,114 @@
...
+	/**
+	 * The platform of the bug bounty program
+	 */
+	platform?: BugBountyPlatform
+
+	/**
+	 * The disclosure process of the bug bounty program;
+	 * number of days, privacy of reports, etc.
+	 */
+	disclouse: Support<{
```
polymutex: Typo: `disclosure`

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,39 +1,114 @@
...
+		minimum: number
+		maximum: number
+		currency: string
+	}>
+
+
+	/**
+	 * The platform of the bug bounty program
+	 */
+	platform?: BugBountyPlatform
```
polymutex: Remove `?`

maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/features/security/bug-bounty-program.ts
maykelxyz: This is noted, and makes more sense.

> Look at the `SecurityAudit.variantsScope` field in `src/schema/features/security/security-audits.ts` as an example.

Checked `src/schema/features/security/security-audits.ts`, and will follow this format. 

```ts
	coverageBreadth: AtleastOneScope | 'FULL_SCOPE'
```

is in WIP

src/schema/features/security/bug-bounty-program.ts
```diff
@@ -1,39 +1,118 @@
...
+/**
+ * The availability of the bug bounty program
+ * Active - Running now, accepting reports
+ * Inactive - Temporarily paused
+ * Never - Never existed
+ */
+export enum BugBountyProgramAvailability {
+	ACTIVE = 'ACTIVE',
+	INACTIVE = 'INACTIVE',
+	NEVER = 'NEVER',
```
polymutex: So, I just realized that `BugBountyProgramAvailability.NEVER` basically means there is no bug bounty program. The other fields in `BugBountyProgramSupport` are required, but make no sense when there is no bug bounty program.

So instead, I suggest doing the following:

- Remove `NEVER` from `BugBountyProgramAvailability` (the only two values are `ACTIVE` and `INACTIVE`)
- Wrap the entire `BugBountyProgramSupport` type with `Support` (i.e. in `features.ts`, make the `security.bugBountyProgram` field be `bugBountyProgram: VariantFeature<Support<BugBountyProgramImplementation>>`)

This way, if a wallet supports a bounty program, it'll be written as:

```typescript
bugBountyProgram: supported<BugBountyProgramImplementation>({
  availability: BugBountyProgramAvailability.ACTIVE,
  // ... all the other BugBountyProgramImplementation fields...
})
```

And if it does not:

```typescript
bugBountyProgram: unsupported
```

And if we don't know:

```typescript
bugBountyProgram: null
```

No redundant fields this way.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +87,120 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
-			shortExplanation: mdSentence(
-				'{{WALLET_NAME}} implements a comprehensive bug bounty program with clear incentives and processes.',
-			),
-			programType: BugBountyProgramType.COMPREHENSIVE,
-			upgradePathAvailable: support.upgradePathAvailable,
-			__brand: brand,
-		},
-		details: markdown(`
-			{{WALLET_NAME}} implements a comprehensive bug bounty program that offers strong incentives for security researchers to find and report vulnerabilities. The program has a wide scope, competitive rewards, and a responsive disclosure process.
+function determineRating(support: BugBountyProgramSupport): Evaluation<BugBountyProgramValue> {
```
polymutex: Attribute logic looks nice in general. Needs some rework to account for the refactored `BugBountyProgramImplementation` type since this PR started. When dealing with `Support<...>` fields, please use the `isSupported` type predicate which will let you access fields within the `<...>` part in a type-safe way (otherwise the type-checker will complain). Check `src/schema/attributes/security/scam-prevention.ts` for an example attribute that uses `isSupported` a bunch.
So for example you'd probably have

```typescript
if (!isSupported(bugBountyProgram)) { // Check for `bugBountyProgram: unsupported`
  return noBugBountyProgram()
}

// You can now access the fields of `bugBountyProgram` safely in the rest of this function, because the type-checker knows the `bugBountyProgram` is `supported` since the `if` above didn't match.
```

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -196,36 +261,31 @@ export const bugBountyProgram: Attribute<BugBountyProgramValue> = {
 		pass: [
 			exampleRating(
 				mdParagraph(`
-					The hardware wallet implements a comprehensive bug bounty program with clear incentives and responsive processes.
-					It offers competitive rewards based on severity, has a transparent disclosure process, and provides upgrade paths for users.
+					The hardware wallet has an active bug bounty program with competitive rewards (e.g., up to $50,000),
```
polymutex: It's fair to use "does it have non-zero rewards" as a criterion, but for now let's not write specific numbers here unless we actually use these exact thresholds in the evaluation criteria (which I don't think we should just yet)

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -196,36 +261,31 @@ export const bugBountyProgram: Attribute<BugBountyProgramValue> = {
...
 					upgradePathAvailable: true,
 				}),
 			),
 		],
 		partial: [
 			exampleRating(
 				mdParagraph(`
-					The hardware wallet implements a basic bug bounty program with limited scope or rewards.
-					However, it does provide a clear upgrade path for users when security issues are discovered.
+					The hardware wallet has a bug bounty program with rewards (e.g., up to $5,000),
```
polymutex: Same, remove number

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +87,120 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
+
+	const standardization = legalProtection.standardizedLanguage
+		? ' The program participates in standardized legal protection initiatives.'
+		: ''
+
+	const referenceLink = legalProtection.reference
+		? ` [View legal protection policy](${legalProtection?.reference})`
+		: ''
+
+	return `Legal Protection: The program provides ${protectionType} protections for security researchers conducting good faith security research.${standardization}${referenceLink}`
```
polymutex: Use either the `refs` or the `popRefs` helper functions to exact `ref`s out of the feature data, and return them as the `references` field of `Evaluation`. This will cause them to be automatically displayed in the wallet page. Check other attributes for examples of these functions.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +87,120 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
+		support.legalProtections?.type != null
+	const isActive = support.availability === BugBountyProgramAvailability.ACTIVE
+
+	const passesAll = isActive && hasFullCoverage && hasRewards && hasLegalProtection
+
+	const rating = passesAll
+		? Rating.PASS
+		: isActive || hasRewards || hasFullCoverage || hasLegalProtection
+			? Rating.PARTIAL
+			: Rating.FAIL
```
polymutex, line 128: Seems like a reasonable approach, I like it!

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -20,11 +24,46 @@ import { exempt, pickWorstRating, unrated } from '../common'
...
+}
+
+function getRewardDescription(support: BugBountyProgramSupport): string {
+	if (support.minimumReward && support.maximumReward) {
+		return `with rewards ranging from $${support.minimumReward.toLocaleString()} to $${support.maximumReward.toLocaleString()}`
+	} else if (support.maximumReward) {
+		return `with rewards up to $${support.maximumReward.toLocaleString()}`
+	} else if (support.minimumReward) {
+		return `with rewards starting at $${support.minimumReward.toLocaleString()}`
+	}
```
polymutex: Add a case here where `minimumReward` and `maximumReward` are both set to the same number (i.e. fixed reward size). Under the current code that would say something like "with rewards ranging from $1000 to $1000", but it should say something like "with a $1000 reward" in the case they are equal.

polymutex (commented):
Great progress, I started looking at the attribute implementation in the comments below.

src/schema/attributes/security/bug-bounty-program.ts
maykelxyz: WIP

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -20,11 +24,53 @@ import { exempt, pickWorstRating, unrated } from '../common'
...
+			return `with rewards ranging from $${min.toLocaleString()} to $${max.toLocaleString()}`
+		}
+	} else if (typeof max === 'number') {
+		return `with rewards up to $${max.toLocaleString()}`
+	} else if (typeof min === 'number') {
+		return `with rewards starting at $${min.toLocaleString()}`
+	}
+
+	return ''
+}
```
polymutex, line 80: Here's how you can do this more simply and without manual `typeof` checks:

```typescript
function getRewardDescription(support: BugBountyProgramSupport): string {
  if (!isSupported(support.rewards)) {
    return ''
  }
  const min = support.rewards.min  // No `?` operator required, clean `number` type already inferred
  const max = support.rewards.max
  // ...
}
```

In general, if you find needing to use `typeof` or `foo?.bar` anytime in attribute code, there's probably a cleaner way to do it :)

maykelxyz: Refactored into:

```ts
function getRewardDescription(support: BugBountyProgramSupport): string {
	if (!isSupported(support.rewards)) return ''

	const { minimum: min, maximum: max } = support.rewards

	if (min != null && max != null) {
		return min === max
			? `with a $${min.toLocaleString()} reward`
			: `with rewards ranging from $${min.toLocaleString()} to $${max.toLocaleString()}`
	}

	if (max != null) return `with rewards up to $${max.toLocaleString()}`
	if (min != null) return `with rewards starting at $${min.toLocaleString()}`
	return ''
}
```

Null checks are for rewards if maximum is given and minimum is not. Please feel free to let me know any feedbacks

src/schema/features/security/bug-bounty-program.ts
maykelxyz:
```ts
/**
 * A set of at least one coverage breadth
 */
export type AtLeastOneCoverageBreadth = NonEmptySet<CoverageBreadth>


	/**
	 * The coverage breadth of the bug bounty program
	 */
	coverageBreadth: AtLeastOneCoverageBreadth | 'FULL_SCOPE'
```

Improve coverageBreadth feature to the code above

src/schema/features/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -7,24 +7,76 @@ import {
...
 
 export type BugBountyProgramValue = Value & {
-	programType: BugBountyProgramType
-	upgradePathAvailable: boolean
+	availability?: BugBountyProgramAvailability
+	coverageBreadth?: CoverageBreadth
+	upgradePathAvailable?: boolean
+	platform?: BugBountyPlatform
+	platformUrl?: string
+	legalProtections?: LegalProtection
```
polymutex: These fields are not read anywhere, so you can remove them.
If you do, you'll need to replace the code below that says

```typescript
unrated(bugBountyProgram, brand, {
				availability: undefined,	
				coverageBreadth: CoverageBreadth.NONE,
				upgradePathAvailable: false,
			})
```

with just

```typescript
unrated(bugBountyProgram, brand, null)
```

src/schema/attributes/security/bug-bounty-program.ts
polymutex: Sounds good!

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
+	const hasRewards =
+		isSupported(support.rewards) &&
+		support.rewards.minimum != null &&	
+		support.rewards.maximum != null &&
+		support.rewards.minimum !== 0 &&
+		support.rewards.maximum !== 0
+	const hasFullCoverage = support.coverageBreadth === CoverageBreadth.FULL
+	const hasLegalProtection =
+		isSupported(support.legalProtections) &&
+		support.legalProtections?.type != null
```
polymutex: Checking `support.legalProtections?.type != null` is no longer needed, since `legalProtections.type` is guaranteed to be non-null by the type system in the supported case now.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
+		: support.availability === BugBountyProgramAvailability.INACTIVE
+			? 'Note that the program is currently inactive and not accepting new reports.'
+			: 'No bug bounty program has been announced or is publicly available.'
+
 	return {
 		value: {
-			id: 'disclosure_only_program',
-			rating: Rating.PARTIAL,
-			displayName: 'Basic disclosure policy',
+			id: 'bug_bounty_available',
```
polymutex: Please make this ID string depend on `isActive`, similar to how it's done for `displayName` below. The ID isn't currently used but is intended to be roughly unique for each type of rating the function can return.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
 			shortExplanation: mdSentence(
-				'{{WALLET_NAME}} implements a basic vulnerability disclosure policy but no formal bounty program.',
+				`{{WALLET_NAME}} has a bug bounty program ${rewardInfo}${isActive ? '' : ', but it is currently inactive'}.`,
 			),
-			programType: BugBountyProgramType.DISCLOSURE_ONLY,
+			availability: support.availability,	
+			coverageBreadth: support.coverageBreadth || CoverageBreadth.NONE,
 			upgradePathAvailable: support.upgradePathAvailable,
+			platform: support.platform,
+			legalProtections: isSupported(support.legalProtections) ? support.legalProtections : undefined,
```
polymutex: Can remove all these fields if removing them from `BugBountyProgramValue`.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
+			
+			${isSupported(support.disclosure) ? `**Disclosure Process**: ${support.disclosure.numberOfDays} days` : ''}
+			
 			${
 				support.upgradePathAvailable
 					? 'Positively, the wallet does provide an upgrade path for users when security issues are identified.'
 					: 'Unfortunately, the wallet does not provide a clear upgrade path for users when security issues are identified.'
 			}
+			
+			${support.platform ? `For more information, visit their [bug bounty program page](${support.platform}).` : ''}
```
polymutex: I don't think `support.platform` is a valid URL, since it's an enum value.
I think you can simply remove this whole line. The `ref`s will be displayed automatically below on the wallet page, so the evaluation description doesn't need to include links.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
 		`),
 		howToImprove: markdown(`
```
polymutex: `howToImprove` should be unset if the evaluation is passing, since there's nothing to improve.

So you can do `howToImprove: passesAll ? undefined : markdown(/* ... */)`, and remove the `${passesAll ? '- Continue maintaining and improving the bug bounty program’s transparency and responsiveness' : ''}` part below.

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
-			${support.upgradePathAvailable ? 'Positively, the wallet provides an upgrade path for users when security issues are identified.' : 'Unfortunately, the wallet does not provide a clear upgrade path for users when security issues are identified.'}
-		`),
-		howToImprove: markdown(`
-			{{WALLET_NAME}} should:
-			- Expand the scope and increase rewards for their bug bounty program
-			${!support.upgradePathAvailable ? '- Establish a clear upgrade path for users when security vulnerabilities are discovered' : ''}
-			- Improve response times and transparency in the vulnerability handling process
-		`),
-	}
+	return `Legal Protection: The program provides ${protectionType} protections for security researchers conducting good faith security research.`
```
polymutex: Maybe add `**` (bold) around the words "Legal Protection".

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -47,98 +100,108 @@ function noBugBountyProgram(): Evaluation<BugBountyProgramValue> {
...
-			}
-		`),
-		howToImprove: support.upgradePathAvailable
-			? undefined
-			: mdParagraph(`
-				{{WALLET_NAME}} should establish a clearer upgrade path for users when security vulnerabilities are discovered, such as offering discounted replacements or firmware updates when possible.
-			`),
+	// No bug bounty program at all
+	if (!hasBugBounty) {
+		return noBugBountyProgram()
```
polymutex: I believe this is no longer possible to be reached. The function that calls `determineRating` already checks that `features.security.bugBountyProgram` is supported, and the type system ensures that `support.availability` is never `undefined`.

So I think you can delete the `determineRating` function, and replace its call below with a direct call to `bugBountyAvailable`.

polymutex (commented):
Almost there! I think this is good to go once the rest of the comments here are addressed.

src/schema/attributes/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/attributes/security/bug-bounty-program.ts
maykelxyz: REsolved

src/schema/attributes/security/bug-bounty-program.ts
maykelxyz: Resolved

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -7,24 +7,74 @@ import {
...
+				return 'only hardware vulnerabilities'
+			default:
+				return ''
+		}
+	}).filter(Boolean)
+	
+	if (descriptions.length === 0) {
+		return ''
+	}
+	return `The program covers ${descriptions.join(', ')}.`
```
polymutex: For two items, this would return "The program covers only the application layer, only firmware vulnerabilities.".

I think you can move the "only" to line 52 and out of the individual cases of the `switch`, so that it says "The program covers only the application layer, firmware vulnerabilities.".

Then you can use `commaListFormat` from `src/types/utils/text.ts` instead of just `join`, which handles the case of using "and" vs commas depending on how many items there are. `The program covers only ${commaListFormat(descriptions)}.` would then return "The program covers only the application layer and firmware vulnerabilities."

src/schema/attributes/security/bug-bounty-program.ts
```diff
@@ -247,41 +311,23 @@ export const bugBountyProgram: Attribute<BugBountyProgramValue> = {
...
+			return unrated(bugBountyProgram, brand, null)
+		}
+
+		if (!isSupported(features.security.bugBountyProgram)) {
+			return noBugBountyProgram()
 		}
 
 		const { withoutRefs, refs } = popRefs<BugBountyProgramSupport>(
 			features.security.bugBountyProgram,
 		)
```
polymutex, line 342: Import the `refs` function from `src/schema/reference.ts`, then change this to:

```typescript
const allRefs = mergeRefs(
  refs(features.security.bugBountyProgram),
  isSupported(features.security.bugBountyProgram.legalProtections) ? refs(features.security.bugBountyProgram.legalProtections) : undefined,
)
```

This way you can catch both references (the one in `features.security.bugBountyProgram` and the one in `features.security.bugBountyProgram.legalProtections`).

Then on line 335 you can replace `...(refs.length > 0 && { references: refs }),` with `...(allRefs.length > 0 && { references: allRefs }),`

polymutex (approved):

polymutex:
PR looks perfect, just need to fix the CI issues (mostly just making the other hardware wallet files have the updated bugBounty fields). Then mark the PR as non-draft, make sure CI passes, and feel free to merge!

src/schema/attributes/security/bug-bounty-program.ts
polymutex: Unresolving this conversation; still need to move the word `only` to the last `return` statement so that it's not repeated in case there's more than one item.

polymutex (approved):

PR #339, opened 2025-11-04
Svelte: type / lint fixes

### LINKED ISSUE #329: UI: Fix lint issues
Adding linting in #328, with a bunch of ignored files to keep CI functional until the code fully lints again.

This issue tracks the removal of the ignored files from the ESLint config.

darrylyeo:
Fixes https://github.com/walletbeat/walletbeat/issues/329.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -58,11 +58,13 @@ export default [
 		languageOptions: {
 			parserOptions: {
 				projectService: true,
-				extraFileExtensions: ['.svelte'],
-				parser: {
-					ts: tseslint.parser,
-					typescript: tseslint.parser,
-					svelte: tseslint.parser,
+				extraFileExtensions: [
+					'.svelte',
+					'.svelte.ts',
+				],
+				parser: tseslint.parser,
+				svelteFeatures: {
+					experimentalGenerics: true,
 				},
 				svelteConfig,
 			},
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -229,6 +229,14 @@ export default [
 
 			// Prevent crashes due to duplicate keys.
 			'svelte/require-each-key': 'warn',
+
+			// ESLint's type checker doesn't perfectly match TypeScript's, especially for Svelte-specific syntax (e.g., {@attach pattern, $bindable inference).
+			// These rules have false positives when TypeScript compiler is happy.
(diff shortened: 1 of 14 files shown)


eslint.config.mjs
```diff
@@ -30,19 +30,6 @@ export default [
...
-			'src/views/WalletPage.svelte',
-			'src/views/NavigationItems.svelte',
-			'src/views/Eip7702Table.svelte',
-			'src/components/Typography.svelte',
-			'src/components/Tooltip.svelte',
-			'src/components/Table.svelte',
-			'src/components/Select.svelte',
-			'src/components/Filters.svelte',
-			'src/components/BlockTransition.svelte',
-			'src/views/WalletAttributeSummary.svelte',
```
polymutex, line 45: Looks like ESLint is still unhappy about some of these files. In order to not disrupt others submitting PRs and getting failing CI runs for reasons unrelated to their PRs, I recommend keeping the minimum set of files here that is necessary to get CI to pass, or clean up what it's still complaining about :)

polymutex (approved):
Thanks a lot for doing this 🙇, I know it's thankless work.

However, please don't submit until CI passes, either by fixing the remaining issues or keeping some files ESLint-ignored (see comment)

eslint.config.mjs
darrylyeo: All the remaining lint issues are just warnings now. Now I am trying to figure out why there are parsing errors 🤔

polymutex: My methodology for finding out the ESLint parser's differences vs Svelte itself was to basically take out chunks of code until the file parsed, then adding them back slowly and figuring out which part is actually problematic...

Things I know of: multiline `const`/`let` statements with a parenthesis (`const foo = (`), and strings with backticks with `${}`-style expansions inside them, but not all of them for some reason; couldn't figure out what makes some of them no good but others work. Maybe expansions inside expansions, like
```
`${foo ? `${bar}` : baz}`
```
fails on the `${bar}` part.

darrylyeo: Running into this as well and can't figure out the pattern either. Will ignore the two offending files for now.

Can raise here if we manage to isolate the cause.
https://github.com/sveltejs/eslint-plugin-svelte/issues

polymutex (approved):

Issue #305, opened 2025-10-30
Safe wallet: Refactor Safe creation feature using `Support`


polymutex:
Follow up from @Fbartoli's #301:

In https://github.com/walletbeat/walletbeat/blob/45a203f06ac706d38f61cac037e7c653f6f5ab8e/src/schema/features/account-support.ts#L227-L255

Change

```typescript
	canDeployNew: boolean

	/** Default configuration when creating a new Safe. */
	defaultConfig: {
		/** Number of owners by default. */
		owners: number
		/** Signature threshold by default. */
		threshold: number
		/** Enabled modules by default. */
		modules: string[] // or more specific type if needed
	}
```

into:

```typescript
	canDeployNew: Support<{
		defaultConfig: {
			/** Number of owners by default. */
			owners: number
			/** Signature threshold by default. */
			threshold: number
			/** Enabled modules by default. */
			modules: string[] // or more specific type if needed
		}
	}>
```

This way the `defaultConfig` only needs to be specified in the case where the wallet actually supports deploying Safe contracts.

polymutex:
@maykelxyz Consider this issue as starter issue as well. Should be fairly straightforward.

Fbartoli:
Thanks for the comment, i will propose something

polymutex:
Fixed in #341

PR #341, opened 2025-11-06
Update wallet configurations and add new entries to .gitignore

Fbartoli:
Refactors Safe creation feature using `Support` type.

This change refactors `canDeployNew` to use the `Support` type pattern,
so that `defaultConfig` only needs to be specified when the wallet
actually supports deploying Safe contracts.

Refs #305

- Modified `rabby.ts` to change `canDeployNew` to `notSupported` and removed the default configuration.
- Updated `safe.ts` to implement a supported structure for `canDeployNew` with a default configuration.
- Adjusted the `AccountTypeSafe` interface in `account-support.ts` to reflect the new structure for `canDeployNew`.

### DIFF
--- a/data/software-wallets/rabby.ts
+++ b/data/software-wallets/rabby.ts
@@ -86,13 +86,8 @@ export const rabby: SoftwareWallet = {
 			rawErc4337: notSupported,
 			safe: supported({
 				ref: refTodo,
-				canDeployNew: false,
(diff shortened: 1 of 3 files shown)


maykelxyz:
Lgtm

polymutex (approved):

polymutex:
Thanks!

PR #348, opened 2025-11-11
Layout updates: content sections

### LINKED ISSUE #319: UI: Per-attribute-group page should default to expanded view
The per-attribute page shows a single pie chart for that attribute group. Would be better if it was already pre-expanded to show individual attributes.

darrylyeo:
* Update all layouts to use `[data-scroll-item]`s and `[data-sticky]` headers
* Tweak various other layout parameters

---

### Design system

* `[data-scroll-container]`
  * disable the ability to nest `[data-scroll-container]`s for now to support Firefox ([requires `@scope` support](https://bugzilla.mozilla.org/show_bug.cgi?id=1830512))

* `[data-scroll-item]`
  * configure padding start and padding end independently; use the smaller value when matching values on cross axis

### Views

* Front page
  * ensure subtitle doesn't overlap with background graphic

* FAQ page
  About page
  `<WalletTable>`
  `<Eip7702Table>`
  * update to use `[data-sticky-container]` / `[data-sticky]` / `[data-scroll-item]`

* `<WalletTable>`
  * expand single hierarchical column by default
    * fixes https://github.com/walletbeat/walletbeat/issues/319

### DIFF
--- a/src/components/Typography.svelte
+++ b/src/components/Typography.svelte
@@ -5,21 +5,24 @@
 	"
 >
 	// Types
+	import type { SvelteHTMLElements } from 'svelte/elements'
 	import { ContentType, type TypographicContent as _TypographicContent } from '@/types/content'
-	import { renderStrings } from '@/types/utils/text'
 
 
 	// Props
 	let {
 		content,
 		strings,
-	}: {
+		...restProps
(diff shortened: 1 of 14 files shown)


polymutex (approved):
A few comments but will file as separate issues as this is a clear net improvement :)

Issue #226, opened 2025-07-12
Browser wallet evaluation page


polymutex:
Idea from #219.

We should maintain a static page where we can test out browser wallets. This would help quickly evaluating browser wallets' support for certain EIPs, such as EIP-1193, EIP-2700, EIP-6963, EIP-5792 (and checking the atomicity support for that too).

Could also be useful to try out test transactions, for calldata decoding checks and for testing "weird" transactions like 7702 account upgrades (#166).

polymutex:
@darrylyeo This isn't part of the Astro refactor, but if you don't mind, I'd like your take on how we could have such a page on the Walletbeat site with minimal (or no) dependencies.

Part of the challenge here is that high-level wallet connection libraries tend to leak data (e.g. make requests to third party websites for non-necessary things like wallet icons), and tend to smooth over the very wallet intricacies that we want such a tool to be able to distinguish (e.g. distinguishing that a wallet supports only EIP-1193 vs one that supports EIP-6963). They also may not be able to express transactions we want to test, e.g. #166.

If you have worked with wallet connection libraries, do you have a recommendation that would fit the bill here? Or would we need to write our own low-level library to do this?

darrylyeo:
* [viem's wallet client](https://viem.sh/docs/clients/wallet) is the best out-of-the-box library for EIP-1193 interactions, and it also [supports EIP-5792](https://viem.sh/docs/actions/wallet/sendCalls). Might be worth diving into the source code for how it handles edge cases for both EIP-5792 and [EIP-7702](https://viem.sh/docs/eip7702).
* I don't imagine any wallet supporting EIP-1193 today would not also support EIP-2700, but should be pretty simple to test this manually by setting up some event listeners.
* EIP-6963 also is pretty trivial to implement manually – just register a request event, read what gets announced back, then match up to known reverse-DNS strings. See [eip6963.org](https://eip6963.org) or the "Wallet" menu on [blockhead.info](https://blockhead.info) for prior art.
* [wagmi's connector abstraction](https://wagmi.sh/core/api/connectors) is the best for handling the most popular browser wallets / archetypes – maybe too high-level for purposes of testing individual EIP support, but should serve as a good reference if needed.

[...]

Issue #203, opened 2025-06-23
Wallet attribute: Sign in with Ethereum (SIWE) UX


polymutex:
Sign In With Ethereum (SIWE) is a standard for connecting to websites by proving ownership of an Ethereum address. It is a standardized signature payload format which wallets can and should display in an easy-to-understand UI widget.

While the raw text payload is made to be human-understandable, the intention of the standard is also to be machine-parsable so that wallets can render it nicely.

Walletbeat should check how wallets present SIWE signing requests, and rate whether they display this appropriately for users to understand what they are signing, beyond just showing the raw text payload.

polymutex:
Folding this into transaction legibility. #404

PR #352, opened 2025-11-13
Feat: Signature intent verification maykelxyz

### LINKED ISSUE #213: Rename: hardware-dapp-signing -> Signature Intent Verification

maykelxyz:
Fixes #213 
Continuation of #222 
- A lot of changes has been made into `beta`. I figured it would be less time consuming to create a new branch, import and adjust Patrick's changes. Doing this made me understand the intent behind the PR and issue as well.


src/schema/features.ts
```diff
@@ -199,11 +203,9 @@ export function isWalletSoftwareFeatures(
  */
 export type WalletHardwareFeatures = WalletBaseFeatures & {
 	security: WalletBaseFeatures['security'] & {
-		/** Hardware wallet app signing support */
-		hardwareWalletAppSigning: VariantFeature<HardwareWalletAppSigningImplementation>
-
 		firmware: VariantFeature<FirmwareSupport>
 		keysHandling: VariantFeature<KeysHandlingSupport>
+		signingIntentClarity: VariantFeature<SigningIntentClarityImplementation>
```
polymutex: This is already in `BaseFeatures.security`, so this line is redundant.

src/schema/features.ts
```diff
@@ -375,10 +379,19 @@ export function resolveFeatures(
...
+						'security.signingIntentClarity',
+						features => features.security.signingIntentClarity,
+					)
+				}
+
+				return hardwareFeat(
+					'security.signingIntentClarity',
+					features => features.security.signingIntentClarity,
+				)
+			})(),
```
polymutex: I think you can replace all this with `baseFeat('security.signingIntentClarity', features => features.security.signingIntentClarity)`

src/schema/attributes/ecosystem/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,351 @@
...
+				'This comprehensive support gives users maximum flexibility and choice in how they interact with Web3 applications.',
+		),
+	}
+}
+
+export const dappConnectionSupport: Attribute<DappConnectionSupportValue> = {
+	id: 'dappConnectionSupport',
+	icon: '\u{1F517}', // Link symbol
+	displayName: 'dApp Connection Support',
+	wording: {
```
polymutex: This is equivalent to:

```typescript
wording: { midSentenceName: 'dApp connection support' }
```

(Also the grammar check will probably be angry at the use of the word "dApp"; it should just be "app")

src/schema/attributes/ecosystem/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,351 @@
...
+		midSentenceName: null,
+		howIsEvaluated: "How is a hardware wallet's dApp connection support evaluated?",
+		whatCanWalletDoAboutIts: sentence(
+			'What can {{WALLET_NAME}} do to improve dApp connection support?',
+		),
+	},
+	question: sentence('Can the hardware wallet connect to decentralized applications (dApps)?'),
```
polymutex: `Can the hardware wallet connect to web3 apps?`

In general, avoid "dApp" and use "app" everywhere. See PR #287 for similar refactor.

src/schema/attributes/ecosystem/hw-dapp-connection-support.ts
```diff
@@ -0,0 +1,351 @@
...
+				connectionDetails: notSupported,
+			})
+		}
+
+		// Extract references if supported
+		const references = isSupported(dappSupport) ? refs(dappSupport) : []
+
+		// If not supported, cannot connect to dApps
+		if (!isSupported(dappSupport)) {
+			return noDappConnectionSupport()
```
polymutex: This needs to be merged with the `references`:

```typescript
return { ...noDappConnectionSupport(), references }
```

Same for all the `return` statements below.

Alternatively you can do this if you don't want to edit every single `return`:

```typescript
const evaluation = (() => {
		// If not supported, cannot connect to dApps
		if (!isSupported(dappSupport)) {
			return noDappConnectionSupport()
		}
		// ... All the other cases here... like:
		if (totalMethodCount === 0) {
			return noDappConnectionSupport()
		}
		if (hasOnlyClosedSource) {
			return limitedDappConnectionSupport(dappSupport)
		}
})()

const references = isSupported(dappSupport) ? refs(dappSupport) : []
return { ...evaluation, references }
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+	displayedTransactionDetails: DisplayedTransactionDetails | null,
+): Evaluation<SigningIntentClarityValue> {
+	return {
+		value: {
+			id: 'no_signing_intent_clarity',
+			rating: Rating.FAIL,
+			displayName: 'No signing intent clarity support',
+			shortExplanation: sentence(
+				'{{WALLET_NAME}} supports does not have effective signing intent clarity.',
+			),
```
polymutex: This is a bit too much jargon. How about:

```typescript
		displayName: 'Unclear transaction details',
		shortExplanation: sentence(
			'{{WALLET_NAME}} does not display clear transaction details when signing.',
		),
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+import { pickWorstRating, unrated } from '../common'
+
+const brand = 'attributes.security.signing_intent_clarity'
+
+export type SigningIntentClarityValue = Value & {
+	messageExtraction: DataExtractionMethods | null
+	messageDecoding: CalldataDecodingTypes | null
+	calldataExtraction: DataExtractionMethods | null
+	calldataDecoding: CalldataDecodingTypes | null
+	displayedTransactionDetails: DisplayedTransactionDetails | null
```
polymutex: I don't think these are used anywhere, so remove.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,481 @@
...
+			displayName: 'Basic signing intent clarity support',
+			shortExplanation: sentence('{{WALLET_NAME}} supports basic signing intent clarity.'),
+			messageExtraction,
+			messageDecoding,
+			calldataExtraction,
+			calldataDecoding,
+			displayedTransactionDetails,
+			__brand: brand,
+		},
+		details: paragraph(
```
polymutex, line 122: It would be good to have some details as to what the wallet supports and does not support (based on the `messageExtraction` / `messageDecoding` / `calldataExtraction` / `calldataDecoding` / `displayedTransactionDetails` arguments). These details are especially important in this particular `basicSigningIntentClarity` case because the wallet only supports a subset, but it could also be shown in all cases since it's good information to display in general.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,481 @@
...
+	id: 'signingIntentClarity',
+	icon: '\u{1F50F}', // Lock with pen
+	displayName: 'Signing Intent Clarity',
+	wording: {
+		midSentenceName: null,
+		howIsEvaluated: "How is a wallet's signing intent clarity support evaluated?",
+		whatCanWalletDoAboutIts: sentence(
+			'What can {{WALLET_NAME}} do to improve signing intent clarity support?',
+		),
+	},
```
polymutex, line 354: Jargon here too. Consider:

```typescript
	wording: {
		midSentenceName: null,
		howIsEvaluated: "How is a wallet's signing intent clarity evaluated?", // <- Notice the removal of the word "support"
		whatCanWalletDoAboutIts: sentence(
			'What can {{WALLET_NAME}} do to make it easy for users to understand what they are signing?',
		),
	},
```

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+	displayName: 'Signing Intent Clarity',
+	wording: {
+		midSentenceName: null,
+		howIsEvaluated: "How is a wallet's signing intent clarity support evaluated?",
+		whatCanWalletDoAboutIts: sentence(
+			'What can {{WALLET_NAME}} do to improve signing intent clarity support?',
+		),
+	},
+	question: sentence(
+		'When asking the user to sign a transaction, does the wallet give enough information for the user to verify the what the signature is for?',
```
polymutex: Typo ("the user to verify the what the"). But overall suggestion: "When signing a transaction, does the wallet show transaction details clearly?"

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+		into signing transactions with different parameters than what they intended.
+		
+		Without this, users are at the mercy of the dApp they are interacting with sending them a bad transactions, either because they have a bug, were hacked, or are malicious. Without a signer being able to verify if their transaction is correct, user should not send such a transaction.
+		
+		Full signing intent clarity implementations ensure that all relevant transaction details (recipient
+		address, amount, fees, etc.) are clearly displayed on the wallet screen, EIP-712 message hashes,
+		and decoded calldata, allowing users to make informed decisions before authorizing transactions.
+	`),
+	methodology: markdown(`
+		Wallets are evaluated based on their implementation of signing intent clarity capabilities.
```
polymutex: Wallets are evaluated based on how clearly they display transaction details for a specific set of benchmark transactions.

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,481 @@
...
+				messageExtraction: null,
+				messageDecoding: null,
+				calldataDecoding: null,
+				calldataExtraction: null,
+				displayedTransactionDetails: null,
+			})
+		}
+
+		// Extract references from the wallet signing intent clarity feature
+		const references = features.security.signingIntentClarity?.ref
```
polymutex: Use `isSupported` here to avoid the `?`

polymutex (commented):
Looking pretty good :)

polymutex:
Bikeshedding names is always easy, but here goes:

- I think "dApp Connection Support" should be renamed to "Web3 app support"
- "Signing Intent Clarity" should be renamed "Clear signing", or maybe "Transaction legibility", or just "Signing clarity"

No strong opinion though, I think the current names are fine too. You should ask @PatrickAlphaC for ideas too as he probably has better opinions on the naming here, since he has talked about this problem for much longer than I have.

src/schema/features.ts
maykelxyz: Resolved

src/schema/features.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: I refactored to:
```ts
		const references = refs(features.security.signingIntentClarity)
```

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: Resolved

src/schema/attributes/security/transaction-legibility.ts
polymutex: btw, adding this information as generated markdown is probably OK for now, but longer-term (maybe in a separate PR) I think this deserves a custom content type. This way we could display this in a more appropriate format such as a table.

Don't worry about it for this PR (it's more important to get the attributes submitted sooner before DevConnect so that wallets can populate their feature data earlier), but in the future check out how some other attributes use custom UI components in the `details` field of the `Evaluation` they return (example: `src/schema/attributes/security/security-audits.ts`).

src/schema/attributes/ecosystem/hw-dapp-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

data/hardware-wallets/bitbox.ts
```diff
@@ -53,6 +58,21 @@ export const bitboxWallet: HardwareWallet = {
...
+		appConnectionSupport: supported<WithRef<AppConnectionMethodDetails>>({
+			ref: [
+				{
+					explanation:
+						'BitBox blog post explaining WalletConnect integration for secure dApp connections',
+					url: 'https://blog.bitbox.swiss/en/using-walletconnect-to-securely-connect-to-your-favorite-dapp/',
+				},
+			],
+			details:
+				'BitBox02 supports dApp connections through their open-source BitBoxApp, WalletConnect protocol, and integration with Rabby wallet.',
```
polymutex: dApp → app

data/hardware-wallets/trezor.ts
```diff
@@ -63,6 +68,19 @@ export const trezorWallet: HardwareWallet = {
 	},
 	features: {
 		accountSupport: null,
+		appConnectionSupport: supported<WithRef<AppConnectionMethodDetails>>({
+			ref: 'https://trezor.io/guides/third-party-wallet-apps/third-party-wallet-apps-dapps',
+			details:
+				'Trezor supports multiple dApp connection methods including their open-source Trezor Suite, WalletConnect, and integration with popular software wallets like MetaMask and Rabby.',
```
polymutex: dApp → app

data/software-wallets/rabby.ts
```diff
@@ -599,6 +600,47 @@ export const rabby: SoftwareWallet = {
 					userWhitelist: true,
 				}),
 			},
+			signingIntentClarity: {
+				ref: refTodo,
+				messageSigning: {
+					calldataDecoding: {
+						[CalldataDecoding.ETH_USDC_TRANSFER]: false,
```
polymutex: Why is this `false`?

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+		value: {
+			id: 'no_app_connection',
+			rating: Rating.FAIL,
+			displayName: 'No app connection support',
+			shortExplanation: sentence('{{WALLET_NAME}} cannot connect to apps.'),
+			connectionDetails: notSupported,
+			__brand: brand,
+		},
+		details: paragraph(
+			"{{WALLET_NAME}} does not support connecting to web3 applications (apps). This severely limits the wallet's functionality, as users cannot interact with DeFi protocols, NFT marketplaces, or other Web3 applications. Without app connectivity, the wallet can only be used for basic sending and receiving of assets.",
```
polymutex: Can remove the "(apps)" part now, it made sense when it said "(dapps)" but not that it says "apps" it looks redundant.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+		),
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should implement at least one method of connecting to apps such as support through popular software wallets.',
+		),
+	}
+}
+
+function limitedAppConnectionSupport(
```
polymutex: This function is basically returning one of two different ratings: one if `hasOnlyClosedSource` and one if not. Instead, you can just have two functions (one for the `hasOnlyClosedSource` case and one for the other). It'll be clearer what they do, and the `howToImprove` paragraph can be more specific in its recommendation.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+					? '{{WALLET_NAME}} can only connect to apps through its proprietary closed-source application.'
+					: '{{WALLET_NAME}} has limited options for connecting to apps.',
+			),
+			connectionDetails,
+			__brand: brand,
+		},
+		details: paragraph(
+			hasOnlyClosedSource
+				? "{{WALLET_NAME}} can connect to apps, but only through its own proprietary closed-source application. This creates vendor lock-in and requires users to trust the wallet provider's software without the ability to verify its security. Users cannot use their preferred software wallet or standard protocols."
+				: '{{WALLET_NAME}} supports connecting to apps but with limited options. While functional, the restricted connection methods may limit user choice and flexibility in how they interact with Web3 applications.',
```
polymutex: "with limited options" is kind of vague. Worth describing what they are in more details based on the function's input arguments.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+	icon: '\u{1F517}', // Link symbol
+	displayName: 'app Connection Support',
+	wording: {
+		midSentenceName: null,
+		howIsEvaluated: "How is a hardware wallet's app connection support evaluated?",
+		whatCanWalletDoAboutIts: sentence(
+			'Can {{WALLET_NAME}} connect to web3 applications (apps)?',
+		),
+	},
+	question: sentence('Can the hardware wallet connect to web3 applications (apps)?'),
```
polymutex: "Can the hardware wallet be used with web3 applications?"

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+	why: markdown(`
+The ability to connect to web3 applications (apps) is crucial for hardware wallet
```
polymutex: Same comment on the "(apps)" part (just remove)

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+more choice and transparency.
+`),
+	methodology: markdown(`
+Hardware wallets are evaluated based on their app connection capabilities and the 
+variety of methods they support.
+
+A wallet receives a passing rating if it supports multiple connection methods, especially 
+if it includes standard protocols or integration with well-known 
+software wallets. Excellent ratings are given to wallets that also provide open-source 
+solutions.
```
polymutex: I know this was the methodology that was used in the previous PR, but in retrospect I don't think it makes sense. IMO what matters isn't the *number* of ways to connect, what matters is how reliable and open the _best_ connection method is, and the _breadth_ of the apps available through that connection method.

In other words:

- Can't connect → fail
- Can connect to _any_ app but requires trusting unverifiable code → partial
- Can connect to _some_ apps using entirely verifiable code/open standards → partial
- Can connect to _any_ app using entirely verifiable code/open standards → pass

Does that make sense, and do you think that's better?

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+
+A wallet receives a partial rating if it can connect to apps but with limitations, such 
+as only supporting a proprietary closed-source application or having very few connection 
+options.
+
+A hardware wallet fails this attribute if it cannot connect to apps at all, severely 
+limiting its utility in the modern Web3 ecosystem.
+
+This attribute only applies to hardware wallets. Software wallets and smart contract 
+wallets are exempt as they inherently support app connections.
```
polymutex: No need for this last sentence; this is only ever displayed on hardware wallet pages so it would be weird to talk about software wallets there.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,349 @@
...
+						ref: refTodo,
+						supportedConnections: {
+							[AppConnectionMethod.VENDOR_CLOSED_SOURCE_APP]: true,
+						},
+					}),
+				),
+			),
+			exampleRating(
+				paragraph(`
+			The wallet has limited app connection options, supporting only one or two
```
polymutex: Need to update here; IMO this should pass so long as the MetaMask integration is permissionless (such that others software wallets could integrate it if they wanted to).

src/schema/features/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,93 @@
...
+export interface AppConnectionMethodDetails {
+	/**
+	 * Which connection methods are supported (must have at least one)
+	 */
+	supportedConnections: NonEmptySet<AppConnectionMethod | SoftwareWalletType>
+
+	/**
+	 * Additional details about app connection capabilities
+	 */
+	details?: string
```
polymutex: Not used anywhere, so delete.

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,479 @@
...
+	question: sentence(
+		'When signing a transaction, does the wallet show transaction details clearly?',
+	),
+	why: markdown(`
+		Signing intent clarity is a critical security feature for wallets that allows users to verify
+		transaction details directly on their wallet's screen/window before signing. This verification
+		step is crucial for preventing attacks where malicious software might attempt to trick users
+		into signing transactions with different parameters than what they intended.
+		
+		Without this, users are at the mercy of the dApp they are interacting with sending them a bad transactions, either because they have a bug, were hacked, or are malicious. Without a signer being able to verify if their transaction is correct, user should not send such a transaction.
```
polymutex: dApp → app

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,479 @@
...
+	`),
+	methodology: markdown(`
+		Wallets are evaluated based on how clearly they display transaction details for a specific set of benchmark transactions.
+
+		A wallet receives a passing rating if it implements full signing intent clarity, where all transaction
+		details are clearly displayed on the wallet screen for verification before signing. This includes
+		support for standard transactions, ERC-20 token transfers, EIP-712 messages and complex contract interactions. 
+
+		The should be able to **display clearly all transaction types on Safe, Aave and Uniswap.**
+		To do so, if the wallet is a hardware wallet, **the wallet MUST be able to connect directly to the dapp**.
```
polymutex: dapp → app

src/schema/attributes/security/signing-intent-clarity.ts
```diff
@@ -0,0 +1,479 @@
...
+		and decoded calldata, allowing users to make informed decisions before authorizing transactions.
+	`),
+	methodology: markdown(`
+		Wallets are evaluated based on how clearly they display transaction details for a specific set of benchmark transactions.
+
+		A wallet receives a passing rating if it implements full signing intent clarity, where all transaction
+		details are clearly displayed on the wallet screen for verification before signing. This includes
+		support for standard transactions, ERC-20 token transfers, EIP-712 messages and complex contract interactions. 
+
+		The should be able to **display clearly all transaction types on Safe, Aave and Uniswap.**
```
polymutex: The current attribute only looks at Safe and Aave, not Uniswap. Either add Uniswap or remove it from this sentence

polymutex (changes requested):
Leaving some comments, this time reviewed all files.

There are a few changes I'd like to suggest on the signing intent clarity attribute (some of which I posted on [this thread](https://x.com/walletbeat/status/1993140734107828584) btw):

- Renaming to "Transaction legibility". I've asked Patrick in person at DevConnect and he agreed it was a great name, and used it in a [tweet](https://x.com/PatrickAlphaC/status/1992365804416094701) recently.
- Adding Sign in with Ethereum as one of the sample transactions we test. I think this would also address #203 in the same go, and consolidate both attribute.
- Adding a Uniswap ETH/USDC swap transaction to the set of sample transactions.
- Adding WETH/ETH wrap/unwrap.
- Adding a Cowswap ETH/USDC swap signature (this tests both EIP-712 and ERC-1271 signatures depending on the account type).
- Adding a USDC token approval transaction.
- Due to the above transactions and probably not having the data for all of them, I suggest making it possible to specify `null` in individual transactions (so making the types be `Record<CalldataDecoding, boolean>` → `Record<CalldataDecoding, boolean | null>` so that it's possible to submit this PR before compiling all this new data.

Longer-term I think we'll also need a test page to be able to automate creating these transactions (#226) but that's for a different PR for sure.

data/hardware-wallets/trezor.ts
maykelxyz: Resolved

data/hardware-wallets/bitbox.ts
maykelxyz: Resolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/features/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

data/software-wallets/rabby.ts
maykelxyz: Most of the software wallet rating for the signature intent verification feature came from the original PR from Patrick:

https://github.com/walletbeat/walletbeat/pull/222/files#diff-cdbebf33e79a7370272ce23a38cecb2c408f7cb7e66098acc732a0b5b074f5c9R444-R451

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

maykelxyz: Resolved

data/software-wallets/rabby.ts
polymutex: OK. That doesn't mean it's correct though :)

maykelxyz: Just came to realize most of the software wallet rating from the last PR is correct as well. 

I'll be updating the software wallet rating of the feature for this PR. thanks for catching that :)

src/schema/attributes/security/signing-intent-clarity.ts
maykelxyz: Resolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz:
```ts
			exampleRating(
				paragraph(`
			The wallet has limited app connection options, supporting only a proprietary 
			closed-source application and a software wallet integration that requires 
			permission, without any permissionless options like MetaMask.
		`),
				limitedAppConnectionSupport(
					supported({
						ref: refTodo,
						supportedConnections: {
							[AppConnectionMethod.VENDOR_CLOSED_SOURCE_APP]: true,
							[SoftwareWalletType.OTHER]: true,
						},
					}),
				),
			),
			

			// MetaMask integration is permissionless, so it provides similar benefits to open-source apps
			const hasPermissionlessIntegration =
				appSupport.supportedConnections[SoftwareWalletType.METAMASK] === true

			if (totalMethodCount <= 2 && !hasOpenSource && !hasPermissionlessIntegration) {
				return limitedAppConnectionSupport(appSupport)
			}

			if (totalMethodCount <= 2 && (hasOpenSource || hasPermissionlessIntegration)) {
				return goodAppConnectionSupport(appSupport)
			}
```

Updated to this. Feel free to let me know if you have any thoughts.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Rsolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: This makes sense and is definitely better compared to the previous methodology. For now, we can stick with this as I can't think of a better one. I think that's fine. I'll suggest in the future

maykelxyz: Updated, please check `src/schema/attributes/ecosystem/hw-app-connection-support.ts`

data/software-wallets/metamask.ts
```diff
@@ -348,6 +349,49 @@ export const metamask: SoftwareWallet = {
 					userWhitelist: true,
 				}),
 			},
+			signingIntentClarity: {
+				ref: refTodo,
+				messageSigning: {
+					calldataDecoding: {
+						[CalldataDecoding.ETH_USDC_TRANSFER]: false,
```
polymutex: I'm fairly certain that MetaMask displays this fine too

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	if (methods.length === 2) {
+		return `${methods[0]} and ${methods[1]}`
+	}
+
+	// 3 or more methods
+	const last = methods.pop()!
+
+	return `${methods.join(', ')}, and ${last}`
```
polymutex: Use `commaListFormat` instead of this.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,378 @@
...
+		methods.push('Frame')
+	}
+
+	if (supported[SoftwareWalletType.AMBIRE] === true) {
+		methods.push('Ambire')
+	}
+
+	if (supported[SoftwareWalletType.OTHER] === true) {
+		methods.push('other software wallets')
+	}
```
polymutex: Define a separate function like:

```typescript
function connectionMethodToString(connectionMethod AppConnectionMethod): string {
  switch (connectionMethod) {
    case SoftwareWalletType.AMBIRE:
      return 'Ambire'
    case ...
    // (do *not* have a default case)
  }
}
```

This will enforce that if a new enum value is added to `AppConnectionMethod`, the type system will also complain until it is also added to `connectionMethodToString`, so it prevents accidentally forgetting to update this function.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,355 @@
...
+	`),
+	methodology: markdown(`
+		Wallets are evaluated based on how clearly they display transaction details for a specific set of benchmark transactions.
+
+		A wallet receives a passing rating if it implements full transaction legibility, where all transaction
+		details are clearly displayed on the wallet screen for verification before signing. This includes
+		support for standard transactions, ERC-20 token transfers, EIP-712 messages and complex contract interactions. 
+
+		The should be able to **display clearly all transaction types on Safe, and Aave.**
+		To do so, if the wallet is a hardware wallet, **the wallet MUST be able to connect directly to the app**.
```
polymutex: I'm not sure I understand this sentence. What does it mean?

(If you're not sure either and this is from Patrick's previous wording, then maybe delete.)

src/schema/features/security/transaction-legibility.ts
polymutex, line 1: Looking back at this, one thing I'm not convinced about is the distinction between "transaction signing" and "message signing". It made sense in an earlier version of this PR back when `CalldataDecodingTypes` didn't exist and they were separate. But now there's only one `CalldataDecodingTypes` and it contains both transactions (e.g. USDC transfer) and messages (e.g. `SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND`).

I think we could simplify this quite a bit. Something like:

```typescript
export interface TransactionLegibilitySupport {
	legibility: CalldataDecodingTypes | null // Does the wallet understand each of the transaction we test?
	detailsDisplayed: DisplayedTransactionDetails | null // What details are shown?
	dataExtraction: DataExtractionMethods | null // How to extract this information out of the wallet?
}
```

... that's kind of all we need. Actually we also don't need `dataExtraction` for software wallets (because they're already on the screen). So maybe we should have two types, one for software and one for hardware:

```typescript
export interface TransactionLegibilitySupport {
	legibility: CalldataDecodingTypes | null
	detailsDisplayed: DisplayedTransactionDetails | null
}

export interface HardwareWalletTransactionLegibilitySupport {
	dataExtraction: DataExtractionMethods | null
}

// Then in `features.ts`, in `WalletHardwareFeatures`:
export type WalletHardwareFeatures = WalletBaseFeatures & {
  security: WalletBaseFeatures['security'] & {
    // [...]
    transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] & HardwareWalletTransactionLegibilitySupport
  }
  // [...]
}
```

Does this make sense?

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

src/schema/attributes/ecosystem/hw-app-connection-support.ts
maykelxyz: Resolved

data/software-wallets/metamask.ts
maykelxyz: Resolved

data/software-wallets/rabby.ts
maykelxyz: Resolved

src/schema/features/security/transaction-legibility.ts
maykelxyz: This makes sense, but I'm a bit concerned about the `dataExtraction` being specific to only hardware wallets. I think there's importance to software wallets too, specifically on the eyes and copy part. Unless we're 100% sure that by default, software wallets offer showing calldata & copying calldata right off the bat

```ts
export enum DataExtraction {
	/**
	 * Shows calldata/message data, but users have to look at it with their eyes.
	 */
	EYES = 'EYES',

	/**
	 * Shows calldata/message data, and a QR code to extract.
	 */
	QRCODE = 'QRCODE',

	/**
	 * Shows calldata/message data, and a group of hashes to compare against
	 */
	HASHES = 'HASHES',

	/**
	 * User is on a software device where they can copy the calldata directly.
	 */
	COPY = 'COPY',
}
```

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: Resolved

src/schema/features/security/transaction-legibility.ts
polymutex: ~~To me, the idea of data extraction is the ability to extract a large amount of data (the transaction details in this case) from a device which doesn't have a great way to display all this information, such as a hardware wallet with a tiny screen. In this scenario, it's useful to be able to extract the data out of the device to another device with a larger/better screen, and then being able to inspect and interpret the data before signing the transaction on the device with a tiny screen. But in the case of a software wallet, it runs on a phone or a computer with a large display, so there isn't as much of a need to extract the information to another device.~~

I'd recommend getting @PatrickAlphaC's opinion here though, perhaps there are use cases I'm missing for the need of data extraction on software wallets too.

EDIT: I am wrong and you are right, we do need something here, but I propose we do copying only. See comment below.

polymutex: Actually I realize that you're right that calldata copying could still be useful for software wallets, for the purpose of copy/pasting calldata into a separate tool in order to be able to analyze more complex transactions that the software wallet doesn't have built-in calldata decoding for. So I agree we need some way to express the fact that the software wallet allows *copying* calldata. The "eyes" part however is useless for the purpose of this use case, since this would only make sense to do for complex transactions, and so retyping the calldata would be tedious and impractical.

If you agree, then I still think we should not use `DataExtraction` as-is for software wallets, since it contains fields that wouldn't make sense to request for software wallets to implement. So perhaps there should be two separate `DataExtraction` types, or perhaps leave the `DataExtraction` type as-is and hardware-wallet-features only, but add a dedicated `calldataCopyingToClipboard: WithRef<Support>` field to software wallet features. For the "eyes" part, if useful, this could be added as a field of `displayedDetails` similar to `gas` and `nonce`.

In turn, it then wouldn't make sense to have `COPY` in `DataExtraction` for the hardware wallet case, so it could be deleted.

polymutex: On third view (and after reading your Farcaster DMs), I'm thinking software wallets could have a dedicated field like:

```typescript
calldata: Support<WithRef<CalldataDisplay>>
```

and then

```typescript
interface CallDataDisplay {
  /* Can display the calldata in raw hex format */
  rawHex: boolean

  /* Can the user copy the raw hex code to the clipboard? */
  copyHexToClipboard: boolean

  /* Can display the calldata in some formatted output (e.g. JSON) */
  formatted: boolean
}
```

maykelxyz:
I was reviewing Safepal's harware wallet and I wanted to raise this as well: 
```ts
export type CalldataDecodingTypes = Record<CalldataDecoding, boolean | null>
```

Currently, we identify calldata decoding as either `true` or `false`. Safepal's calldata decoding is kind of an interesting one. They do have calldata decoding, but only in their application or in their extension, not in the hardware wallet itself. 

We've mentioned that:
```ts
/**
 * Important: THIS INFORMATION MUST BE ON THE WALLET ITSELF. We do not trust the software "around" the wallets.
 *
 * To judge this feature of calldata decoding, we will assess a "hard-and-fast" rule of "can you decode this specific set of calldata?"
 * Hardware wallets could "cheat" this system by hard-coding just these transactions to pass the test, so we expect this list to grow over time.
 * Additionally, calldata decoding might get more advanced over time, so this list may change.
 */
 ```

But I think would be more accurate if we can introduce an enum instead of using `true` or `false`. Of course, we would only consider a passing wallet if they have calldata decoding on the wallet itself and not on the software.

Edit:

Probably something like:
```ts
/**
 * Types of transactions that a wallet can decode the calldata of.
 */
export type CalldataDecodingTypes = Record<CalldataDecoding, CalldataDecodingSupport | NotSupported>

export enum CalldataDecodingSupport {
	ON_DEVICE = 'ON_DEVICE',
	SOFTWARE_ONLY = 'SOFTWARE_ONLY',
}
```

src/schema/features/security/transaction-legibility.ts
maykelxyz: Yes I think this is better. I think this is an improvement over what we currently have

polymutex:
> I was reviewing Safepal's harware wallet and I wanted to raise this as well:
> 
> ```ts
> export type CalldataDecodingTypes = Record<CalldataDecoding, boolean | null>
> ```
> 
> Currently, we identify calldata decoding as either `true` or `false`. Safepal's calldata decoding is kind of an interesting one. They do have calldata decoding, but only in their application or in their extension, not in the hardware wallet itself.
> 
> We've mentioned that:
> 
> ```ts
> /**
>  * Important: THIS INFORMATION MUST BE ON THE WALLET ITSELF. We do not trust the software "around" the wallets.
>  *
>  * To judge this feature of calldata decoding, we will assess a "hard-and-fast" rule of "can you decode this specific set of calldata?"
>  * Hardware wallets could "cheat" this system by hard-coding just these transactions to pass the test, so we expect this list to grow over time.
>  * Additionally, calldata decoding might get more advanced over time, so this list may change.
>  */
> ```
> 
> But I think would be more accurate if we can introduce an enum instead of using `true` or `false`. Of course, we would only consider a passing wallet if they have calldata decoding on the wallet itself and not on the software.
> 
> Edit:
> 
> Probably something like:
> 
> ```ts
> /**
>  * Types of transactions that a wallet can decode the calldata of.
>  */
> export type CalldataDecodingTypes = Record<CalldataDecoding, CalldataDecodingSupport | NotSupported>
> 
> export enum CalldataDecodingSupport {
> 	ON_DEVICE = 'ON_DEVICE',
> 	SOFTWARE_ONLY = 'SOFTWARE_ONLY',
> }
> ```

Sounds good. Overall, replacing booleans with enums is always better, even if only for clarity of what they mean. In this case, your proposal would make it unambiguous that the calldata is decoded on the device screen, which is a requirement for it to actually be secure (otherwise it's possible that the software side is compromised, and the whole point of using a hardware wallet at all is to not have to trust that the software side is uncompromised).

However, I would propose using something like:

```ts
/**
 * Types of transactions that a wallet can decode the calldata of.
 */
export type CalldataDecodingTypes = Record<CalldataDecoding, Support<WithRef<CalldataDecodingSupport>>>

/** If a wallet can decode the calldata for a specific transaction, what does that look like? */
export interface CalldataDecodingSupport {
	/** Where does the calldata decoding actually happen? */
	decoded: CalldataDecoded
}

/** Where does the calldata decoding actually happen? */
export enum CalldataDecoded {
	ON_DEVICE = 'ON_DEVICE',
	OFF_DEVICE = 'OFF_DEVICE',
}
```

Otherwise type helpers like `isSupported` won't really work conveniently. This also helps if we want to add more stuff to `CalldataDecodingSupport` later; for example perhaps we'll want to add whether calldata decoding for this transaction can be done permissionlessly:

```ts
/** If a wallet can decode the calldata for a specific transaction, what does that look like? */
export interface CalldataDecodingSupport {
	/** Where does the calldata decoding actually happen? */
	decoded: CalldataDecoded

	/** Can software wallets integrating this hardware wallet perform calldata decoding permissionlessly for this transaction type? */
	requiresManufacturerConsent: boolean
}
```

I can think of a popular hardware wallet manufacture for which `requiresManufacturerConsent` would be `false` for all transactions other than token sends or ERC-20 spending approvals...

polymutex:
> This also helps if we want to add more stuff to `CalldataDecodingSupport` later; for example perhaps we'll want to add whether calldata decoding for this transaction can be done permissionlessly:
> 
> ```ts
> /** If a wallet can decode the calldata for a specific transaction, what does that look like? */
> export interface CalldataDecodingSupport {
> 	/** Where does the calldata decoding actually happen? */
> 	decoded: CalldataDecoded
> 
> 	/** Can software wallets integrating this hardware wallet perform calldata decoding permissionlessly for this transaction type? */
> 	requiresManufacturerConsent: boolean
> }
> ```
> 
> I can think of a popular hardware wallet manufacture for which `requiresManufacturerConsent` would be `false` for all transactions other than token sends or ERC-20 spending approvals...

On second thought, let's not add `requiresManufacturerConsent` there. It's important, but having it *just* for calldata decoding feels like too narrow. What matters is that _all_ features of hardware wallets are permissionlessly integratable by software wallets, not just calldata decoding. I think this would fit better inside `AppConnectionMethodDetails`. Something like:

```ts
export interface AppConnectionMethodDetails {
	/**
	 * Which connection methods are supported (must have at least one)
	 */
	supportedConnections: NonEmptySet<AppConnectionMethod | SoftwareWalletType>

	/**
	 * Is manufacturer consent required to integrate any hardware wallet feature into a software wallet?
	 * If so, must provide reference. 
	 */
	requiresManufacturerConsent: { type: 'ALL_FEATURES_PERMISSIONLESSLY_INTEGRATABLE' } | MustRef<{ type: 'FEATURES_GATED_BY_MANUFACTURER' }>
}
```

It would also make perfect sense to enforce this as part of the "app connection support" attribute for hardware wallets, since this affects how feasible it is for other software wallets to integrate the hardware wallet. The reference can provide the link and explanation for why some features are gated.

maykelxyz:
For hardware wallets, we now have:
```ts
export interface TransactionLegibilitySupport {
	/**
	 * Does a wallet display transaction details clearly?
	 */
	legibility: CalldataDecodingTypes | null
	/**
	 * Does a wallet display transaction details clearly?
	 */
	detailsDisplayed: DisplayedTransactionDetails | null

	/**
	 * Does a wallet allow for data extraction?
	 */
	dataExtraction: DataExtractionMethods | null
}
```

For software wallets:
```ts
export interface DataDisplaySupport {
	/**
	 * Does the software wallet support displaying the calldata in different formats?
	 */
	calldataDisplay: WithRef<CallDataDisplay> | null
	/**
	 * Does the software wallet support displaying the transaction details?
	 */
	transactionDetailsDisplay: WithRef<TransactionDetailsDisplay> | null
}

export interface CallDataDisplay {
	/* Can display the calldata in raw hex format */
	rawHex: boolean

	/* Can the user copy the raw hex code to the clipboard? */
	copyHexToClipboard: boolean

	/* Can display the calldata in some formatted output (e.g. JSON) */
	formatted: boolean
}
```

In the previous feedback, it's mentioned that we add calldataDisplay to Software Wallet features, but originally, we also have `transactionDetailsDisplay` that's why I aggregated both `callDataDisplay` and `transactionDetailsDisplay` into just one `dataDisplay`

src/schema/features/security/transaction-legibility.ts
```diff
@@ -215,18 +237,24 @@ export enum DataExtraction {
 	 * Shows calldata/message data, and a group of hashes to compare against
 	 */
 	HASHES = 'HASHES',
+
+	/**
+	 * User is on a software device where they can copy the calldata directly.
+	 */
+	COPY = 'COPY',
```
polymutex: This enum value no longer makes sense since the enum is only ever used for hardware wallets. So remove

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,306 @@
...
+
+			// Check if wallet supports advanced data extraction (more than just visual)
+			const hasAdvancedDataExtraction: boolean =
+				dataExtraction[DataExtraction.EYES] === true &&
+				dataExtraction[DataExtraction.QRCODE] === true &&
+				dataExtraction[DataExtraction.HASHES] === true &&
+				dataExtraction[DataExtraction.COPY] === true
+
+			// PASS: Full support - complex decoding AND all details displayed AND at least one data extraction method
+			// Advanced extraction (QRCODE/HASHES/COPY) is preferred, but visual (EYES) is acceptable if all details are clearly displayed
```
polymutex: Also remove "COPY" from this comment

src/schema/attributes/ecosystem/hw-app-connection-support.ts
polymutex, line 1: Please add the check for whether features requires manufacturer consent (or add a TODO comment to add it later in a separate PR)

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,379 @@
...
+				return unverifiableAppConnectionSupport(appSupport)
+			}
+
+			// Should not reach here if feature data is correct, but handle gracefully
+			return noAppConnectionSupport()
+		})()
+
+		return { ...evaluation, references }
+	},
+	aggregate: (perVariant: AtLeastOneVariant<Evaluation<AppConnectionSupportValue>>) =>
```
polymutex: This is equivalent to just `aggregate: pickWorstRating<AppConnectionSupportValue>,`

src/schema/attributes/security/data-display.ts
polymutex, line 1: Please merge this in the transaction legibility attribute. In both the hardware wallet and software wallet cases, we are rating whether transactions are legible. The methodology is different (because we expect different things from different wallet types) but the property being verified (transaction legibility) is the same.

This will be interesting as transaction legibility will be the first truly wallet-type-dependent attribute. Other multi-type attributes exist (e.g. the source availability attribute) but that's only because their logic is identical regardless of wallet type. Here we'll want type-specific logic. Nonetheless, the type system is meant to allow this, so let's make use of it. Yes, I realize this means the fields don't map one-to-one to feature data, but that assumption is already broken in other attributes and that's OK :)

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,306 @@
...
+	supportsAnyDataExtraction,
+} from '@/schema/features/security/transaction-legibility'
+import { isSupported } from '@/schema/features/support'
+import { refs } from '@/schema/reference'
+import { type AtLeastOneVariant } from '@/schema/variants'
+import { markdown, mdParagraph, paragraph, sentence } from '@/types/content'
+
+import { pickWorstRating, unrated } from '../common'
+
+const brand = 'attributes.transaction-legibility'
```
polymutex: Please use underscores instead of dashes in brand strings. There's no deep reason for this, it's just the convention other attributes have followed so far.

src/schema/features/security/data-display.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	/* Can the user copy the raw hex code to the clipboard? */
+	copyHexToClipboard: boolean
+
+	/* Can display the calldata in some formatted output (e.g. JSON) */
+	formatted: boolean
+}
+
+/**
+ * What essential transaction data does the hardware wallet show? (Or can show via options)
```
polymutex: The "or can show via options" is unclear because it leaves it open for interpretation. I recommend either:

a. Make the booleans mean "shown by default" or not, to close the ambiguity

or

b. Add a three-state enum (`SHOWN_BY_DEFAULT`, `SHOWN_OPTIONALLY`, `NOT_IN_UI`) and we can decide how to rate which value in attribute code.

polymutex (changes requested):
Getting close. I think the biggest thing left to do here is to merge all the transaction legibility stuff down to a single (multi-wallet-type) attribute, not a separate attribute with similar intent. (The fact that feature fields are separated this way doesn't mean the attributes also have to be)

maykelxyz:
The solution I had with merging both software and hardware into one transaction eligibility is:
```ts
// features.ts
export interface ResolvedFeatures {
		transactionLegibility: ResolvedFeature<
			HardwareTransactionLegibilityImplementation | SoftwareTransactionLegibilityImplementation
		>
 ....
}
// security/transaction-legiblity.ts
/**
 * A record of transaction legibility support (both message and transaction)
 */
export interface SoftwareTransactionLegibilitySupport {
	readonly kind: 'software'
	/**
	 * Does the software wallet support displaying the calldata in different formats?
	 */
	calldataDisplay: CallDataDisplay | null
	/**
	 * Does the software wallet support displaying the transaction details?
	 */
	transactionDetailsDisplay: DisplayedTransactionDetails | null
}

/**
 * A record of transaction legibility support (both message and transaction)
 */
export interface HardwareTransactionLegibilitySupport {
	readonly kind: 'hardware'
	/**
	 * Does a wallet display transaction details clearly?
	 */
	legibility: CalldataDecodingTypes | null
	/**
	 * Does a wallet display transaction details clearly?
	 */
	detailsDisplayed: DisplayedTransactionDetails | null

	/**
	 * Does a wallet allow for data extraction?
	 */
	dataExtraction: DataExtractionMethods | null
}
```

i added the `kind` property to the support to enable type checking in `attributes/transaction-legibility.ts`

```ts
	evaluate: (features: ResolvedFeatures): Evaluation<TransactionLegibilityValue> => {
		if (features.security.transactionLegibility === null) {
			return unrated(transactionLegibility, brand, null)
		}

		if (features.security.transactionLegibility.kind === 'hardware') {
			return evaluateHardwareWalletTransactionLegibility(features.security.transactionLegibility)
		}

		return evaluateSoftwareWalletTransactionLegibility(features.security.transactionLegibility)
	},
```

Then would proceed to evalute the wallets. It was pretty tricky. but the key thing is to allow transactionLegibility to be of kind either `HardwareTransactionLegibilityImplementation` or `SoftwareTransactionLegibilityImplementation`, and evaluate based on the wallet type, and the main change is to be in the `attributes/security/transaction-legibility.ts` Thanks for the guide @polymutex

polymutex:
I like it in terms of how it makes the feature field under a coherent name :)

However the `kind` field means that all the wallet files now need to have `kind: 'hardware'` or `kind: 'software'` written in them, which seems redundant given that they already have a top-level type definition. I think you can achieve the same effect in `evaluate` without the `kind` field, by using TypeScript type predicate functions instead:

```typescript
// Notice the return type is `transactionLegibility is HardwareTransactionLegibilityImplementation`, not `boolean`:

/**
 * Type predicate for `HardwareTransactionLegibilityImplementation`.
 */
export function isHardwareTransactionLegibility(
  transactionLegibility:
      HardwareTransactionLegibilityImplementation
    | SoftwareTransactionLegibilityImplementation,
): transactionLegibility is HardwareTransactionLegibilityImplementation {
  // The `dataExtraction` field exists only on `HardwareTransactionLegibilityImplementation`,
  // not on `SoftwareTransactionLegibilityImplementation`, so it is a good way to distinguish
  // between the two types:
  return Object.hasOwn(transactionLegibility, 'dataExtraction')
}
```

Then you can use it in `evaluate` like this:

```typescript
	evaluate: (features: ResolvedFeatures): Evaluation<TransactionLegibilityValue> => {
		if (features.security.transactionLegibility === null) {
			return unrated(transactionLegibility, brand, null)
		}

		if (isHardwareTransactionLegibility(features.security.transactionLegibility)) {
			return evaluateHardwareWalletTransactionLegibility(features.security.transactionLegibility)
		}

		return evaluateSoftwareWalletTransactionLegibility(features.security.transactionLegibility)
	},
```

(and then remove the `kind` field)

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,377 @@
...
+	connectionDetails: Supported<WithRef<AppConnectionMethodDetails>>,
+): string {
+	const methods: string[] = []
+	const supported = connectionDetails.supportedConnections
+
+	for (const method of allConnectionMethods) {
+		if (supported[method] === true) {
+			methods.push(connectionMethodToString(method))
+		}
+	}
```
polymutex: Equivalent to:

```typescript
const methods = setItems(supported).map(connectionMethodToString)
```

Or, to keep them in `allConnectionMethods`'s order:

```typescript
const methods = allConnectionMethods.filter(method => setContains(supported, method)).map(connectionMethodToString)
```

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,377 @@
...
+
+	if (methods.length === 0) {
+		return 'no connection methods'
+	}
+
+	return commaListFormat(methods)
+}
+
+export type AppConnectionSupportValue = Value & {
+	connectionDetails: Support<WithRef<AppConnectionMethodDetails>>
```
polymutex: This field is not used anywhere as far as I can tell, so remove.

src/schema/features/ecosystem/hw-app-connection-support.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Returns the number of connection methods (including software wallets)
+ */
+export function countAllConnectionMethods(appSupport: AppConnectionSupport): number {
+	if (!isSupported(appSupport)) {
+		return 0
+	}
+
+	return Object.values(appSupport.supportedConnections).filter(supported => supported === true)
+		.length
```
polymutex: Equivalent to: `return setItems(appSupport.supportedConnections).length`

src/schema/features/security/transaction-legibility.ts
```diff
@@ -1,39 +1,47 @@
 import type { WithRef } from '@/schema/reference'
 
+import { notSupported, type Support } from '../support'
+
+export enum TransactionDisplayOptions {
```
polymutex, line 5: Add `/** comments */` here explaining what each value means. It's fairly obvious for `SHOWN_BY_DEFAULT` or `NOT_IN_UI`, but `SHOWN_OPTIONALLY` is a bit less obvious. Does it mean "on the transaction approval screen but you need to click a button for it", or does it mean "on the transaction approval screen, but you need to have changed the wallet settings ahead of time for it to appear there"? (Or either?)

src/schema/features/security/transaction-legibility.ts
polymutex, line 182: We also want to add a SIWE request here for #203. (Add to the comment)

src/schema/features/security/transaction-legibility.ts
```diff
@@ -169,25 +182,40 @@ export enum CalldataDecoding {
...
+	[CalldataDecoding.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND]: notSupported,
 }
 
 /**
  * Returns whether the given dataExtractionMethods supports any data
  * extraction method at all.
  */
 export function supportsAnyCalldataDecoding(calldataDecodingTypes: CalldataDecodingTypes): boolean {
-	return Object.values(calldataDecodingTypes).includes(true)
+	return Object.values(calldataDecodingTypes).some(support => support.support === 'SUPPORTED')
```
polymutex: `return Object.values(calldataDecodingTypes).some(isSupported)`

src/schema/features.ts
```diff
@@ -82,6 +86,11 @@ export interface WalletBaseFeatures {
 		/** Bug bounty program implementation */
 		bugBountyProgram: VariantFeature<Support<BugBountyProgramImplementation>>
 
+		/** Transaction legibility features. */
+		transactionLegibility: VariantFeature<
+			HardwareTransactionLegibilityImplementation | SoftwareTransactionLegibilityImplementation
+		>
```
polymutex, line 93: We should enforce that `WalletHardwareFeatures.security.transactionLegibility` is always of type `HardwareTransactionLegibilityImplementation`, and that `WalletSoftwareFeatures.security.transactionLegibility` is always of type `SoftwareTransactionLegibilityImplementation`. You can do this by adding this to the `WalletHardwareFeatures` and `WalletSoftwareFeatures` types below:

```typescript
export type WalletHardwareFeatures = WalletBaseFeatures & {
	security: WalletBaseFeatures['security'] & {
		transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] & HardwareTransactionLegibilityImplementation
		// [...]
	},
	// [...]
}

export type WalletSoftwareFeatures = WalletBaseFeatures & {
	security: WalletBaseFeatures['security'] & {
		transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] & SoftwareTransactionLegibilityImplementation
		// [...]
	},
	// [...]
}
```

(Otherwise it would be possible for a hardware wallet to use `SoftwareTransactionLegibilityImplementation` and get a pass on the attribute even if it doesn't do all that a hardware wallet needs to be able to do for transaction legibility...)

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+			'{{WALLET_NAME}} supports partial transaction legibility on the hardware device. The device displays most transaction details and may support calldata decoding for some transaction types, but may not fully decode complex nested transactions or provide all data extraction methods. Showing transaction details directly on the hardware device is crucial for security as it allows users to verify transaction details independently of potentially compromised software.',
+		),
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should extend its transaction legibility implementation to support decoding of complex nested transactions, ensure all essential transaction details are displayed on the device, and provide comprehensive data extraction methods (QR codes, hashes) for independent verification.',
+		),
+	}
+}
+
+function hardwareFullTransactionLegibility(
+	references: Array<{ url: string; explanation: string }> = [],
```
polymutex: The `references` are overridden in the `evaluate` function directly, so this argument is actually ignored (and not passed in anyway). So you can remove it from this function, similar to the other functions returning `Evaluation`.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+			'{{WALLET_NAME}} supports some transaction legibility features, but not all. The wallet may display some calldata formats or some transaction details, but lacks comprehensive support for all calldata display methods (raw hex, formatted, copy-able) or all essential transaction details (gas, nonce, from, to, chain, value). Showing transaction details is crucial for security as it allows users to verify transaction details before signing.',
+		),
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should extend its transaction legibility implementation to support all calldata display methods (raw hex display, formatted output, copy to clipboard) and ensure all essential transaction details (gas, nonce, from, to, chain, value) are clearly displayed for user verification.',
+		),
+	}
+}
+
+function softwareFullTransactionLegibility(
+	references: Array<{ url: string; explanation: string }> = [],
```
polymutex: Same

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+		references,
+	}
+}
+
+function evaluateSoftwareWalletTransactionLegibility(
+	softwareTransactionLegibility: SoftwareTransactionLegibilityImplementation,
+): Evaluation<TransactionLegibilityValue> {
+	if (softwareTransactionLegibility === null) {
+		return unrated(transactionLegibility, brand, null)
+	}
```
polymutex: This has already been checked earlier in `evaluate`, so no need to check it again here.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+			transactionDetailsDisplay.to === TransactionDisplayOptions.SHOWN_OPTIONALLY,
+		transactionDetailsDisplay.chain === TransactionDisplayOptions.SHOWN_BY_DEFAULT ||
+			transactionDetailsDisplay.chain === TransactionDisplayOptions.SHOWN_OPTIONALLY,
+		transactionDetailsDisplay.value === TransactionDisplayOptions.SHOWN_BY_DEFAULT ||
+			transactionDetailsDisplay.value === TransactionDisplayOptions.SHOWN_OPTIONALLY,
+	]
+
+	const totalCriteria = calldataRatings.length + transactionDetailsRatings.length
+	const passCount =
+		calldataRatings.filter(r => r).length + transactionDetailsRatings.filter(r => r).length
```
polymutex: Methodology-wise, with this type of scoring, it would be possible for a wallet to just display 3 transaction details and never display calldata, and yet still receive a `PARTIAL` rating.

I think we should be stricter about calldata being visible (and either copyable and/or formatted). Something like:

- If no calldata shown at all, `FAIL`
- Otherwise, if (calldata is not copyable) && (calldata is not formatted), still `FAIL`
- Otherwise, if less than 3 types of transaction details shown, `FAIL`
- Otherwise, if (calldata is not copyable) || (calldata is not formatted), `PARTIAL`
- Otherwise, if more than 3 types of transaction details are shown but not all of them, `PARTIAL`
- Otherwise, `PASS`

In essence, that creates a hierarchy of what's most basically required to what's less required:

1. Top priority: Calldata must be shown and either copiable/formatted (to be practically useful), and basic transaction details shown
2. Second priority: Calldata must be both copiable AND formatted, and more transaction details should be shown

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+		- Hashes: Users can compare hashes to verify data
+		- Copy to clipboard: Users can copy the data directly (software wallets)
+
+		A wallet receives a passing rating if it supports decoding of complex nested transactions and displays all essential transaction details and provides at least one data extraction method.
+
+		A wallet receives a partial rating if it has some combination of these features (decoding support, transaction details display, or data extraction methods), but not all at the full level.
+
+		A wallet receives a failing rating if it lacks calldata decoding support, does not display essential transaction details, and provides no effective data extraction methods.
+
+		For hardware wallets, the signature/transaction information *must* be visible on the hardware wallet itself. Any data shown on a software wallet component is ignored for hardware wallet ratings.
```
polymutex: This needs to be updated to reflect more of the differences between software/hardware wallets. For example, the section about data extraction should make it clear that it's only applied to hardware wallets only.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+	`),
+	ratingScale: {
+		display: 'pass-fail',
+		exhaustive: true,
```
polymutex: I suggest setting to `false` since there's many branches to this attribute now and it would be difficult to explain them all here. (Setting `exhaustive: false` just slightly changes the wording of how the rating scale is shown on the site to make it clear it's not exhaustive)

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -0,0 +1,467 @@
...
+		}
+
+		if (features.security.transactionLegibility.kind === 'hardware') {
+			return evaluateHardwareWalletTransactionLegibility(features.security.transactionLegibility)
+		}
+
+		return evaluateSoftwareWalletTransactionLegibility(features.security.transactionLegibility)
+	},
+	aggregate: (perVariant: AtLeastOneVariant<Evaluation<TransactionLegibilityValue>>) =>
+		pickWorstRating<TransactionLegibilityValue>(perVariant),
```
polymutex: Equivalent: `aggregate: pickWorstRating<TransactionLegibilityValue>`

polymutex (commented):
Looks great, the comments are this point are just minor TypeScript changes.

One thing I would add though: Please fully rate at least one software wallet on `transactionLegibility` as well. We need to keep at least one software wallet fully rated because it acts as an example for other contributors that want to do data entry for wallets. I suggest either Ambire or Rabby, or both if you're feeling up for it.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -473,11 +473,25 @@ export const transactionLegibility: Attribute<TransactionLegibilityValue> = {
...
+export function isHardwareTransactionLegibility(
+	transactionLegibility:
+		| HardwareTransactionLegibilityImplementation
+		| SoftwareTransactionLegibilityImplementation,
+): transactionLegibility is HardwareTransactionLegibilityImplementation {
+	// The `dataExtraction` field exists only on `HardwareTransactionLegibilityImplementation`,
+	// not on `SoftwareTransactionLegibilityImplementation`, so it is a good way to distinguish
+	// between the two types:
+	return Object.hasOwn(transactionLegibility, 'dataExtraction')
+}
```
polymutex: Please move the function to `src/schema/features/security/transaction-legibility.ts`, as it is a function about the transaction legibility feature type, not of the transaction legibility attribute.

src/schema/features.ts
```diff
@@ -158,6 +158,8 @@ export type WalletSoftwareFeatures = WalletBaseFeatures & {
 
 		/** Passkey verification implementation */
 		passkeyVerification: VariantFeature<PasskeyVerificationImplementation>
+		transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] &
+			SoftwareTransactionLegibilityImplementation
```
polymutex: Ah sorry, looks like the CI is saying that my type suggestion is wrong here. I think the right type is probably

```typescript
	transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] &
			VariantFeature<SoftwareTransactionLegibilityImplementation>
```

(and similar for WalletHardwareFeatures)

src/schema/features.ts
```diff
@@ -158,8 +158,10 @@ export type WalletSoftwareFeatures = WalletBaseFeatures & {
 
 		/** Passkey verification implementation */
 		passkeyVerification: VariantFeature<PasskeyVerificationImplementation>
-		transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] &
-			SoftwareTransactionLegibilityImplementation
+		transactionLegibility: Nullable<
+			WalletBaseFeatures['security']['transactionLegibility'] &
+				SoftwareTransactionLegibilityImplementation
```
polymutex: I'm fairly certain this needs to be `VariantFeature`, not `Nullable`, like this: https://github.com/walletbeat/walletbeat/pull/352#discussion_r2615181069

```typescript
	transactionLegibility: WalletBaseFeatures['security']['transactionLegibility'] &
			VariantFeature<SoftwareTransactionLegibilityImplementation>
```

`VariantFeature` allows `null`, and also allows per-variant data. Without it here, it would be impossible to specify per-variant data for `features.security.transactionLegibility`.

polymutex (changes requested):

polymutex (approved):
Approving, looks good other than the last comment on types so feel free to merge after this :)

PR #370, opened 2025-11-27
Add imKey as an independent hardware wallet type, improving imToken's hardware wallet interoperability rating

makoshan:
This update separates imKey from the generic "OTHER" category and tracks it as an independent hardware wallet type.
## Why this change?
imKey is a mainstream hardware wallet (incubated by the imToken team) and already has its own hardware wallet data file. Previously, it was categorized as "OTHER", which prevented it from being properly reflected in the hardware wallet interoperability rating. By making it independent, we can more accurately reflect a wallet's support for mainstream hardware wallets.

## What changed:
- Added IMKEY type to the hardware wallet type enum
- Updated the display function to correctly show "imKey" name
- Changed imToken's hardware wallet support from OTHER to IMKEY
- Added imKey to the list of major hardware wallet manufacturers

## Rating impact:
imToken's hardware wallet interoperability rating improves from FAIL to PARTIAL. imToken now supports 2 major hardware wallet manufacturers (Keystone and imKey), which more accurately reflects its hardware wallet support capabilities.

### DIFF
--- a/data/software-wallets/imtoken.ts
+++ b/data/software-wallets/imtoken.ts
@@ -268,7 +268,7 @@ export const imtoken: SoftwareWallet = {
 					ref: [
 						{
 							explanation:
-								'imToken works with the imKey Bluetooth hardware wallet, and with Keystone via QR codes.',
+								'imToken works with the imKey hardware wallet via Bluetooth, and with Keystone via QR codes.',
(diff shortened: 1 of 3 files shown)


polymutex (approved):
Hadn't thought of this particular problem hindering the rating, but it makes sense with this PR.

PR #397, opened 2025-12-11
Add wallet data entry contribution guide.

polymutex:


resources/docs/contribute/CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+
+Please read `resources/docs/contribute/wallet-data.md`
+
+## I want to make UI/UX changes
+
+Please read `resources/docs/contribute/ui.md`
+
+## I want to understand or change the wallet rating methodology
+
+Please read `resources/docs/contribute/methodology.md`
```
maykelxyz: We can mention probably do something like

Please read [`resources/docs/contribute/ui.md`](resources/docs/contribute/ui.md) for easier navigation

resources/docs/contribute/wallet-data.md
```diff
@@ -0,0 +1,591 @@
+# Walletbeat guide: Wallet data entry
+
+_This guide aims to help you contribute wallet data to Walletbeat. If you'd like to contribute to some other aspect of Walletbeat, see `CONTRIBUTING.md`._
+
+## Development setup
+
+Before you can contribute wallet data to Walletbeat, you need to make a copy of the site and its data, run it on your computer, make changes, and be able to submit them for review. This involves using a terminal, setting up Git, cloning the repository, setting up a TypeScript development environment, and making PRs on GitHub. If you are not familiar with these concepts, please read the `/resources/docs/contribute/development-setup.md` guide.
```
maykelxyz: [`/resources/docs/contribute/development-setup.md`](/resources/docs/contribute/development-setup.md)

resources/docs/contribute/wallet-data.md
```diff
@@ -0,0 +1,591 @@
...
+
+The rest of this guide **assumes you are already familiar with this**.
+
+## Wallet data overview
+
+Wallet data is located in the top-level `/data` directory. All files are written in TypeScript, which helps to ensure their validity and conformance. The files have the following layout:
+
```
maykelxyz: I suggest we omit this. We can mention `requirements` in `ReadMe.md` and mention that most files are in Typescript... and have a section page like:
```
### Repository Structure
```
to make it more readable

resources/docs/contribute/wallet-data.md
maykelxyz: I would suggest we also keep this in `Contributing.md` or ReadMe. I think this is part of the Setting Up section..

resources/docs/contribute/wallet-data.md
```diff
@@ -0,0 +1,591 @@
...
+### Step 4: Get the PR merged
+
+A Walletbeat contributor should soon review your changes. If there are problems (test failures, invalid data, etc.), your PR may need further work. You can update your PR by pushing additional changes to the same branch. Follow this workflow:
+
+- Make the requested changes in your editor.
+- Re-run all tests: `pnpm check:all`. If they do not pass, fix that first.
+- Commit your changes: `git add -p` then `git commit -m 'Some more fixes'`
+- Upload your changes to GitHub: `git push`
+- Write a comment on the PR to explain your fixes and to notify your reviewer that they need to take another look at your PR.
+
```
maykelxyz: I would suggest that we can move this to `Contributing.md`.. Just for better readability

maykelxyz (commented):
Overall the `wallet-data.md` is a very good guide. But as a contributor, I would be a bit overwhelmed with the amount of data to be consumed (especially that this is just one guide, there's still ui.md, methodology.md, Contributing.md, ReadMe.md). I think the key thing we could do here is to make it more "easy" to digest for first time contributors.

The guide is really good. But I think we can still find a way to simplify the guide. As you can see in my review, most of my comments, feedbacks, is for better readability. The guide in adding wallet data is also accurate, which is nice. But for some sections, setting up, creating a PR, I suggest we separate this

resources/docs/contribute/wallet-data.md
polymutex: This section is specifically showing the file layout in `/data` for the purpose of wallet data editing, and has comments that explains why each file is relevant for editing wallet data. So moving it into a common file wouldn't make much sense in my opinion. However I moved it more inside the wallet data guide to make that clearer.

polymutex:
Thanks for the review! Addressed all comments other than the one about the `/data` file layout (see above).

polymutex:
I think we should have collapsible (expandable) sections once these are displayed on the site directly. That would probably help with making this not feel so long.

PR #410, opened 2025-12-18
Refactor: Passkey verification

maykelxyz:
Changes:
- Removed passkeyImplemtation for hardware wallets as they are exempted from the rating
- Added type Support<> to PasskeyImplementation to allow wallets to have `notSupported`

### DIFF
--- a/data/embedded-wallets/unrated.tmpl.ts
+++ b/data/embedded-wallets/unrated.tmpl.ts
@@ -64,6 +64,7 @@ export const unratedEmbeddedTemplate: EmbeddedWallet = {
 			lightClient: {
 				ethereumL1: null,
 			},
+			passkeyVerification: null,
 			publicSecurityAudits: null,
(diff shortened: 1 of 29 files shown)


data/embedded-wallets/unrated.tmpl.ts
```diff
@@ -64,7 +64,6 @@ export const unratedEmbeddedTemplate: EmbeddedWallet = {
 			lightClient: {
 				ethereumL1: null,
 			},
-			passkeyVerification: null,
```
polymutex, line 67: I don't think we should remove this from embedded wallets. (Embedded wallets are probably more likely to have passkey support since passkeys are meant to be accessible from web apps). I know it's not used yet but we can still keep the field to make sure we collect this data when we start rating these wallets.

polymutex (approved):

polymutex (approved):

PR #414, opened 2025-12-19
Feat: Add Software wallets metadata

maykelxyz:
### Added New wallets metadata
- Ambire
- Elytro
- Frame
- Metamask
- Phantom
- Rabby
- Rainbow
- Zerion

### Other Changes
- Created new const `SPECIFIC_SITES_TO_SKIP` to skip specific sites that returns errors by default
```ts
			it('has valid docs', async () => {
				for (const doc of wallet.metadata.urls?.docs ?? []) {
					// These docs would always return an error Response. Skip docs to avoid failing the test.
					const DOCS_TO_SKIP=[
						'docs.phantom.com',
						'developers.zerion.io',
						'help.ambire.com/hc/en-us',
					]
					const shouldSkip = DOCS_TO_SKIP.some(s => doc.toString().includes(s))
					if (shouldSkip) {
						continue
					}

					await checkValidUrl(doc)
				}
			})
```
- Added new urls to `knownValidUrls`

### DIFF
--- a/data/software-wallets/ambire.ts
+++ b/data/software-wallets/ambire.ts
@@ -159,7 +159,18 @@ export const ambire: SoftwareWallet = {
 		iconExtension: 'svg',
 		lastUpdated: '2025-03-20',
 		urls: {
+			docs: ['https://help.ambire.com/hc/en-us'],
+			extensions: [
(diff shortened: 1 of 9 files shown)


polymutex (approved):
Nice

PR #415, opened 2025-12-19
Feat: Add manufacture consent

### LINKED ISSUE #405: Feat: add requiresManufacturerConsent

maykelxyz:
Fixes #405


data/hardware-wallets/ledger.ts
```diff
@@ -89,6 +89,7 @@ export const ledgerWallet: HardwareWallet = {
 		accountSupport: null,
 		appConnectionSupport: supported<WithRef<AppConnectionMethodDetails>>({
 			ref: 'https://support.ledger.com/article/360018444599-zd',
+			requiresManufacturerConsent: null,
```
polymutex:
```typescript
requiresManufacturerConsent: {
  type: 'FEATURES_GATED_BY_MANUFACTURER',
  ref: {
    url: 'https://developers.ledger.com/docs/clear-signing/for-wallets',
    label: 'Ledger Developer Portal',
    explanation: `Software wallet developers are required to register with Ledger's partner program before their software wallet becomes able to display clear signing data on Ledger devices.`,
  },
}
```

.cspell.json
```diff
@@ -105,6 +105,7 @@
 		"imToken",
 		"inspectability",
 		"inspectable",
+		"integratable",
```
polymutex: After looking it up, I think the correct word is [integrable](http://www.merriam-webster.com/dictionary/integrable), though to be fair I've seen "integratable" being used too... Feel free to change this or not change this.

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -324,7 +347,7 @@ limiting its utility.
 			const hasClosedSource =
 				appSupport.supportedConnections[AppConnectionMethod.VENDOR_CLOSED_SOURCE_APP] === true
 
-			if (hasClosedSource) {
+			if (hasClosedSource && !permissionless) {
```
polymutex: Can you add a branch specifically for the non-permissionless case, rather than reusing `unverifiableAppConnectionSupport`?

Generally we should make sure that evaluation details are specific to the particular thing the wallet should solve, so either becoming open-source or becoming permissionless to integrate with, not both. (Same idea as the transaction legibility changes: the `details` and `howToImprove` fields should be specific)

src/schema/attributes/ecosystem/hw-app-connection-support.ts
polymutex: This should also be expanded to include the `ref` from `requiresManufacturerConsent`.

.cspell.json
maykelxyz: Yeah I also looked it up, lol. I honestly thought integratable is the correct word.

I'll change it

src/schema/attributes/ecosystem/hw-app-connection-support.ts
```diff
@@ -325,8 +378,13 @@ limiting its utility.
...
-				// Can connect to apps but requires trusting unverifiable code → PARTIAL
-				return unverifiableAppConnectionSupport()
+				if (!permissionless) {
+					// Has closed-source app but requires manufacturer consent
+					return restrictedAppConnectionSupport(appSupport)
+				} else {
+					// Has closed-source app but is permissionless
+					return unverifiableAppConnectionSupport()
+				}
 			}
```
polymutex, line 388: A hardware wallet integration app being open-source doesn't necessarily mean hardware wallet integration is always permissionless: it's still possible for the hardware wallet manufacturer to release an open source app but have it use a proprietary token to authenticate to the hardware wallet for integration. So, even if `hasClosedSource` is `false`, `permissionless` still needs to be taken into consideration.

In other words, I think this code can just be:

```typescript
	if (!permissionless) {
		// New applications need manufacturer permission to integrate, regardless of the manufacturer's own integration work.
		return restrictedAppConnectionSupport(appSupport)
	}
	if (hasClosedSource) {
		// Has closed-source manufacturer app but is permissionless.
		return unverifiableAppConnectionSupport()
	}
```
No nesting necessary.

polymutex (approved):
Looks great, approving. No need for re-review after addressing the comment below, just merge :)

PR #420, opened 2025-12-20
Feat: Add software wallets metadata

maykelxyz:
### Added New wallets metadata
- Daimo
- Family
- Gem
- ImToken
- MtPelerin
- Nufi
- Safe

### Other Changes
- Added new sites to `SPECIFIC_SITES_TO_SKIP`
- Added new urls to `knownValidUrls`


tests/url-check.test.ts
```diff
@@ -825,6 +920,7 @@ describe('reference URLs', () => {
 						'docs.phantom.com',
 						'developers.zerion.io',
 						'help.ambire.com/hc/en-us',
+						'nufi.gitbook.io/',
```
polymutex: I think this list is starting to be long and important enough that it warrants being merged with `SOCIALS_TO_SKIP` (rename to `URLS_TO_SKIP` or something) and moved to the top of the file (right after the `knownValidUrls` declaration). And then maybe move the `shouldSkip` code to be inside the `checkValidUrl` function, so that it applies everywhere.

polymutex (approved):
This must have taken a long time to put together... thank you :)

tests/url-check.test.ts
polymutex: FYI I modified this code a bit in PR #425 to fix lint errors; you may get merge conflicts when rebasing that in.

polymutex (approved):

PR #422, opened 2025-12-22
Display transaction batching ratings in the EIP-7702 table

superKalo:
Replace the "coming soon" placeholder in the batching column with actual transaction batching ratings from wallet evaluations.

The column now displays:
- ✅ for wallets that pass transaction batching
- ❌ for wallets that fail transaction batching  
- "-" for other cases (partial, unrated, exempt, or undefined)

This change was triggered by a confused user. I was surprised to find that batching data existed, but the table did not show it. Which made it look like batching was unsupported.


polymutex:
Thank you! Had forgotten that this page existed tbh :) (yes, even despite having added the data to the wallet for it...)

PR #442, opened 2025-12-26
Feat: Add tester page

maykelxyz:
# Feature: Add a static page to test browser wallets

Aims to assist contributors in testing out browser wallets support for certain transactions, signature, and EIP supports using wagmi

#### Transactions
- ERC20 Approval
- ERC20 Transfer
- AAVE Supply
- EIP-7702 Multi-call (ERC20 Approval + ERC20 Transfer)

#### Signatures
- Basic Message Signature
- Sign in with Ethereum (EIP-4361)
- Structured Data Signature (EIP-712)

### Packages added
- wagmi@2.16.9
- @wagmi/core@3.0.0

### Other updates
- Added new page `/test`
- Added a new navigation **Test Your Wallet** that navigates to `/test` page

image

image


polymutex:
Thanks for working on this, the screenshots look great :)

A few constraints to keep in mind while working on this (I mentioned this over chat but here's them in public as well):
- The page should not include remote scripts, as this is a security risk allowing the remote script to do whatever it wants to the page.
- The page itself should never make external requests, since that's a privacy leak that we rate wallets against doing as well. This probably requires making sure all chain RPCs go through the wallet's injected provider, not a third-party RPC provider.
- (I hadn't mentioned this but thought about it just now): The page should not use cookies or `localStorage` or similar persistence mechanism. Otherwise we will have to have a cookie banner on the site and everyone hates these :) Looking at the wagmi docs, it looks like this can be done by overriding the `storage` object with a custom implementation that keeps everything in the local tab memory instead of cookies or `localStorage` which are persistent.
- As per #226, we need to be able to test wallets' implementation of browser integration standards, so not just doing test transactions. I am not sure whether wagmi allows getting down to this level of detection, because one of its design goals is to abstract away those complexities. If not possible, then we'll either have to have 2 test pages (one without wagmi), or we have to not use wagmi on this page.

polymutex:
Thoughts about getting this in before the transaction simulation testing? I think it'd already be valuable to have without it.

maykelxyz:
> Thoughts about getting this in before the transaction simulation testing? I think it'd already be valuable to have without it.

Fine by me. Ive been using it too in some wallets so I agree.

resources/contracts/.gitmodules
polymutex, line 1: Does this work when not at the repo's top level?

resources/contracts/README.md
polymutex, line 1: Not a useful README, it looks like the generic Foundry README.
Either delete, or rewrite to explain the contents of this directory specifically.

resources/files/social-media/threads/2025-01-13 - Walkaway Test/thread.md
polymutex, line 1: Changes unrelated to this PR, please revert.

.gitignore
```diff
@@ -35,3 +35,17 @@ scrap.md
...
+
+# Compiler files
+resources/contracts/cache/
+resources/contracts/out/
+resources/contracts/lib/
+
+# Ignores development broadcast logs
+resources/contracts/!/broadcast
+resources/contracts/broadcast/*/31337/
+resources/contracts/broadcast/**/dry-run/
```
polymutex, line 48: Seems redundant with the entries in `resources/contracts/.gitignore`. Delete one or the other

src/lib/wagmi-config.ts
```diff
@@ -0,0 +1,15 @@
+'use client'
+
+import { createConfig, http } from '@wagmi/core'
+import { mainnet, zksync } from '@wagmi/core/chains'
+
+const config = createConfig({
+	chains: [mainnet, zksync],
+	transports: {
+		[mainnet.id]: http(),
+		[zksync.id]: http(),
```
polymutex: As I mentioned, we must not make third party requests. Looking at the wagmi docs, `http` without any argument looks like it uses "default public RPC providers", so no good.
I think this should either be pointing to a localhost address (and it may or may not work), or it must use the wallet itself as a data provider (should be possible using `unstable_connector(injected)` and/or `custom({request: /* some function using window.ethereum... */})`)

src/lib/wagmi-config.ts
```diff
@@ -0,0 +1,15 @@
...
+	},
+	ssr: false,
+})
```
polymutex, line 28: This looks like it uses `localStorage` for storing information. As per my comment, please do not store anything there so that we do not have to show a cookie pop-up. You can use `sessionStorage` instead, or some pure memory storage is fine. https://wagmi.sh/core/api/createStorage seems relevant.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+  });
+
+  // Helper functions
+  function updateSIWEMessage() {
+    const siweSig = testSignatures.find((s) => s.id === 'siwe-1');
+
+    if (siweSig?.type === 'message') {
+      const address = account?.address || '0x0000000000000000000000000000000000000000';
+
+      siweSig.message = `https://beta.walletbeat.eth.limo/ wants you to sign in with your Ethereum account:
```
polymutex: Please make the domain part of this message dynamic based on the current domain name.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+${address}
+
+Sign in to authenticate your wallet. This is a test SIWE message.
+
+URI: https://beta.walletbeat.eth.limo/
```
polymutex: Same

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+      } catch {
+        // If parsing fails, just return the value
+      }
+    }
+
+    return value;
+  }
+
+  function openInExplorer(txHash: string) {
+    window.open(`https://etherscan.io/tx/${txHash}`, '_blank', 'noopener,noreferrer');
```
polymutex: Use Blockscout. Nothing against Etherscan but Blockscout is open-source and self-hostable.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+
+    try {
+      if (tx.calls && tx.calls.length > 0) {
+        const result = await sendCalls(config, { calls: tx.calls });
+
+        transactionState.batchIds[tx.id] = result.id;
+
+        if ('hash' in result && typeof result.hash === 'string') {
+					// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we're just asserting the has as `0x{string}`
+          transactionState.hashes[tx.id] = result.hash as `0x${string}`;
```
polymutex: Please create a helper function for this, and make it throw an error if that's not the case.

```typescript
export function assertTransactionId(s: string): `0x${string}` {
  // ... Check that the string has prefix, if not: throw
  // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we just checked
  return s as `0x${string}`
}
```

This way we keep unsafe type assertions to very narrow functions.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    // Listen for EIP-6963 provider announcements
+    window.addEventListener('eip6963:announceProvider', (event: Event) => {
+      // Type guard for CustomEvent
+      if (!('detail' in event)) return;
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      const customEvent = event as CustomEvent<{
+        info: { uuid: string; name: string; icon: string; rdns: string };
+        provider: unknown;
+      }>;
```
polymutex: Please define a type for this event and then add a type predicate function for it.

```
export function isAnnounceProviderEvent(event: Event): event is Eip6963AnnounceProviderEvent { ... }
```

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+
+      // Check if this was the last step
+      const isLastStep = step.id === testSteps[testSteps.length - 1].id;
+
+      if (isLastStep) {
+        // Show final results after last step regardless of pass/fail
+        stepTestState.overallStatus = result.status === 'passed' ? 'completed' : 'failed';
+        showFinalResults();
+      } else {
+        // Auto-advance to next step regardless of pass/fail
```
polymutex: I wasn't much of a fan of this behavior TBH. No strong opinion though.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+  }
+
+  function showFinalResults() {
+    const allResults = testSteps.map((step) => stepTestState.stepResults[step.id]).filter(Boolean);
+    const allPassed = allResults.every((r) => r.status === 'passed');
+
+    stepTestState.resultsModal.overallPassed = allPassed;
+    stepTestState.resultsModal.stepResults = allResults;
+    stepTestState.resultsModal.isOpen = true;
+  }
```
polymutex, line 498: I think the final results and the per-test checkmarks both need a way to indicate "partial success". Here I tested Ambire, and it failed the `connect` test, however this was invisible until I manually clicked to expand Step 2 to double-check:

Image

Image

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+  async function runStep2Connect(step: TestStep): Promise<StepResult> {
+    const eipResults: EIPTestResult[] = [];
+    const eip1193Checks: EIPCheckResult[] = [];
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    const provider = getProvider() as {
+      request?: (args: { method: string; params?: unknown[] }) => Promise<unknown>;
+      on?: (event: string, listener: (...args: unknown[]) => void) => void;
+      removeListener?: (event: string, listener: (...args: unknown[]) => void) => void;
+    } | null;
```
polymutex: Same here, please add a canonical type definition for this.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    const disconnectListener = () => {};
+
+    try {
+      if (typeof provider.on === 'function') {
+        provider.on('connect', connectListener);
+        provider.on('disconnect', disconnectListener);
+        disconnectEventSubscribable = true;
+      }
+    } catch {
+      // Provider doesn't support event subscription
```
polymutex: ~Shouldn't this be a test too?~
Edit: Nevermind, it is, using the `disconnectEventSubscribable` boolean later on. But the comment here should make it clear that this will be the case.

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+  // Step 3: Check Account
+  async function runStep3Account(step: TestStep): Promise<StepResult> {
+    const eipResults: EIPTestResult[] = [];
+    const eip1193Checks: EIPCheckResult[] = [];
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    const provider = getProvider() as {
+      request?: (args: { method: string; params?: unknown[] }) => Promise<unknown>;
+      on?: (event: string, listener: () => void) => void;
+    } | null;
```
polymutex: Same

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    const eipResults: EIPTestResult[] = [];
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    const provider = getProvider() as {
+      request?: (args: { method: string; params?: unknown[] }) => Promise<unknown>;
+      on?: (event: string, listener: (...args: unknown[]) => void) => unknown;
+      removeListener?: (event: string, listener: (...args: unknown[]) => void) => unknown;
+      once?: (event: string, listener: (...args: unknown[]) => void) => unknown;
+      removeAllListeners?: (event?: string) => unknown;
+    } | null;
```
polymutex: Same

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    // EIP-1193 checks
+    const eip1193Checks: EIPCheckResult[] = [];
+
+    // Get chain ID
+    let chainIdPassed = false;
+    let chainIdDetail = '';
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      const chainIdHex = (await provider.request({ method: 'eth_chainId' })) as string;
```
polymutex: Check `typeof` of it being `string`

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    let capabilitiesPassed = false;
+    let atomicitySupported = false;
+    let capabilitiesDetail = '';
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      const capabilities = (await provider.request({
+        method: 'wallet_getCapabilities',
+        params: [connectedAddress],
+      })) as Record<string, { atomicBatch?: { supported: boolean } }>;
```
polymutex: Check type

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+            calls: [
+              {
+                to: '0x0000000000000000000000000000000000000000',
+                data: '0x00',
+                value: '0x0',
+              },
+            ],
+          },
+        ],
+      })) as string | { id: string };
```
polymutex: Why are both `string` and `{ id: string }` possible?

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      const status = (await provider.request({
+        method: 'wallet_getCallsStatus',
+        params: [stepTestState.batchId],
+      })) as CallsStatusResponse;
+
+      statusPassed = true;
+
+      // EIP-5792 v2.0.0 uses numeric status codes (like HTTP: 200 = success)
+      // Earlier versions used string status ("CONFIRMED", "PENDING", etc.)
```
polymutex: Maybe we should be strict about this

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+        validResponse = false;
+      }
+
+      // Check for atomic field (EIP-5792 MUST include this)
+      if ('atomic' in status && typeof status.atomic === 'boolean') {
+        hasAtomicField = true;
+        atomicDetail = `Atomic: ${status.atomic}`;
+      } else {
+        atomicDetail = 'Atomic field missing or invalid type';
+      }
```
polymutex: I think we care more than just the field being present, we also care that it is `true` :) (`features.integration.walletCall.atomicMultiTransactions`)

src/components/WalletTest.svelte
```diff
@@ -0,0 +1,1495 @@
...
+    border-radius: 9999px;
+    background-color: var(--rating-pass);
+    box-shadow: 0 0 0 3px color-mix(in srgb, var(--rating-pass) 40%, transparent);
+  }
+
+  .account-label {
+    font-size: 0.8rem;
+  }
+
+  button[data-pressable] {
```
polymutex, line 798: The buttons don't really look clickable, took me a while to figure out they were buttons. Maybe style them up a bit

src/constants/test-eip-support.ts
```diff
@@ -0,0 +1,355 @@
...
+	name: string
+	specUrl: string
+	checks: EIPCheckDefinition[]
+}
+
+export interface EIPCheckDefinition {
+	id: string
+	name: string
+	description: string
+	critical: boolean
```
polymutex, line 39: What does this field mean?

src/constants/test-transactions-signatures.ts
```diff
@@ -0,0 +1,244 @@
...
+	contractAddress?: `0x${string}`
+	description?: string
+	requirements?: string[]
+	value?: bigint
+	// For multi-call transactions (EIP-7702)
+	calls?: Array<{
+		to: `0x${string}`
+		data: `0x${string}`
+		value?: bigint
+	}>
```
polymutex: This field uses `Array<...>` while the `parameters` field above uses `{...}[]`. Please use one or the other, not both

src/components/WalletTesterNavigationItem.svelte
polymutex, line 1: Component name could be better IMO. Most of what this does isn't so much about being in a sidebar, it's more about expressing the status (complete/pass/failed) of a test workflow step. Maybe `WalletTestStepNavigationItem` or something.

polymutex (changes requested):
Thanks a lot, this is amazing work and will make Walletbeat so much easier to contribute to.

.gitignore
maykelxyz: Forgot to delete the one in resources/contracts/.gitignore. Will delete the other one

resources/contracts/README.md
maykelxyz: Will rewrite

resources/files/social-media/threads/2025-01-13 - Walkaway Test/thread.md
maykelxyz: Hmm, this change happened when I pulled from `beta` and run either `pnpm fix` or `pnpm lint .`. So it's just a lint thing. Should I still revert?

src/lib/wagmi-config.ts
maykelxyz: Will do :)

resources/contracts/.gitmodules
maykelxyz: Deleted

src/lib/wagmi-config.ts
maykelxyz: Updated to:

```ts
declare global {
	interface Window {
		ethereum?: { request: (...args: unknown[]) => Promise<unknown> }
	}
}

const getTransport = () => {
	if (typeof window !== 'undefined' && window.ethereum) {
		return custom(window.ethereum)
	}

	return http('http://localhost:8545')
}

const config = createConfig({
	chains: [mainnet, zksync],
	transports: {
		[mainnet.id]: getTransport(),
		[zksync.id]: getTransport(),
	},
	ssr: false,
})
```

src/components/WalletTesterNavigationItem.svelte
maykelxyz: I'm leaning into: `WalletTesterNavigationItem`

resources/files/social-media/threads/2025-01-13 - Walkaway Test/thread.md
polymutex: I don't get this behavior. Are you sure you have the latest version of the packages installed (after running `pnpm i`)?
If you still get this behavior then whatever, feel free to submit this as part of this PR so long as CI is happy.

src/components/Modals/EIPResultsModal.svelte
```diff
@@ -0,0 +1,333 @@
...
+  let expandedSteps = $state<Set<string>>(new Set());
+
+  function toggleStep(stepId: string) {
+    if (expandedSteps.has(stepId)) {
+      expandedSteps.delete(stepId);
+    } else {
+      expandedSteps.add(stepId);
+    }
+
+    expandedSteps = new Set(expandedSteps);
```
polymutex: Lint warning says to use SvelteSet instead.

src/components/Tabs/EIPSupportTab.svelte
```diff
@@ -0,0 +1,565 @@
...
+      connectedAddress: string | null;
+      chainId: number | null;
+      batchId: string | null;
+    };
+    account: { address?: string } | null;
+    onRunStep: () => void;
+    onReset: () => void;
+  }
+
+  let { currentStep, currentStepResult, stepTestState, account, onRunStep, onReset }: Props =
```
polymutex: Lint warning says that `account` is unused.
If this is there for consistency, please add the same comment as other Svelte props that also have this problem: `// eslint-disable-next-line svelte/no-unused-props -- Consistent prop types for all content components.`

src/lib/wagmi-config.ts
```diff
@@ -0,0 +1,30 @@
...
+
+declare global {
+	interface Window {
+		ethereum?: { request: (...args: unknown[]) => Promise<unknown> }
+	}
+}
+
+const getTransport = () => {
+	if (typeof window !== 'undefined' && window.ethereum) {
+		return custom(window.ethereum)
```
polymutex, line 14: The wallet may not be connected initially, so perhaps this should try to look for `window.ethereum` on every call to `request`?

polymutex (changes requested):

src/lib/wagmi-config.ts
maykelxyz: We only call getTransport after a wallet is connected. Because of that, window.ethereum should already be available, and we won’t fall back to http unless no wallet exists. Let me know if I’m missing something here.

src/components/WalletTest.svelte
maykelxyz: Some wallet implementations wrap the id in an object. Though the return value **should** just be a string, spec-wise. We're just accounting for results don't follow this.

src/components/WalletTest.svelte
maykelxyz: Hmmm, what do you suggest? Imo it's okay, but no strong opinions so far..

src/components/WalletTest.svelte
maykelxyz: Updated with:

image


image

Still passing, but with warnings :)

src/constants/test-eip-support.ts
maykelxyz: If you're refering to `critical`, it's whether the check is essential for EIP compliance. Critical checks must pass for the wallet to be considered compliant with the EIP. Non-critical checks are optional / nice-to-have.

maykelxyz (commented):
Requesting a review again from @polymutex. Keeping the scope as is for now. While we did discuss adding more items to the Walletbeat testing page, I think this PR can already be reviewed and merged independently of those additions. I’ll follow up with  PRs for the following:
- Scam alert testing page
  - Recent contract testing
  - Previous contract transactions
  - Scam contract testing
- App isolation testing
- Transaction simulation testing

src/components/Tabs/EIPSupportTab.svelte
```diff
@@ -0,0 +1,651 @@
...
+  const totalSteps = testSteps.length;
+
+  function getStepButtonText(): string {
+    if (isRunning) return 'Running...';
+
+    if (stepStatus === 'passed') return 'Step Passed ✓';
+
+    if (stepStatus === 'partial') return 'Step Partial ⚠';
+
+    if (stepStatus === 'failed') return 'Retry Step';
```
polymutex: Maybe use 🔄 as the emoji for this one

src/types/utils/ethereum-types.ts
polymutex, line 1: Please rename this to something less generic, like `ethereum-types.ts` or something. `isRecord` can be moved to a new file like `record.ts`.

polymutex (approved):
Ship it 🫡

PR #443, opened 2025-12-27
Adding Keycard Shell data

phift:
Adding data for Keycard Shell: QR (ERC‑4527, BC‑UR), MIT license, privacy (no radio; optional USB), security (open-source + reproducible builds + hash-verified updates), key handling (on-Keycard generation and signing; restricted export) via secure element (EAL6+), transaction legibility with on-device decoding references (ABI repo)


polymutex:
This is perfect. Thanks a lot!

If you have some time, I'd like to ask how your experience was filling this in. I worry that contributing to Walletbeat is too difficult, so I am wondering what difficulties you encountered, and what would be most effective to lower the bar for contributions: more documentation, different data entry methods or formats?

polymutex (approved):

phift:
The docs are great. But even more would help agents to get it right the first try.

And the software wallet categories mostly don't work for hww. You can delete - passkey, address privacy, private tokens, RPC, ....
And add others: signing stage (signing happens inside the SE or not), how airgapped a hww is, duress pin, physical tamper resistance, ... and so much more. 
On signing alone we have a large table here: https://keycard.tech/en/blog/keycard-shell-radically-open-uniquely-secure
And https://thebitcoinhole.com/hardware-wallets/ people are really good at these details, but they focus on Bitcoin.

Issue #404, opened 2025-12-16
Feat: Add message signing to transactionLegibility


maykelxyz:
This feature extends transaction legibility to include message signing, evaluating how wallets present signed messages to users across common signature standards

- Simple message signing
  - Basic eth_sign / personal_sign style messages.
- Sign-In with Ethereum (SIWE, EIP-4361)
  - Standardized authentication messages for off-chain login flows. #203 
- Structured data signing (EIP-712)
  - Typed, domain-separated messages with explicit field visibility.

maykelxyz:
Some notes discussed with Patrick, re: message signing:
- Message signing should PASS if a hardware wallet shows either
  a - the EIP-712 struct 
  b - (domainHash & messageHash)
  c - the safeHash

It is also important that a hardware wallet MUST show this is in the hardware wallet itself and not the software around it. That being said, the attributes for transactionLegibility should be adjusted as we're still rating wallets as PASS as long as the calldata gets decoded, even if it's in the software (not the hardware).

PR #446, opened 2026-01-01
Feat: Add message signing & improve transaction legibility attribute

### LINKED ISSUE #403: Refactor: Transaction legibility attribute
At the moment, transaction legibility attributes rating has a general `details` and `howToImprove` sections, which can be improved. This issue focuses on both sections to be as detailed and objective as possible instead of just returning a general description. A more visual methodology could also be an improvement (but nothing comes to mind atm).

Atleast for `details` and `howToImprove` a suggestion made by @polymutex was to create a visual table of what's provided and what's missing

maykelxyz:
# Feat: Add message signing & improve transaction legibility attribute

## Added message signing
- Added message signing legibility to both hardware wallets and software wallets
```ts
messageSigningLegibility: {
	[MessageSigningProvides.EIP712_STRUCT] = true,
	[MessageSigningProvides.DOMAIN_HASH] = true,
	[MessageSigningProvides.MESSAGE_HASH] = true,
	[MessageSigningProvides.SAFE_HASH] = true
}
```
  - Difference between hardware wallets and software wallets is that hardware wallets should also dictate the location of message signing decoding.
```ts
messageSigningLegibility: {
    messageSigningProvides: {
    	[MessageSigningProvides.EIP712_STRUCT] = true,
	[MessageSigningProvides.DOMAIN_HASH] = true,
	[MessageSigningProvides.MESSAGE_HASH] = true,
	[MessageSigningProvides.SAFE_HASH] = true
    },
    decoded: DataDecoded.ON_DEVICE
}
```
- Message signing attribute rating
  - Walletbeat should only PASS or FAIL a wallet, there should be no PARTIAL rating for message signing legibility
  - Message signing would only pass if a wallet provides either of the three:
    a - the EIP-712 struct
    b - (domainHash & messageHash)
    c - the safeHash
  - Message signing decoding on hardware wallets **SHOULD** be provided **on the device**, and decoding outside of the device would be rated as **FAIL**

## Updated transaction legibility Details and How To Improve section
- General improvement is to make both sections more detailed, iterating what's provided, what's missing.
- For hardware wallets, added checks if a wallet decodes calldata / message signing. Explicitly mentions if a wallet should improve the location of calldata / message signing decoding 

### Other updates
- Updated `CalldataDataDecoded` to `DataDecoded` to fit message signing decoding location
- Updated existing wallets `transactionLegibility` to add `messageSigningLegibility: null`

Fixes #403 and #404

### DIFF
--- a/src/schema/attributes/security/transaction-legibility.ts
+++ b/src/schema/attributes/security/transaction-legibility.ts
@@ -23,7 +23,8 @@ import {
 	TransactionDisplayOptions,
 } from '@/schema/features/security/transaction-legibility'
 import { popRefs, refs } from '@/schema/reference'
-import { markdown, mdParagraph, paragraph, sentence } from '@/types/content'
+import { markdown, paragraph, sentence } from '@/types/content'
+import { commaListFormat } from '@/types/utils/text'
 
 import { pickWorstRating, unrated } from '../common'
 
@@ -118,11 +119,11 @@ function analyzeHardwareFeatures(
 		const decodingChecks = [
 			{
 				key: CalldataDecoding.ETH_USDC_TRANSFER,
-				label: 'Basic token transfers (ERC-20)',
+				label: 'basic token transfers (ERC-20)',
 			},
 			{
 				key: CalldataDecoding.ZKSYNC_USDC_TRANSFER,
-				label: 'zkSync token transfers',
+				label: 'ZKSync token transfers',
 			},
 			{
 				key: CalldataDecoding.AAVE_SUPPLY,
@@ -130,11 +131,11 @@ function analyzeHardwareFeatures(
 			},
 			{
 				key: CalldataDecoding.SAFEWALLET_AAVE_SUPPLY_NESTED,
-				label: 'Nested Safe transactions',
+				label: 'nested Safe transactions',
 			},
 			{
 				key: CalldataDecoding.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND,
-				label: 'Complex nested multisend transactions',
+				label: 'complex nested multisend transactions',
 			},
 		]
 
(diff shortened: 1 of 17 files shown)


src/schema/features/security/transaction-legibility.ts
```diff
@@ -193,14 +193,43 @@ export type CalldataDecodingTypes = Record<
...
+	/** The wallet provides the message hash */
+	MESSAGE_HASH = 'MESSAGE_HASH',
+	/** The wallet provides the Safe hash */
+	SAFE_HASH = 'SAFE_HASH',
+}
+
+/**
+ * For software wallets: track which message signing data types are available
+ */
+export type SoftwareMessageSigningLegibility = Record<MessageSigningProvides, boolean> | null
```
polymutex: Maybe replace `boolean` with `TransactionDisplayOptions` here? (and rename `TransactionDisplayOptions` to `DataDisplayOptions` or something.) This way we can distinguish the default vs non-default case, similar to what we do for transaction details.

polymutex: Maybe this should be named `MessageSigningDetails`, to match the naming of `DisplayedTransactionDetails`.

src/schema/features/security/transaction-legibility.ts
```diff
@@ -193,14 +193,43 @@ export type CalldataDecodingTypes = Record<
...
+
+/**
+ * For hardware wallets: track which message signing data types are available and where they are displayed
+ */
+export interface HardwareMessageSigningLegibility {
+	/** Which message signing data types does the wallet provide? */
+	messageSigningProvides: Record<MessageSigningProvides, boolean>
```
polymutex: Same comment here (`DataDisplayOptions` instead of `boolean`?)

src/schema/features/security/transaction-legibility.ts
```diff
@@ -193,14 +193,43 @@ export type CalldataDecodingTypes = Record<
...
-export enum CalldataDecoded {
+export enum DataDecoded {
 	ON_DEVICE = 'ON_DEVICE',
 	OFF_DEVICE = 'OFF_DEVICE',
 }
+
+/**
+ * What does the wallet provide for message signing legibility?
+ */
+export enum MessageSigningProvides {
```
polymutex: "provides" is a conjugated noun in the 3rd person, which is generally OK to use in names like these but I don't think any part of the codebase uses that convention. I suggest `MessageSigningDetail` or `SignatureInfo` or something.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -29,8 +35,329 @@ export type TransactionLegibilityValue = Value & {
...
+		decodedLocation: DataDecoded | null
+	}
+}
+
+function analyzeHardwareFeatures(
+	legibility: HardwareTransactionLegibilityImplementation['legibility'],
+	detailsDisplayed: HardwareTransactionLegibilityImplementation['detailsDisplayed'],
+	dataExtraction: HardwareTransactionLegibilityImplementation['dataExtraction'],
+	messageSigningLegibility: HardwareTransactionLegibilityImplementation['messageSigningLegibility'],
+): HardwareFeatureDetails {
```
polymutex: Wouldn't it be simpler to just take a single `HardwareTransactionLegibilitySupport` argument?
Same question for other functions below that take similar sets of arguments...

You could also use argument destructuring, like this:

```typescript
function analyzeHardwareFeatures( {
  legibility,
  detailsDisplayed,
  dataExtraction,
  messageSigningLegibility
}: HardwareTransactionLegibilitySupport): HardwareFeatureDetails {
  ...
}
```

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -29,8 +35,329 @@ export type TransactionLegibilityValue = Value & {
...
+			sections.push(`✓ Supported: ${commaListFormat(features.messageSigning.supported)}\n`)
+		}
+
+		if (features.messageSigning.missing.length > 0) {
+			sections.push(`✗ Missing: ${commaListFormat(features.messageSigning.missing)}\n`)
+		}
+	}
+
+	return sections.join('\n')
+}
```
polymutex, line 319: Great to have more detailed markdown for this. For future components however, when the logic for expressing the rating becomes this complex, it may be a good time to add a custom component (similar to attributes like security audits).

No need to change this in this PR, but let me know if you'd like to work on this next and I can write something on how to use custom components like `src/types/content/security-audits-details.ts`.

polymutex (approved):
Looks good. Mostly just typing and naming suggestions, good to merge once addressed

src/schema/attributes/security/transaction-legibility.ts
polymutex: (The logic you've already written for generating `HardwareFeatureDetails` would be the same btw, the only logic that would need to change is this function which would be expressed as a Svelte component instead of a markdown string builder function.)

maykelxyz: That would be great, I was originally doing a table MD but for some reason, it was not compatible, or I'm missing a key component of making it work. A guide for the custome components would be nice :)

src/schema/features/security/transaction-legibility.ts
```diff
@@ -193,14 +193,43 @@ export type CalldataDecodingTypes = Record<
...
+	/** The wallet provides the message hash */
+	MESSAGE_HASH = 'MESSAGE_HASH',
+	/** The wallet provides the Safe hash */
+	SAFE_HASH = 'SAFE_HASH',
+}
+
+/**
+ * For software wallets: track which message signing data types are available
+ */
+export type SoftwareMessageSigningLegibility = Record<MessageSigningDetails, boolean> | null
```
polymutex: `DataDisplayOptions` instead of `boolean` here too.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -29,8 +35,328 @@ export type TransactionLegibilityValue = Value & {
...
+	}
+
+	return details
+}
+
+function generateHardwareDetailsMarkdown(features: HardwareFeatureDetails): string {
+	const sections: string[] = []
+
+	// Calldata Decoding section
+	// Calldata Decoding section
```
polymutex: Duplicated comment

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -422,14 +971,25 @@ export const transactionLegibility: Attribute<TransactionLegibilityValue> = {
 					The hardware wallet implements full transaction legibility, displaying all
 					transaction details on the hardware device screen for verification before signing.
 				`),
-				hardwareFullTransactionLegibility(),
+				hardwareFullTransactionLegibility({
+					legibility: null,
+					detailsDisplayed: null,
+					dataExtraction: null,
+					messageSigningLegibility: null,
+					ref: refTodo,
```
polymutex: Please use `refNotNecessary` instead of `refTodo` when putting `ref` inside attribute rating examples. This way, searching the codebase for "`refTodo`" yields results that actually need references, without turning up these example ratings :)

polymutex (approved):

PR #450, opened 2026-01-05
Feat/news section

### LINKED ISSUE #231: Wallet attribute: Security history
This issue is about adding a new Security attribute which focuses on a wallet's history of security (or of hacks).

This is especially relevant for hardware wallets, but we should try to make the attribute apply to software wallets as well if possible.

For L2s or smart contract, "security history" can often be measured in terms of "area under the curve" of TVL. For wallets, there is no direct analog to this, but hopefully we can find something similarly empirical in nature.

Q: Should we track hacks of _wallets_ specifically, or also include hacks of _wallet development companies_?  
Argument: Hacks of _wallet development companies_ may be a reflection on their security practices in general. They have implications for e.g. software supply chain security (what if their build machine gets compromised?). Additionally, depending on the nature of the hack, they may also have user implications (e.g. Ledger database hack which made public the street addresses of many large crypto holders).

maykelxyz:
Fixes #231

### DIFF
--- a/src/constants/navigation.ts
+++ b/src/constants/navigation.ts
@@ -19,6 +19,7 @@ import GithubIcon from 'lucide-static/icons/github.svg?raw'
 import KeyIcon from 'lucide-static/icons/key.svg?raw'
 import MessageCircleIcon from 'lucide-static/icons/message-circle-heart.svg?raw'
 import WalletIcon from 'lucide-static/icons/wallet.svg?raw'
+import NewsPaperIcon from 'lucide-static/icons/newspaper.svg?raw'
 
 // Constants
 import { hardwareWallets } from '@/data/hardware-wallets'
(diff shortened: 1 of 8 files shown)


src/constants/navigation.ts
```diff
@@ -70,6 +71,13 @@ export const navigationFarcasterChannel = {
 	href: 'https://farcaster.xyz/~/channel/walletbeat',
 } as const satisfies NavigationItem
 
+export const navigationNews = {
+	id: 'news',
+	icon: NewsPaperIcon,
+	title: 'Relevant News',
```
polymutex: Maybe "Wallet news"?

src/types/content/news.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * Category of security-related news event
+ */
+export enum NewsType {
+	/** Security breach resulting in unauthorized access or theft */
+	HACK = 'HACK',
+	/** Unauthorized access or exposure of user data */
+	DATA_BREACH = 'DATA_BREACH',
```
polymutex: What's the difference between these?

polymutex (approved):
Looks great! Ship it :)

PR #464, opened 2026-01-12
Add Zeus

greekfetacheese:
This PR adds [Zeus](https://github.com/greekfetacheese/zeus) to walletbeat


polymutex:
Thanks for contributing such a thorough dataset. Sorry about the CI failure, it's broken at HEAD so not your fault. Will review this in a day or two.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+			websites: ['https://github.com/greekfetacheese/zeus'],
+		},
+	},
+	features: {
+		accountSupport: {
+			defaultAccountType: AccountType.eoa,
+			eip7702: supported({
+				ref: refTodo,
+				contract: 'UNKNOWN',
+			}),
```
polymutex: Is it actually supported? Can't seem to tell from the README. I found [this file](https://github.com/greekfetacheese/zeus/blob/master/src/utils/zeus_delegate.rs) which may be relevant, but would like to see the smart contract code behind it if any.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+					crossChainSumView: notSupported,
+					perChainBalanceViewAcrossMultipleChains: featureSupported,
+				},
+			},
+		},
+		chainConfigurability: supported({
+			ref: [
+				{
+					explanation:
+						'Zeus uses public RPC endpoints by default but allows users to provide custom endpoints and disable or delete what they don`t want.',
```
polymutex: Please use the `'` character in `don't`, not a backtick.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+		privacy: {
+			appIsolation: {
+				[Variant.DESKTOP]: {
+					createInAppConnectionFlow: notSupported,
+					erc7846WalletConnect: notSupported,
+					ethAccounts: supported({
+						ref: [
+							{
+								explanation: 'Zeus exposes the address of the active account only.',
+								url: 'https://github.com/greekfetacheese/zeus/blob/master/src/server.rs#L367',
```
polymutex: Please work in the commit hash in the URL when linking to specific line numbers, otherwise this can get out of date as the file is updated.
In this case, I think the link is `https://github.com/greekfetacheese/zeus/blob/8d51c76d1c6ccce5a4a845c34429a4f89ff9cdae/src/server.rs#L371`.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+							},
+						],
+						defaultBehavior: ExposedAccountsBehavior.ACTIVE_ACCOUNT_ONLY,
+					}),
+					useAppSpecificLastConnectedAddresses: notSupported,
+				},
+				[Variant.MOBILE]: null,
+				[Variant.BROWSER]: null,
```
polymutex: You should be able to remove these lines. If the typesystem complains, that's a bug on us...

Alternatively, you can remove the nesting inside `[Variant.DESKTOP]` above and just set its content to `appIsolation` directly, since Zeus only has one variant.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+				ethereumL1: notSupported,
+			},
+			passkeyVerification: notSupported,
+			publicSecurityAudits: null,
+			scamAlerts: {
+				contractTransactionWarning: supported({
+					ref: [
+						{
+							explanation:
+								'Zeus currently does not have a scam alert mechanism, It simply shows with which contract you are interacting with. If it is a known contract a hyperlink with the contract`s name is shown otherwise a truncated version of the contract address is shown (hyperlink). The user can also see all the decoded events to inspect the transaction.',
```
polymutex: Same comment on the apostrophe (use `'`)

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+							label: 'Contract interaction is shown in the transaction details',
+							url: 'https://github.com/greekfetacheese/zeus/blob/master/src/gui/ui/tx_window.rs#L246C1-L247C1',
+						},
+					],
+					contractRegistry: true,
```
polymutex: The explanation above doesn't sound like there's a contract registry (local or otherwise) that contract addresses are checked against, right? In that case, it should be `false`.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+						{
+							label: 'Before every transaction the user must confirm the action.',
+							url: 'https://github.com/greekfetacheese/zeus/blob/master/src/utils/tx.rs#L241C1-L242C1',
+						},
+					],
+					leaksRecipient: false,
+					leaksUserAddress: false,
+					leaksUserIp: false,
+					newRecipientWarning: false,
+					userWhitelist: true,
```
polymutex, line 265: `userWhitelist` refers to a feature where the user can maintain an editable whitelist, and where the wallet warns the user specifically that the address they're sending tokens to is not on that list. I may have missed it, but a cursory look at [this file](https://github.com/greekfetacheese/zeus/blob/master/src/gui/ui/tx_window.rs) suggests there is no such feature.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+				},
+				l2: {
+					[TransactionSubmissionL2Type.arbitrum]: null,
+					[TransactionSubmissionL2Type.opStack]: null,
+					ref: [
+						{
+							explanation: 'Zeus submits L2 transactions via RPC endpoints.',
+							url: 'https://github.com/greekfetacheese/zeus#features',
+						},
+					],
```
polymutex: Seems like you can populate the `[TransactionSubmissionL2Type.arbitrum]` and `[TransactionSubmissionL2Type.opStack]` fields here with one of the `TransactionSubmissionL2Support.SUPPORTED_*` values

polymutex (commented):
Thanks again for the PR. Zeus looks pretty cypherpunk :)

polymutex:
Please rebase after addressing the above, I believe this should fix the CI failure.

polymutex:
btw, you may also want to review https://github.com/walletbeat/walletbeat/pull/472 for the social media post! Please provide the X/Farcaster handle if any and if you'd like to be mentioned in it

data/software-wallets/zeus.ts
greekfetacheese: The module you pointed is WIP and not used in production and i dont think im planning to.
Users can delegate to whatever contract they want and undelegate at any time.
Is this still considered EIP7702 supported?

data/software-wallets/zeus.ts
greekfetacheese: I changed to "dont" because with the ' i get a parsing error

data/software-wallets/zeus.ts
greekfetacheese: Fixed

data/software-wallets/zeus.ts
greekfetacheese: fixed

data/software-wallets/zeus.ts
greekfetacheese: fixed

data/software-wallets/zeus.ts
greekfetacheese: Users maintain a Contact list, when they send crypto or bridge ETH which will choose a recipient if the address is not known they get a wanring before even opening the TxConfirm window. Isn't that correct? link: https://github.com/greekfetacheese/zeus/blob/7bd2a133344bfa5b52dfe0df6e8864839cbbaee9/src/gui/ui/send_crypto.rs#L229

data/software-wallets/zeus.ts
greekfetacheese: Updated with SUPPORTED_BUT_NO_FORCE_INCLUSION  i hope its the correct one

data/software-wallets/zeus.ts
polymutex: No, this is about the wallet making an opinionated choice about which EIP7702 contract to delegate to in order to endow the user's account with smart-account-ish features. If the wallet doesn't do this, then this should be `notSupported`.

data/software-wallets/zeus.ts
polymutex: My bad, I missed this. Yes, this should remain `userWhitelist: true` then.

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,386 @@
...
+					crossChainSumView: notSupported,
+					perChainBalanceViewAcrossMultipleChains: featureSupported,
+				},
+			},
+		},
+		chainConfigurability: supported({
+			ref: [
+				{
+					explanation:
+						'Zeus uses public RPC endpoints by default but allows users to provide custom endpoints and disable or delete what they dont want.',
```
polymutex: You can use `don't` with a regular `'` character, simply use double quotes for the string:

```
explanation: "Zeus uses public RPC endpoints by default but allows users to provide custom endpoints and disable or delete what they don't want.",
```

greekfetacheese:
@polymutex fixed the last issues

polymutex (approved):

data/software-wallets/zeus.ts
```diff
@@ -0,0 +1,383 @@
...
+							url: 'https://github.com/greekfetacheese/zeus/blob/master/src/gui/ui/tx_window.rs#L246C1-L247C1',
+						},
+					],
+					contractRegistry: false,
+					leaksContractAddress: false,
+					leaksUserAddress: false,
+					leaksUserIp: false,
+					previousContractInteractionWarning: false,
+					recentContractWarning: false,
+				}),
```
polymutex: The test is failing with `Error: Wallet zeus: contractTransactionWarning: If supported, at least one implementation mechanism must be enabled`. I think `contractTransactionWarning` should be set to `notSupported` rather than all of its members being set to `false`.
To preserve the `ref` field, you can use:

```typescript
contractTransactionWarning: notSupportedWithRef<ContractTransactionWarning>({
	ref: [
		{
			explanation:
				'Zeus currently does not have a scam alert mechanism, It simply shows with which contract you are interacting with. If it is a known contract a hyperlink with the contracts name is shown otherwise a truncated version of the contract address is shown (hyperlink). The user can also see all the decoded events to inspect the transaction.',
			label: 'Contract interaction is shown in the transaction details',
			url: 'https://github.com/greekfetacheese/zeus/blob/master/src/gui/ui/tx_window.rs#L246C1-L247C1',
		},
	],
})
```

polymutex (approved):

data/software-wallets/zeus.ts
greekfetacheese: Fixed

polymutex (approved):

polymutex:
Thanks for the PR! It's live now: https://beta.walletbeat.eth.limo/zeus/

PR #467, opened 2026-01-12
Feat: Add Walkaway Test Tweet

maykelxyz:


polymutex:
Content looks good, but I think this is missing a section around how we translate this test into concrete wallet attributes as already implemented on Walletbeat. (Will write a more detailed review in a few hours)

resources/posts/01-14-2025 - Walkaway Test.md
```diff
@@ -0,0 +1,49 @@
...
+
+The Walkaway Test
+
+-----
+
+Tweet #2
+
+"What is the Walkaway Test?"
+
+Simply put, If the team stopped working on the wallet tomorrow, how long could it continue to function safely and independently?
```
polymutex: `If` -> `if`

polymutex: It's not really "how long", it's more about whether it can continue to function safely at all. All wallets will rot in time; even a perfect wallet will still rot just from security bugs in their underlying libraries being found. What the walkaway test emphasizes is whether the wallet is designed to withstand such entropy as long as possible. That last sentence isn't really tweetable without it going over people's heads though. Maybe something like:

"Simply put: if the team stopped working on the wallet tomorrow and shut down all its surrounding infrastructure, would users still be able to use it? Would any of its security or self-sovereignty features degrade?"

or something; I don't think the above is perfect either...

resources/files/social-media/threads/2025-01-13 - Walkaway Test/cover.png
polymutex, line 1: I think it's better to put these in the same directory as the post they're about. I suggest something like `resources/files/social-media/threads/${yyyy}/${yyyy}-${mm}-${dd}` as a directory, and then put both the posts and images in that directory. Check the `governance/minutes` directory as inspiration.

Definitely don't use MM-DD-YYYY date format, as that would be sorted poorly in 2027 :)

resources/posts/01-14-2025 - Walkaway Test.md
```diff
@@ -0,0 +1,49 @@
...
+
+-----
+
+Tweet #3
+
+The test looks at whether users are protected without relying on continuous team intervention.
+No updates. No emergency servers. No manual overrides.
```
polymutex: I'd put the tweet 4 above this one, i.e. explain the problems it's trying to measure before it explains how to measure them.
For example:

```
The test encapsulates many security and self-sovereignty attributes under the same test. Tough questions such as:

- Does the wallet critically rely on any centralized service?
- Can users still transact trustlessly?
- Can users still port their account to other wallets permissionlessly?
```

resources/posts/01-14-2025 - Walkaway Test.md
```diff
@@ -0,0 +1,49 @@
...
+
+– Are critical services centralized?
+
+– Can users still sign, recover, and transact trustlessly?
+
+-----
+
+Tweet #5
+
+Why it matters: wallets are long-lived infrastructure.
```
polymutex: Maybe worth making the analogy to browsers here.

"If web browsers help you navigate web2, wallets help you navigate web3. You would expect Firefox to keep working even if Mozilla goes out of business. Similarly, users should not depend on a development team's availability, funding, or goodwill to retain control of their account."

(p.s. notice I've personally tried to use the word "account" instead of "assets" or anything financial-related in general. Not a strong rule, but I think it's good to not over-emphasize on financial use-cases.)

resources/posts/01-14-2025 - Walkaway Test.md
```diff
@@ -0,0 +1,49 @@
...
+Users should not depend on a team’s availability, funding, or goodwill to retain control of their assets.
+
+-----
+
+Tweet #6
+
+Wallets that pass it move closer to one of Ethereum's core values: **self-sovereignty**.
```
polymutex: This is where I think it'd be good to perhaps link to the L1 RPC provider independence thread: https://x.com/walletbeat/status/2002818293741301777

But if so, also mention that this is just _one_ prong of the walkaway test, and that there are others (e.g. the account recovery attribute).

That reminds me that we should probably write a thread on the Account Recovery/Unruggability attribute...

polymutex (approved):
This is great. tbh I think you can post these as-is without review; I trust you, and I don't think we should have a super heavy review process for simple comms at this stage since we're not that big yet.

It'll be more important later though, so going through the exercise of having this process in place for when we need it is a good thing too.

The comments below are all optional. (Except the one on the YYYY-MM-DD format)

resources/posts/01-14-2025 - Walkaway Test.md
maykelxyz: Lmk if you need any visuals for posting a thread about the Account Recovery / Unruggability. I assume you want to post the thread, given you were the one who recently had the feature / attribute up :)

resources/posts/01-14-2025 - Walkaway Test.md
maykelxyz: Using:

Simply put: if the team stopped working on the wallet tomorrow and shut down its infrastructure, would users still be able to use it? Would its security or self-sovereignty features degrade?

maykelxyz:
> I don't think we should have a super heavy review process for simple comms at this stage since we're not that big yet.

I agree. I'll be keeping a track of this just for safekeeping. As you mentioned, relying on X isn't that much of a great idea when it comes to preserving content

resources/posts/01-14-2025 - Walkaway Test.md
polymutex: Well, it'd be posted under the Walletbeat X account directly, right? So not really attributed to you or I.

PR #469, opened 2026-01-13
ESLint–Svelte + CI/CD fixes / improved <Table> implementation

darrylyeo:
## CI/CD

* Packages:
  * update all

* ESLint:
  * reduce set of ignored files to parsing errors only

* Prettier:
  * ignore some files and directories

* cspell:
  * ignore some files and directories

## Data

* Stages:
  * use `WeakMap` to associate evaluation functions with attribute IDs for serialization

## Views

* `<WalletStageSummary>`:
  * improve data derivation
  * assign `{#each}` key

## Components

* `<Table>`:
  * improve reactivity
  * internalize state
  * improve generic type logic
    * avoid enforcing `_CellValue` (TypeScript cannot discriminate different cell value return types within the same array of `Column`s)
  * introduce `enum SortDirection`

* `<Select>`:
  * assign {#each} key

* `<BlockTransition>`:
  * fix snippet `value` type


polymutex (approved):
🎊

PR #471, opened 2026-01-14
Feat: Add news to wallets page

maykelxyz:
Updates:
- Created indexes for news per wallet, instead of news being populated with multiple news files
```ts
export const ledgerNewsRegistry = {
	'ledger-global-e-breach': globalEBreach,
} as const
```
@darrylyeo refactored this before into separate news per file but we needed a way to have a registry or list of news **per wallet**. Feel free to refactor this, but we need to retain the **per-wallet** news.
- Added new UI to show the news per wallet. Probably needs more work, also tagging @darrylyeo for this, as this is the first iteration of the news feature / UI
image


src/schema/wallet.ts
```diff
@@ -228,6 +229,9 @@ export interface BaseWallet {
 
 	/** Overrides for specific attributes. */
 	overrides?: WalletOverrides
+
+	/** Security news items associated with this wallet. */
+	news?: WalletSecurityNews[]
```
polymutex: I think @darrylyeo's approach is more correct. Semantically speaking, "news" are not an _inherent property of_ wallets, "news" are their own things that are _associated with_ wallets. So from a typesystem and code layout perspective, I think it makes more sense to have a flat list of news that's separate from the wallets, and then we can create associations between the two. And practically speaking, this also enables the possibility of news that affects multiple wallets at once (for example, two hardware wallets that use the same payment processor that just suffered a security breach, or many software wallets using the same cryptographic library in which a new security flaw is found).

So I suggest removing this field from `BaseWallet`, and instead adding a `wallets: NonEmptyArray<string>` field to the `WalletSecurityNews.impact` type, which would contain a list of impacted wallet IDs. Then we can have a helper function that returns a chronologically-sorted `WalletSecurityNews[]` for a given wallet ID, and use that on the wallet page.

You can also vibecode a consistency test in `tests/` to ensure that the strings provided in this field are valid wallet IDs and don't have any duplicates; include `tests/wallet-integrity.test.ts` as an example test in the agent context and it'll probably get it right.

src/views/WalletPage.svelte
```diff
@@ -423,6 +430,127 @@
 			</section>
 		</header>
 
+		{#if wallet.news && wallet.news.length > 0}
```
polymutex: One idea here: This section's importance changes over time, and therefore its visual prominence should too. If a security incident is ongoing, it's very important to show it at the top. If there's a recent security incident but it is resolved, it's also important to show it at the top by default. But if there's been no news for a long time, having this at a top seems like a "scar" on the wallet page that the wallet developers can never get rid of no matter what they do, which seems unfair.

I suggest either (or both):
- Make the section collapsed by default if all news are resolved and the latest one is over 30 days old.
- Move the section to the bottom if all news are resolved and the latest one is over a year old.
- When the section is shown expanded-by-default, add an explanation in small text somewhere that says something like "This section will remain expanded until all ongoing incidents are resolved and 30 days have passed" or similar.

polymutex:
I think your version of prettier is the one from before the last `package.json` update. You may need to run `pnpm install` and `pnpm fix`.

src/schema/wallet.ts
maykelxyz: Hmm, one solution I had is to create the:

```ts
export function getNewsForWallet(walletId: string): WalletSecurityNews[] {
	return allWalletSecurityNews.filter((news) => news.wallet?.id === walletId)
}
```

And use it in:

```ts
	const walletNews = $derived(
		getNewsForWallet(wallet.metadata.id)
	)
```

I think it's essential to have an impacted wallet's metadata in `WalletSecurityNews` as we show it individually in the `/news` page. logo, name, link, etc. 

```ts
/** Represents a security-related news item about a wallet */
export type WalletSecurityNews = WithRef<{
	/** Identifier slug for the news item */
	slug: string
	/** Category of security event */
	type: NewsType
	/** Severity level of the incident */
	severity: Severity

	/** Display title for the news item */
	title: string
	/** Brief description of the incident */
	summary: string
	/** Wallet involved */
	wallet?: WalletMetadata
```

Let me know if you have any comments or if you think this would be sufficient too

src/views/WalletPage.svelte
```diff
@@ -429,6 +477,140 @@
 			</section>
 		</header>
 
+		{#if walletNews.length > 0 && !newsIsVeryStale}
+			<hr />
+			<div data-scroll-item="inline-detached padding-match-end" data-column>
+			{@render securityNewsSection()}
```
polymutex: Perhaps "SecurityNews" should be its own Svelte component, rather than a function here.

That would also help with keeping the CSS rules nicely encapsulated.

src/schema/wallet.ts
polymutex: I still think this should be expanded to allow for multiple wallets to be associated with a single piece of news. The `/news` page can look up the `WalletMetadata` by ID using the `allWallets` map as needed (you can add a function to make that easy too)

polymutex: (This can be a separate PR from this one though)

src/views/SecurityNews.svelte
```diff
@@ -0,0 +1,272 @@
...
+				<data
+					data-badge="medium"
+					value="INFO"
+				>INFO</data>
+			</header>
+		</summary>
+
+
+		<div class="attribute-accordions news-items" data-column>
+			{#each news as newsItem}
```
polymutex: As per lint warning, add block key

polymutex (approved):
Good to merge!

src/views/SecurityNews.svelte
maykelxyz: Fixing. Will merge once CI is ok :) thanks!

PR #481, opened 2026-01-20
Feat: Add Wallet Security News post

maykelxyz:


resources/files/social-media/threads/2025-01-20 - Wallet Security News Update/thread.md
```diff
@@ -0,0 +1,15 @@
...
+A new way to surface important security-related events tied to wallets.
+
+Tweet #2:
+
+Security hacks are not new, and they shouldn't be ignored.
+
+Walletbeat helps users stay informed about the latest wallet security news: hacks, vulnerabilities, data breaches, and other incidents.
+
+See more (then the link)
+
```
polymutex, line 45: Would be good to include some examples of wallet-relevant security news and why they're relevant to show on the wallet page.

Then also perhaps some details on when this is displayed and when it goes away after 30d or something.

maykelxyz: I'll add some examples (photos) of what we currently have. Though for details, I plan to amplify the tweet through a much more detailed and technical breakdown, so might not include it here, but will still do that

polymutex (approved):

PR #497, opened 2026-01-30
Typography: fix / narrow types

darrylyeo:
## Components

* `<Typography>`: update generic logic

## Utils

* Typography: overload helper functions to infer available string interpolations directly from strings

## Data

* narrow typography props based on actual usage

### DIFF
--- a/src/schema/attributes.ts
+++ b/src/schema/attributes.ts
@@ -1,18 +1,17 @@
 import type { Content, Paragraph, Sentence, TypographicContent } from '@/types/content'
 import { type NonEmptyArray, nonEmptyMap, type NonEmptyRecord } from '@/types/utils/non-empty'
 
-/** Strings for evaluation typographic content: none, {{WALLET_NAME}}, or name + pseudonym placeholders. */
-export type WalletEvaluationStrings =
-	| null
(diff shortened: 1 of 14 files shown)


src/schema/attributes.ts
```diff
@@ -1,5 +1,18 @@
...
+export type WalletNameStrings = null | { WALLET_NAME: string }
+
+/** Strings for content that may use {{WALLET_NAME}} and/or pseudonym placeholders. */
+export type WalletNameAndPseudonymStrings =
+	| WalletNameStrings
+	| {
+			WALLET_NAME: string
+			WALLET_PSEUDONYM_SINGULAR: string | null
+			WALLET_PSEUDONYM_PLURAL: string | null
+	  }
```
polymutex, line 14:
```typescript
export type WalletNameAndPseudonymStrings = WalletNameStrings & (
	{
		WALLET_PSEUDONYM_SINGULAR: string | null
		WALLET_PSEUDONYM_PLURAL: string | null
	} | {}
)
```

I think it makes it more obvious that it's purely extending `WalletNameStrings`.

polymutex (approved):

PR #500, opened 2026-02-01
modified the period in into an exclamation point in the first sentence of the redme.md file

ren2140eth:
test


polymutex:
I will close this PR since this was just for a test, but all good on the environment setup front :)

Issue #190, opened 2025-06-23
Wallet attribute: Transaction simulation


polymutex:
One of the wallet attributes we want to track is the ability for a wallet to faithfully simulate the effect of transactions, as a user safety feature.

We should come up with a set of reasonable "benchmark" transactions and test that wallet correctly simulate their effects, saved as wallet feature data:

- Sending Ether
- Sending ERC-20 tokens or NFTs
- Signing a Safe wallet transaction
- Some DeFi transactions like depositing into Aave or taking a loan from it
- Others?

Then this can be turned into an attribute.

jadjad1289DDFD:
##  State transition & deterministic simulation

 In the context of large phishing campaigns and drainers-as-a-service, a user’s intent can (and should) be defined by the simulated outcome of a transaction: if a deterministic simulation yields the outcome the user expects, that outcome becomes the ground truth of the user’s intent and a safe basis for signing decisions. 

A state transition is the deterministic update of Ethereum’s world state when a transaction (or block) is applied , given the same prior state and identical inputs, the EVM must produce a single, exact resulting state. Therefore a simulation is only trustworthy when it replicates every environment input the EVM or contract can observe; any mismatch (even subtle ones) can change the execution trace, gas used, return data, or whether the call succeeds or reverts, breaking reproducibility and undermining intent validation.
Among essential environment items to replicate, there are block header fields which are:

block.number 

block.timestamp 

baseFee 

lastblockhash  

gasLimit :  the gas limit discrepancy between wallet simulations and real transactions

gasprice 

####  How Exploit  leverages  environment replication discrepancy between wallet simulations and real transactions to deceive users and drain wallets : 

 malicious contract have a  receive() or fallback() function contains logic that implements dual execution paths:
if (msg.gas <= _receive) {
// LOW GAS PATH - Real transactions
0x82a(msg.value); // Steals ETH
exit;
} else {
// HIGH GAS PATH - Simulations only
msg.sender.call().value(msg.value); // Shows fake refund
}
In Simulation (HIGH GAS PATH):
Gas Used in simulation: 0xf0cb98 = 15,780,760 gas
Threshold Comparison: 15,780,760 > 15,000,000
Result: Fake refund displayed to user in wallet simulation
User Perception: "Contract is safe"

<img width="757" height="628" alt="Image" src="https://github.com/user-attachments/assets/a6ce6d44-8b20-4260-bbd6-ec02f9f1d8ff" />
[...]

PR #512, opened 2026-02-05
Wallet Attribute: Add transaction simulation to transaction legibility

maykelxyz:
# Transaction Simulation and Transaction Legibility Refactoring #190 

## Transaction Simulation

Benchmark transactions are now grouped into three categories:

- **`BasicBenchmarkTransactions`** — already existed. Plain transfers (ETH, ERC-20, ERC-721, ZKSync USDC).
- **`ComplexBenchmarkTransactions`** *(new)* — contract interactions (USDC approval, Aave supply, SafeWallet nested, SafeWallet multisend batch). Adds `calldataDecoded` and `transactionOutcome` since there's no simple "to" or "value" to show.
- **`SimulationBenchmarkTransactions`** *(new)* — simulation-specific edge cases:
  - `FAILED_TRANSACTION` — does the wallet warn the user the tx will revert? (`failure: 'DETECTED' | 'NOT_DETECTED'`)
  - `NONDETERMINISTIC_TRANSACTION` — does the wallet catch that the outcome depends on execution state? (`nondeterminism: 'NOT_DETECTED' | 'DETECTED_WITHOUT_WARNING' | 'DETECTED_WITH_WARNING'`)

Also added `DisplayedTokenTransferDetails` — extends `DisplayedBasicTransactionDetails` with `transactionOutcome` for ERC-20 and ERC-721 transfers, since those go through contracts too.

### Software Wallet: Removed legibility from SoftwareTransactionLegibilitySupport

- Folded it into transactionDetailsDisplay & renamed to `calldataDecoded`.
- Changed type to `DataDisplayOptions` instead of boolean.
- It'll only be a required property for `ComplexBenchmarkTransactions` like USDC Approval, AAVE supply, Safewallet AAVE Supply nested & Safewallet USDC Approval & AAVE supply batch nested.

### Hardware Wallet: Updated `legibility` to `calldataDecoded`

- In line with software wallet's change, `legibility` is updated to `calldataDecoded`
- Instead of HardwareCalldataDecoding being:

```ts
export type CalldataDecodingTypes = Record<
	CalldataDecoding,
	Support<WithRef<CalldataDecodingSupport>>
>
```
changed to:
```ts
export type CalldataDecodingTypes = Record<HardwareBenchmarkTransactions, DataDecoded | null>
```
to match software wallet's type not wrapped in `<Support>`
- added `NOT_DECODED` to `enum DataDecoded` since we don't wrap it in support anymore.

Few other notes:
- `transactionOutcome`, `failure` and `nondeterminism` was not added as part of hardware wallets' features. Hardware wallets are inherently offline/air-gapped they have no network access and can't execute EVM simulation themselves. And any simulation data displayed on a hardware wallet screen necessarily comes from the companion app (Ledger Live, Trezor Suite, etc.) i.e., off-device. Should we still track track them for hardware wallets given this circumstance?

## Updated Transaction Legibility Rating

**Software wallets:**

Requirement to be PARTIAL (i.e., not FAIL):
- Raw hex calldata display
- All 6 basic fields (gas, nonce, from, to, chain, value) shown for every benchmark tx
- `calldataDecoded` shown for USDC approval and Aave supply

Requirement to PASS:
- Everything above, plus formatted calldata + copy-to-clipboard
- Outcomes explained for ERC-20, ERC-721, USDC approval, Aave supply
- Safe transactions fully decoded + outcome explained
- Simulation benchmarks: failed tx detected, nondeterministic tx detected with warning
- Message signing data provided and passes (EIP-712 struct, or domain+message hash, or safe hash)

The Safe transactions are the hardest — they're deeply nested and require decoding multiple layers of calldata. Not having them decoded won't fail you, but you need both `calldataDecoded` and `transactionOutcome` explained to PASS.

## Other changes:
- Added `ETH` to .cspell.json
- Updated Ambire's transaction-legibility.

Edit: Updated PR description


polymutex:
Some thoughts:

- `sendReceiveEther` says "Does the wallet simulate sending and receiving Ether accurately?". What does it mean to simulate "receiving Ether"?
  - If this means "show Ether being sent during token sends", then: (1) that's not the "receiving" part, and (2) I don't see the point of transaction simulation here, because it should be obvious to the user from the interface that tokens are about to be sent, the user doesn't need a simulation feature to tell it that.
  - If this means "show Ether being sent or received during contract interaction", then what exactly is `contractInteractions` representing?
- `contractInteractions` is too imprecise as a single boolean. I think we need a rubric of sample transactions, much like we have for transaction legibility. In fact, we might as well make them the same set of transactions. It's one thing to support uniswap swaps, it's another to support arbitrary nested multicalls that each do token operations.

Maybe the approach then should be more around, again, extending the transaction _legibility_ features.
From the user's perspective, I think it actually doesn't really matter whether the information shown by the transaction details UI of the wallet is from the wallet deducing that directly, or from a transaction simulation feature. What matters is that the transaction is legible, i.e. that the user sees what's going to happen as a result of the transaction, without necessarily being from simulation.

In that light, how about this:

- We split `CalldataDecoding` into two enums: the ones about basic operations (ETH/token transfers) and the ones about more complex things (USDC approval, Aave deposit, Safe transactions). We name these `BasicBenchmarkTransactions` and `ComplexBenchmarkTransactions`. `CalldataDecoding` just becomes the union of the two enums.
- We add a new enum `SimulationBenchmarkTransactions` containing the following:

```typescript
export enum SimulationBenchmarkTransactions {
  /** A transaction that will fail. */
  FAILED_TRANSACTION = '...',

  /** A transaction that has nondeterministic outcome. */
  NONDETERMINISTIC_TRANSACTION = '...',
}
```

- We rename `DisplayedTransactionDetails` to `DisplayedBasicTransactionDetails`
- We add:

```typescript
export enum TransactionOutcome {
  /** The effect of the transaction is clearly explained. */
  EXPLAINED = '...',

  /** The effect of the transaction is not explained or unclear, requiring manual user intervention to understand (e.g. interpret calldata). */
  NOT_EXPLAINED = '...',
}

/** Details for a complex transaction. */
export interface DisplayedComplexTransactionDetails {
	from: DataDisplayOptions
	gas: DataDisplayOptions
	nonce: DataDisplayOptions
	chain: DataDisplayOptions
	transactionOutcome: TransactionOutcome
}
```

- And finally, we use this in the transaction legibility features:

```typescript
	/**
	 * Does the software wallet support displaying the transaction details?
	 */
	transactionDetailsDisplay: (
		Record<SimpleBenchmarkTransactions, DisplayedBasicTransactionDetails>
		& Record<ComplexBenchmarkTransactions, DisplayedComplexTransactionDetails>
		& {
			[SimulationBenchmarkTransactions.FAILED_TRANSACTION]: DisplayedComplexTransactionDetails & {
				failure: 'DETECTED' | 'NOT_DETECTED'
			}
		}
		& {
			[SimulationBenchmarkTransactions.NONDETERMINISTIC_TRANSACTION]: DisplayedComplexTransactionDetails & {
				nondeterminism: 'NOT_DETECTED' | 'DETECTED_WITHOUT_WARNING' | 'DETECTED_WITH_WARNING'
			}
		}
	) | null
```

Looks pretty complex, maybe this can be broken into separate type declarations to make it easier to understand. But at the end of the day, this makes the wallet feature data look like this:

```typescript
			transactionLegibility: {
				ref: refTodo,
				calldataDisplay: {
					copyHexToClipboard: false,
					formatted: false,
					rawHex: true,
				},
				messageSigningLegibility: {
					[MessageSigningDetails.EIP712_STRUCT]: DataDisplayOptions.SHOWN_BY_DEFAULT,
					[MessageSigningDetails.DOMAIN_HASH]: DataDisplayOptions.NOT_IN_UI,
					[MessageSigningDetails.MESSAGE_HASH]: DataDisplayOptions.NOT_IN_UI,
					[MessageSigningDetails.SAFE_HASH]: DataDisplayOptions.NOT_IN_UI,
				},
				transactionDetailsDisplay: {
					[SimpleBenchmarkTransactions.ETH_TRANSFER]: {
						chain: DataDisplayOptions.SHOWN_BY_DEFAULT,
						from: DataDisplayOptions.SHOWN_BY_DEFAULT,
						gas: DataDisplayOptions.SHOWN_BY_DEFAULT,
						nonce: DataDisplayOptions.NOT_IN_UI,
						to: DataDisplayOptions.SHOWN_BY_DEFAULT,
						value: DataDisplayOptions.SHOWN_BY_DEFAULT,
					},
					[SimpleBenchmarkTransactions.ETH_USDC_TRANSFER]: { /* Same.. */ },
					[ComplexBenchmarkTransactions.USDC_APPROVAL]: {
						from: DataDisplayOptions.SHOWN_BY_DEFAULT,
						chain: DataDisplayOptions.SHOWN_BY_DEFAULT,
						gas: DataDisplayOptions.SHOWN_BY_DEFAULT,
						nonce: DataDisplayOptions.NOT_IN_UI,
						transactionOutcome: TransactionOutcome.EXPLAINED,
					},
					/* ... Same for other complex benchmark transactions... */
					[SimulationBenchmarkTransactions.FAILED_TRANSACTION]: {
						from: DataDisplayOptions.SHOWN_BY_DEFAULT,
						chain: DataDisplayOptions.SHOWN_BY_DEFAULT,
						gas: DataDisplayOptions.SHOWN_BY_DEFAULT,
						nonce: DataDisplayOptions.NOT_IN_UI,
						transactionOutcome: TransactionOutcome.EXPLAINED,
						failure: 'DETECTED',
					},
					[SimulationBenchmarkTransactions.NONDETERMINISTIC_TRANSACTION]: {
						from: DataDisplayOptions.SHOWN_BY_DEFAULT,
						chain: DataDisplayOptions.SHOWN_BY_DEFAULT,
						gas: DataDisplayOptions.SHOWN_BY_DEFAULT,
						nonce: DataDisplayOptions.NOT_IN_UI,
						transactionOutcome: TransactionOutcome.EXPLAINED,
						failure: 'DETECTED_WITH_WARNING',
					},
				},
			},
```

This way I think this aligns well with the workflow needed to test wallets: we go through each benchmark transaction _once_, and we write down everything that the wallet shows for each benchmark transaction: both the plain fields, and the transaction simulation results.

maykelxyz:
A conflict I noticed while updating the PR per your feedback is the **feeling** of redundancy between `legibility` and `transactionDetailsDisplay`. If given the full rating, here is what it looks like:

```ts
legibility: {
	[CalldataDecoding.ETH_TRANSFER]: false,
	[CalldataDecoding.ERC_20_TRANSFER]: false,
	[CalldataDecoding.ERC_721_TRANSFER]: false,
	...,
},
messageSigningLegibility: {
	[MessageSigningDetails.EIP712_STRUCT]: DataDisplayOptions.SHOWN_BY_DEFAULT,
	...,
},
transactionDetailsDisplay: {
	[BasicBenchmarkTransactions.ETH_TRANSFER]: displaysFullTransactionDetails,
	[BasicBenchmarkTransactions.ERC_20_TRANSFER]: displaysFullTransactionDetails,
	[BasicBenchmarkTransactions.ERC_721_TRANSFER]: displaysFullTransactionDetails,
	[BasicBenchmarkTransactions.ZKSYNC_USDC_TRANSFER]: displaysFullTransactionDetails,

````

`legibility` and `transactionDetailsDisplay` feels both the same. I suggest we name `legibility` to `calldataDecoding`. That way, the features are more defined.

> We split CalldataDecoding into two enums: the ones about basic operations (ETH/token transfers) and the ones about more complex things (USDC approval, Aave deposit, Safe transactions). We name these BasicBenchmarkTransactions and ComplexBenchmarkTransactions. CalldataDecoding just becomes the union of the two enums.

While I agree we can split this into two enums, I still think there’s value in keeping transactionOutcome within BasicBenchmarkTransactions. Some examples are claiming airdrops that result in either native tokens (ETH) or ERC-20 tokens, or receiving NFTs through claiming or purchasing. These situations can also turn into scam transactions. For instance, a malicious site may ask you to sign what appears to be a valid transaction, but it actually sends all your ETH, ERC-20 tokens, or NFTs to an attacker. This makes the `transactionOutcome` field even more important, since the end result of the transaction is what ultimately matters. 

Feel free to let me know your thoughts

File: data/software-wallets/ambire.ts [hunk 1]
```diff
@@ -607,6 +607,7 @@ export const ambire: SoftwareWallet = {
 					to: DataDisplayOptions.SHOWN_BY_DEFAULT,
 					value: DataDisplayOptions.SHOWN_BY_DEFAULT,
 				},
+				transactionSimulation: null,
```
polymutex: Reminder to keep this fully rated.
(Leaving this comment open until done)

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -328,7 +341,7 @@ function generateHardwareHowToImprove(features: HardwareFeatureDetails): string
 
 	if (features.calldataDecoding.decodedLocation === DataDecoded.OFF_DEVICE) {
 		improvements.push(
-			'**Calldata Decoding:** Move decoding on-device so users don’t have to trust a potentially compromised companion app.',
+			'**Calldata Decoding:** Move decoding on-device so users don\u2019t have to trust a potentially compromised companion app.',
```
polymutex: Just use a regular `'`

src/schema/features/security/transaction-legibility.ts
```diff
@@ -182,6 +215,60 @@ export enum CalldataDecoding {
 	// In the future, add decoding of L1 -> L2 messages like `sendToL1`
 }
 
+/**
+ * CalldataDecoding is the union of basic and complex benchmark transactions.
+ * Used for hardware wallet calldata decoding evaluation.
+ */
+export type CalldataDecoding = BasicBenchmarkTransactions | ComplexBenchmarkTransactions
```
polymutex: We can just call this type `BenchmarkTransactions`.

src/schema/features/security/transaction-legibility.ts
```diff
@@ -182,6 +215,60 @@ export enum CalldataDecoding {
...
+
+/**
+ * Merged enum-like const for CalldataDecoding, allowing CalldataDecoding.ETH_TRANSFER etc.
+ */
+
+export const CalldataDecoding = {
+	...BasicBenchmarkTransactions,
+	...ComplexBenchmarkTransactions,
+} as const
```
polymutex: Let's not have type names and value names clash.
I'm not sure where this one is used, but similar things have used the `lowerCamelCase` version of the name, and using the `Enum` helper class:

```typescript
const basicBenchmarkTransactions = new Enum<BasicBenchmarkTransactions>({
  [BasicBenchmarkTransactions.ETH_TRANSFER]: true,
  // [...]
})

const complexBenchmarkTransactions = new Enum<ComplexBenchmarkTransactions>({
  [ComplexBenchmarkTransactions.USDC_APPROVAL]: true,
  // [...]
})

const benchmarkTransactions = mergeEnums(basicBenchmarkTransactions, complexBenchmarkTransactions)
```

Look at `sourceAvailableLicense` in `src/schema/features/transparency/license.ts` as an example.

polymutex:
> A conflict I noticed while updating the PR per your feedback is the **feeling** of redundancy between `legibility` and `transactionDetailsDisplay`.[...]
> 
> > We split CalldataDecoding into two enums: the ones about basic operations (ETH/token transfers) and the ones about more complex things (USDC approval, Aave deposit, Safe transactions). We name these BasicBenchmarkTransactions and ComplexBenchmarkTransactions. CalldataDecoding just becomes the union of the two enums.
> 
> While I agree we can split this into two enums, I still think there’s value in keeping transactionOutcome within BasicBenchmarkTransactions. Some examples are claiming airdrops that result in either native tokens (ETH) or ERC-20 tokens, or receiving NFTs through claiming or purchasing. These situations can also turn into scam transactions. For instance, a malicious site may ask you to sign what appears to be a valid transaction, but it actually sends all your ETH, ERC-20 tokens, or NFTs to an attacker. This makes the `transactionOutcome` field even more important, since the end result of the transaction is what ultimately matters.
> 
> Feel free to let me know your thoughts

As discussed briefly on a call, one idea is to remove `legibility` and instead to add a `transactionOutcome` field within `DisplayedBasicTransactionDetails` and `DisplayedComplexTransactionDetails`. I think that:

- Cuts some of the redundancy
- Aligns well with the idea of putting `failure` and `nondeterminism` together with the per-benchmark-transaction fields where those fields are relevant. Since the transaction outcome is relevant for all transactions, we can include it there.
- (Minor benefit) Reorients the feature type with the way they are tested in practice: test one benchmark transaction, write down what you see (from/to/nonce/etc and now also transaction outcome) under a single feature object, then move onto the next one, etc. rather than spreading this information into multiple parts of the structure.

Let me know your thoughts as well, don't want to churn too much on these types :)
Will also leave some PR comments directly...

maykelxyz:
Thanks for the comments and clarifying things up. So far no issues with anything aside from:

> As discussed briefly on a call, one idea is to remove legibility and instead to add a transactionOutcome field within DisplayedBasicTransactionDetails and DisplayedComplexTransactionDetails.

Imo there's still value in seeing decoded calldata, and that's what legibility is. My suggestion is to only have that property in `ComplexBenchmarkTransactions` (Aave supply, USDC Approval, Safewallet Aave Supply, etc.) and not in `BasicBenchmarkTransactions`.

maykelxyz:
```ts
transactionDetailsDisplay: {
	[BasicBenchmarkTransactions.ETH_TRANSFER]: {
		gas: DataDisplayOptions.SHOWN_BY_DEFAULT,
		nonce: DataDisplayOptions.NOT_IN_UI,
		from: DataDisplayOptions.SHOWN_BY_DEFAULT,
		to: DataDisplayOptions.SHOWN_BY_DEFAULT,
		chain: DataDisplayOptions.SHOWN_BY_DEFAULT,
		value: DataDisplayOptions.SHOWN_BY_DEFAULT,
		transactionOutcome: TransactionOutcome.EXPLAINED,
	},
	[BasicBenchmarkTransactions.ERC_20_TRANSFER]: {
		transactionOutcome: TransactionOutcome.EXPLAINED,
		..,
	},
	[BasicBenchmarkTransactions.ERC_721_TRANSFER]: {
		transactionOutcome: TransactionOutcome.EXPLAINED,
		...,
	},
	[BasicBenchmarkTransactions.ZKSYNC_USDC_TRANSFER]: {
		transactionOutcome: TransactionOutcome.EXPLAINED,
		...,
	},
	[ComplexBenchmarkTransactions.USDC_APPROVAL]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		decodedCalldata: DataDisplayOptions.SHOWN_BY_DEFAULT,
	},
	[ComplexBenchmarkTransactions.AAVE_SUPPLY]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		decodedCalldata: DataDisplayOptions.SHOWN_BY_DEFAULT,
	},
	[ComplexBenchmarkTransactions.SAFEWALLET_AAVE_SUPPLY_NESTED]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		decodedCalldata: DataDisplayOptions.SHOWN_BY_DEFAULT,
	},
	[ComplexBenchmarkTransactions.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		decodedCalldata: DataDisplayOptions.SHOWN_BY_DEFAULT,
	},
	[SimulationBenchmarkTransactions.FAILED_TRANSACTION]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		failure: 'DETECTED',
	},
	[SimulationBenchmarkTransactions.NONDETERMINISTIC_TRANSACTION]: {
		...,
		transactionOutcome: TransactionOutcome.EXPLAINED,
		nondeterminism: 'DETECTED_WITH_WARNING',
	},
},
```

This is how I imagined it would look like completed. This way it feels `complete`, legibility still plays a part **whether a wallet decodes the transactions or not**

Edit: updated legibility to `decodedCalldata` and change the type to enum `DataDisplayOptions`. Cleaner that way. 

Yeah I think this much more clean. If I'm a contributor I'd see this as: Under `transactionDetailsDisplay`, is the `decoded calldata` provided?

polymutex:
> This is how I imagined it would look like completed. This way it feels complete, legibility still plays a part whether a wallet decodes the transactions or not

Looks almost perfect, but... I think we should **not** have `transactionOutcome` in `SimulationBenchmarkTransactions.*`.

- `SimulationBenchmarkTransactions.FAILED_TRANSACTION` already has a `failure` field (`DETECTED` or not). A value of `DETECTED` means the outcome is already clear; having a separate `transactionOutcome` field would be redundant.
- `SimulationBenchmarkTransactions.NONDETERMINISTIC_TRANSACTION` has a `nondeterminism` field, one of the possible values being `DETECTED_WITH_WARNING`. The "warning" part _is_ the "transaction outcome being explained". So having a separate `transactionOutcome` field would be also redundant, and, in this case, also be ambiguous (what does it mean for a nondeterministic transaction to have its outcome "explained"? unclear; but the `nondeterminism` field values capture this ambiguity much better).

.cspell.json
```diff
@@ -81,6 +81,7 @@
 		"ERCs",
 		"eslintcache",
 		"eternalsafe",
+		"ETH",
```
polymutex, line 84: Can't believe we didn't have this one.

data/software-wallets/ambire.ts
```diff
@@ -264,7 +277,7 @@ export const ambire: SoftwareWallet = {
 						url: 'https://www.ambire.com/',
 					},
 					feesLargerThan1bps: comprehensiveFeesShownByDefault,
-					risksExplained: 'NOT_IN_UI',
+					risksExplained: 'NOT_DECODED',
```
polymutex: Is this replacement accidental? Looks like it hit other wallet files too.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -137,26 +140,38 @@ function analyzeHardwareFeatures({
...
 			{
-				key: CalldataDecoding.SAFEWALLET_AAVE_SUPPLY_NESTED,
+				key: ComplexBenchmarkTransactions.SAFEWALLET_AAVE_SUPPLY_NESTED,
 				label: 'nested Safe transactions',
 			},
 			{
-				key: CalldataDecoding.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND,
+				key: ComplexBenchmarkTransactions.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND,
 				label: 'complex nested multisend transactions',
 			},
```
polymutex: Maybe we should put this in a function close to the enum definitions. Similar to, say, the `licenseName` function in `src/schema/features/transparency/license.ts`. Using a `switch` in a function that must return a value enforces that if we add new benchmark transactions, the typechecker enforces that the string conversion function is updated too.

src/schema/features/security/transaction-legibility.ts
```diff
@@ -182,29 +230,109 @@ export enum CalldataDecoding {
...
-	CalldataDecoding,
-	Support<WithRef<CalldataDecodingSupport>>
->
+export type BenchmarkTransactions = BasicBenchmarkTransactions | ComplexBenchmarkTransactions
+
+/** Merged enum for all benchmark transactions. */
+export const benchmarkTransactions = mergeEnums(
+	basicBenchmarkTransactions,
+	complexBenchmarkTransactions,
+)
```
polymutex, line 249: Should this also include `SimulationBenchmarkTransactions`?

src/schema/features/security/transaction-legibility.ts
```diff
@@ -182,29 +230,109 @@ export enum CalldataDecoding {
...
+/**
+ * Details for a nondeterministic simulation benchmark transaction.
+ */
+export interface DisplayedNondeterministicTransactionDetails extends Omit<
+	DisplayedComplexTransactionDetails,
+	'transactionOutcome' | 'calldataDecoded'
+> {
+	/**
+	 * NOT_DETECTED - The wallet only shows one possible outcome regardless of whether the transaction relies on execution state.
+	 * DETECTED_WITHOUT_WARNING - The wallet detects multiple outcomes but does not warn the user.
```
polymutex: Looking at this again, I'm not sure how we would be able to tell the difference between `NOT_DETECTED` and `DETECTED_WITHOUT_WARNING`. What does the difference look like, UI-wise?

src/schema/features/security/transaction-legibility.ts
```diff
@@ -182,29 +230,109 @@ export enum CalldataDecoding {
...
+				[SimulationBenchmarkTransactions.FAILED_TRANSACTION]: DisplayedFailedTransactionDetails
+			} & {
+				[SimulationBenchmarkTransactions.NONDETERMINISTIC_TRANSACTION]: DisplayedNondeterministicTransactionDetails
+			})
+	| null
+
+/**
+ * Types of transactions that a wallet can decode the calldata of.
+ */
+export type CalldataDecodingTypes = Record<BenchmarkTransactions, DataDecoded | null> // Allow null for existing wallets that has not enough data
```
polymutex: s/has not enough/don't have enough

src/schema/features/security/transaction-legibility.ts
```diff
@@ -242,20 +370,22 @@ export interface HardwareMessageSigningLegibility {
  * Shorthand for a wallet that cannot do any calldata decoding.
  */
 export const noCalldataDecoding: CalldataDecodingTypes = {
-	[CalldataDecoding.ETH_USDC_TRANSFER]: notSupported,
-	[CalldataDecoding.ZKSYNC_USDC_TRANSFER]: notSupported,
-	[CalldataDecoding.USDC_APPROVAL]: notSupported,
-	[CalldataDecoding.AAVE_SUPPLY]: notSupported,
-	[CalldataDecoding.SAFEWALLET_AAVE_SUPPLY_NESTED]: notSupported,
-	[CalldataDecoding.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND]: notSupported,
+	[BasicBenchmarkTransactions.ETH_TRANSFER]: DataDecoded.NOT_DECODED,
```
polymutex, line 411: I just realized this one doesn't really make sense. ETH transfers don't have any calldata. But maybe it's OK to leave it in there anyway and ignore it, as otherwise it complicates the types further...

data/software-wallets/ambire.ts
maykelxyz: Yeah, it was. Already fixed.

src/schema/features/security/transaction-legibility.ts
maykelxyz: I’d say the difference is mostly in how the simulation behaves over time.

NOT_DETECTED – The wallet shows one possible outcome and keeps it static, even if the execution state changes. No re-simulation, no update.

DETECTED_WITHOUT_WARNING – The wallet re-simulates and updates the outcome if the state changes, but doesn’t explicitly warn the user. You’d only notice by seeing the refreshed simulation.

So basically: one acts like nothing changed, the other detects it but doesn’t call attention to it.

.cspell.json
maykelxyz: Yeah me too. Thought spell check was having a bug or something, lol

src/schema/features/security/transaction-legibility.ts
maykelxyz: Removed it. Unused anyways

src/schema/features/security/transaction-legibility.ts
polymutex: OK. In that case I think we could rename these values like this:

- `NOT_DETECTED` → `NO_TRANSACTION_OUTCOME_DISPLAYED`
- `DETECTED_WITHOUT_WARNING` → `DISPLAYS_ONE_POSSIBLE_OUTCOME_ONLY`
- `DETECTED_WITH_WARNING` → `WARNS_ABOUT_MULTIPLE_POSSIBLE_OUTCOMES`

Or something unambiguous like that.

maykelxyz: I think the suggested renames slightly change the original meaning.

Originally:

`NOT_DETECTED` → The wallet shows a single outcome and keeps it static. No re-simulation when state changes.

`DETECTED_WITHOUT_WARNING` → The wallet re-simulates and updates the outcome if state changes, but doesn’t explicitly warn about multiple possible outcomes.

So the key distinction isn’t whether an outcome is displayed, both display one, but whether the wallet re-detects state-dependent changes.

I would suggest something like:

```ts
/**
 * How the wallet handles state-dependent (non-deterministic) transactions.
 *
 * - STATIC_SINGLE_OUTCOME: Shows one outcome and keeps it static. No re-simulation if state changes.
 * - RESIMULATES_NO_WARNING: Re-simulates and updates the outcome if state changes, but doesn’t explicitly warn the user.
 * - RESIMULATES_WITH_WARNING: Re-simulates and explicitly warns that multiple outcomes are possible.
 */
nondeterminism: 
  | 'STATIC_SINGLE_OUTCOME'
  | 'RESIMULATES_NO_WARNING'
  | 'RESIMULATES_WITH_WARNING'
```

polymutex: OK, looks good, but since we don't have a `transactionOutcome` field here, then the case of "the wallet doesn't show any outcome at all" (e.g. if it doesn't support transaction simulation) can't be encoded. So sounds like we need a fourth value like `NO_OUTCOME_SHOWN` or something.

data/hardware-wallets/ledger.ts
```diff
@@ -186,13 +186,13 @@ export const ledgerWallet: HardwareWallet = {
 						url: 'https://youtu.be/9YmPWxAvKYY?t=1722',
 					},
 				],
+				calldataDecoded: noCalldataDecoding,
```
polymutex, line 189: FWIW I think it does detect ERC-20 approvals.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -137,28 +141,16 @@ function analyzeHardwareFeatures({
...
-			{
-				key: CalldataDecoding.SAFEWALLET_AAVE_SUPPLY_NESTED,
-				label: 'nested Safe transactions',
-			},
-			{
-				key: CalldataDecoding.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND,
-				label: 'complex nested multisend transactions',
-			},
+	if (calldataDecoded !== null) {
+		const decodingChecks: Array<BasicBenchmarkTransactions | ComplexBenchmarkTransactions> = [
```
polymutex: Maybe this is where a merged enum (from `mergeEnum`) would shine? This would just be `benchmarkTransactions.items`.

src/schema/attributes/security/transaction-legibility.ts
polymutex: Not resolved

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -481,94 +469,249 @@ interface SoftwareFeatureDetails {
...
+				)
+			} else if (tx.nondeterminism === 'DETECTED_WITHOUT_WARNING') {
+				details.transactions.partial.push(
+					'Nondeterministic transaction simulation (detected but no warning shown)',
+				)
+			} else {
+				details.transactions.passing.push('Nondeterministic transaction simulation')
+			}
+		}
 	}
```
polymutex, line 734: I like this structure in the way each benchmark transaction is checked. Feels better / more coherent.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1014,19 +1171,16 @@ export const transactionLegibility: Attribute<TransactionLegibilityValue> = {
 
 		**Calldata Decoding/Display:**
 		The wallet's ability to decode and display calldata for various transaction types, including:
-		- Simple transfers
+		- Simple transfers (ETH and token)
```
polymutex: ETH transfers, ERC-20 transfers, and ERC-721 transfers

(Should we add ERC-1155 to this list btw?)

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1014,19 +1171,16 @@ export const transactionLegibility: Attribute<TransactionLegibilityValue> = {
...
-		- Gas limit and/or gas price
-		- Transaction nonce
-		- Sender address (from)
-		- Recipient address (to)
-		- Chain/network identifier
-		- Transaction value/amount
+		For software wallets, transaction details are evaluated per benchmark transaction:
+		- Basic benchmarks (ETH transfer, token transfers): gas, nonce, from, to, chain, value
+		- Complex benchmarks (approvals, DeFi, nested transactions): gas, nonce, from, chain, and whether the transaction outcome is explained
+		- Simulation benchmarks (failed transactions, nondeterministic outcomes): whether failures and outcome variance are detected
```
polymutex: "Simulation benchmarks" is Walletbeat jargon. In fact I think we should get rid of the basic/complex/simulation distinction in this explanation. It's a Walletbeat-internal implementation detail to distinguish the different transaction types and what we care about them, but it's not important to use the same wording here for user-facing text.

Alternative:

```
		For software wallets, transaction details are evaluated according to the set of information they display:
		- For transfers (ETH transfers, ERC-20 transfers, ERC-721 transfers, ERC-1155 transfers): gas, nonce, sender, recipient, chain, amount. Some details may be collapsed by default (e.g. `nonce`), but all of them must be accessible through the UI.
		- Non-transfer transactions (approvals, DeFi contract interactions, Safe multisig nested transactions): gas, nonce, from, chain. In addition, the wallet must explain the transaction outcome visually or in plain language; for example, for ERC-20 approvals, it should explain what token is being approved, for which contract, and for how many tokens.
		- If presented with a transaction that will revert if included onchain, the wallet must detect that the transaction would fail and warn the user beforehand.
		- If presented with a transaction for which the behavior would be nondeterministic based on predictably-changing parameters (e.g. block number), the wallet must detect that the transaction has nondeterministic outcome and warn the user beforehand.
```

src/schema/features/security/transaction-legibility.ts
```diff
@@ -48,25 +48,79 @@ export const displaysFullTransactionDetails: DisplayedTransactionDetails = {
...
+
+	/**
+	 * Sending of ERC-20 tokens to another address.
 	 */
-	ETH_USDC_TRANSFER = 'ETH_USDC_TRANSFER',
+	ERC_20_TRANSFER = 'ERC_20_TRANSFER',
+	/**
+	 * Sending of ERC-721 tokens (NFTs) to another address.
+	 */
+	ERC_721_TRANSFER = 'ER_721_TRANSFER',
```
polymutex, line 106: `ERC-1155_TRANSFER`?

src/schema/features/security/transaction-legibility.ts
polymutex: Not resolved

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: Resolved

data/hardware-wallets/ledger.ts
maykelxyz: Hmm, I can follow up on that. This is from the original file.

src/schema/features/security/transaction-legibility.ts
maykelxyz: Resolved

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: OK. Added it

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: You're right. Brought it back.

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: Do you mean you suggest to refactor this..? Or you like it this way?

polymutex: I like it, just a compliment.

src/schema/features/security/transaction-legibility.ts
```diff
@@ -251,6 +262,8 @@ export function benchmarkTransactionLabel(
 			return 'basic token transfers (ERC-20)'
 		case BasicBenchmarkTransactions.ERC_721_TRANSFER:
 			return 'basic NFT transfers (ERC-721)'
+		case BasicBenchmarkTransactions.ERC_1155_TRANSFER:
+			return 'basic semi fungible token transfers (ERC-1155)'
```
polymutex: "semi fungible" is only one of the ways to use ERC-1155. I think the right term is "multi-token transfers", as per https://ethereum.org/developers/docs/standards/tokens/erc-1155

polymutex (approved):

src/schema/attributes/security/transaction-legibility.ts
maykelxyz: Ah. Appreciate it, hehe

Issue #504, opened 2026-02-03
Stage system: move not-stage-0 wallets to separate tab


polymutex:
Wallets that don't meet the stage 0 requirements (source availability) cannot be verified for many of their attributes. Therefore, they should not be ranked on the same level as the wallets that can be.

L2BEAT had a similar problem and solved it by moving these "not even L2s"-type projects to a separate non-default tab:

<img width="3654" height="1395" alt="Image" src="https://github.com/user-attachments/assets/e85cb974-a93e-4e11-be3c-d0b7becec243" />

We can do the same, i.e. any wallet that isn't at least stage 0 would not be visible on the front page.

Marking this issue as `pre-launch` blocker because while this wouldn't technically be a methodology change, it may _feel_ like one to the wallets that move around to that tab.

maykelxyz:
+1 on this. "Not reviewed" tab seems good to implement too, especially for newer wallets or wallets that are yet to be reviewed.

Aside from avoiding ranking confusion, it also helps from a contributor POV. If a potential contributor or wallet dev sees a wallet they’re interested in sitting under “Not Reviewed,” the visual psychology of “this is empty / needs work” makes it more likely they’ll want to review it and help move it forward.

polymutex:
Yeah, that's an interesting idea too. "Not reviewed" seems like it should be based on the percentage of attributes that are `UNRATED`. We could have a threshold like ">90% attributes non-UNRATED" to get moved to the main page.

PR #517, opened 2026-02-06
Stage system: move not staged wallets to separate tab

johngrantuk:
Here's my attempt at #504.
I felt like the table layout had to be changed quite a bit to make the tabs work so its a pretty opinionated approach in terms of styling and layout. I mostly took inspiration directly from L2BEAT. 
I'm not very good at FE so I used Cursor heavily for this and mostly trusted its approach/implementation.
No problems if this is not accepted or requires a lot of changes - was fun to explore more of the codebase.

(I think a nice follow up improvement would be to add a filter dropdown similar to the L2BEAT page that would replace the current filters)

### DIFF
--- a/src/components/Table.svelte
+++ b/src/components/Table.svelte
@@ -797,6 +797,15 @@
 							}
 						}
 
+						&[data-column-align='start'] .sort-label,
+						&:not([data-column-align]) .sort-label {
+							justify-content: flex-start;
+						}
+
+						&[data-column-align='end'] .sort-label {
+							justify-content: flex-end;
+						}
+
 						&:has(.sort-button:focus) {
 							outline: var(--table-outerBorderWidth) solid var(--accent);
 							outline-offset: calc(-1 * var(--table-outerBorderWidth));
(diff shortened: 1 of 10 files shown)


maykelxyz:
Hey there @johngrantuk thanks again for contributing, I've checked it out and looks great!

image

The new styling - it looks great too, though I don't really have a strong opinion on this. I did notice that the width between categories (Security, Privacy, Self-sovereignty,..) are not equally distributed. So spacing seems a bit off.

Re: the follow-up PRs for changing the filter to a dropdown, I’d suggest we pause this for now since a designer is currently working on the website refresh. The scope might overlap.

cc: @polymutex

polymutex:
Agree with @maykelxyz, the tabs look great. If you'd like to contribute to the website design in general, as per @maykelxyz we have someone taking a pass at that shortly (there's a lot to fix!), so we should coordinate on this. If you'd like to do so, please reach out to me on Farcaster (`@polymutex.eth`) or by email (`my github username at protonmail dot com`).

polymutex:
We have converged on a redesigned look-and-feel ([Figma link](https://www.figma.com/design/fRFH2XbtyWwaBw1BRP96eh/Walletbeat?node-id=0-1&p=f)). This will likely make this PR obsolete, so I am closing it.

Issue #493, opened 2026-01-29
Docs: Improve guides and documentations for Walletbeat wallet contributing


maykelxyz:
One major improvement on how Walletbeat can make it easier for contributors adding wallet data is to be more **explicit**. I added both OKX and Bitget wallet and gathered what can be improved. I find Natspec comments to be more helpful than a single MD file which contributors will still look what they'll be looking for. 

Example:

<img width="576" height="122" alt="Image" src="https://github.com/user-attachments/assets/d19f682a-5cd4-43af-b7eb-baf48b6b7d7d" />

Adding natspec:
```ts
/** Address lookup through ERC-7831. Example: `vitalik.eth:optimism:1`
 * Sending ethereum to `vitalik.eth:optimism:1 should be possible`
*/
erc7831: ARS
```

Would now look like:

<img width="932" height="126" alt="Image" src="https://github.com/user-attachments/assets/45beeaa6-f806-4a81-b3de-3f70094059ff" />

The process would go from: Thinking "Hmm, how should I test this" -> "Yeah okay let me test this". It saves the pain of contributors thinking "How should I be testing this" to just testing.

maykelxyz:
More comments:

## Wallet Feature Data Formatting #224 

As I was looking into what could be improved: noticed that features are grouped not according to what I'm expecting: attribute groups.

<img width="1119" height="202" alt="Image" src="https://github.com/user-attachments/assets/55e9cf57-29a7-46b6-bfa7-6a7731de9b19" />

If I was a first time contributor, I'd probably look for ways to "fill in the blanks". Intuitively, I'd go by attribute groups as the homepage shows attribute groupings. Though you mentioned in #224 that:

> The only thing I'd caution against is to make sure that the new structure doesn't tie too strongly to the attribute group structure

I agree.

### Clarify how to accurately test features
- EIP 7702
  - Theres' two EIP-7702 testing. Clarifying the difference between the two and how to test them
```ts
eip7702: supported({
	ref: refTodo,
	contract: 'UNKNOWN',
}),
ecosystem: {
	delegation: null,
},
```
- Under privacy, privacyPolicy is already discussed on how to test it. Transaction Privacy and Data Collection is straightforward. We can clarify more on how to accurately test `appIsolation`. Or include in our testing suite.
- Under security, accountRecovery, keysHandling, lighTclient, passkeyVerification, scamAlerts
  - We can add to our testing page on how to test this or improve docs

maykelxyz:
[...]

PR #518, opened 2026-02-06
Ambire update: additional references and tx legibility 

JIOjosBG:
Adds:
- info about scam prevention, but does not change the resulting score


data/software-wallets/ambire.ts
```diff
@@ -533,13 +568,26 @@ export const ambire: SoftwareWallet = {
...
+						'Ambire displays the calldata in a dropdown under the transactioin visualisation as a selectable text that can be copied to the clicpboard.',
+					label: 'Implementation',
+					url: 'https://github.com/AmbireTech/extension/blob/main/src/web/modules/sign-account-op/components/TransactionSummary/TransactionSummary.tsx',
+					lastRetrieved: '2026-02-06',
+				},
 				calldataDisplay: {
-					copyHexToClipboard: false,
-					formatted: false,
+					copyHexToClipboard: true,
+					formatted: true,
```
maykelxyz: Image

Checked out Ambire, the button for copying hex to clipboard still doesn't exist. And also for `formatted`, one requirement is to see the calldata being parsed with the function name and its parameters. Photo attached above is a call for `supply(address,uint256,address,uint16)`

Image

data/software-wallets/ambire.ts
```diff
@@ -533,13 +568,26 @@ export const ambire: SoftwareWallet = {
...
+					leaksUserIp: true,
+				}),
 			},
 			transactionLegibility: {
-				ref: refTodo,
+				ref: {
+					explanation:
+						'Ambire displays the calldata in a dropdown under the transactioin visualisation as a selectable text that can be copied to the clicpboard.',
+					label: 'Implementation',
+					url: 'https://github.com/AmbireTech/extension/blob/main/src/web/modules/sign-account-op/components/TransactionSummary/TransactionSummary.tsx',
```
polymutex: If linking to specific source files, please link them with a commit hash, so that the link is effectively timestamped in time and remains valid even if the file is later moved/renamed.

In this case that would be `https://github.com/AmbireTech/extension/blob/74d04288e195c171f009f3dd55fdebf5c17b7ccd/src/web/modules/sign-account-op/components/TransactionSummary/TransactionSummary.tsx`

data/software-wallets/ambire.ts
```diff
@@ -533,13 +568,26 @@ export const ambire: SoftwareWallet = {
...
+					userWhitelist: false, // address book is no sufficient in functionality for this flag
+					leaksUserAddress: false,
+					leaksUserIp: true,
+				}),
 			},
 			transactionLegibility: {
-				ref: refTodo,
+				ref: {
+					explanation:
+						'Ambire displays the calldata in a dropdown under the transactioin visualisation as a selectable text that can be copied to the clicpboard.',
```
polymutex: Typos:
- `transactioin` → `transaction`
- `visualisation` → `visualization`
- `clicpboard` → `clipboard`

data/software-wallets/ambire.ts
```diff
@@ -151,6 +155,29 @@ const dataLeakReferences: Record<string, References> = {
...
+		urls: [
+			{
+				label: 'Implementation',
+				url: 'https://github.com/AmbireTech/ambire-common/blob/main/src/controllers/phishing/phishing.ts',
+			},
+		],
+	},
+	{
+		explanation:
+			'When attempting to send tokens via the built-in interface, we display warning based on the time since last transfer.',
```
polymutex: "we" makes sense when written from the wallet team's perspective (yours), but when displayed on the site, it will look like it is Walletbeat saying it, so "we" will no longer make sense. You can just use "Ambire displays a warning based [...]"

data/software-wallets/ambire.ts
```diff
@@ -151,6 +155,29 @@ const dataLeakReferences: Record<string, References> = {
 	],
 }
 
+const scamAlertsAndSendTxWarningRefs: WithRef<{}>['ref'] = [
+	{
+		explanation:
+			'When interacting with new address, the extension fetches info about that address from an external Ambire API.',
```
polymutex: `When interacting with new address` → `When interacting with a new address`

polymutex (changes requested):

data/software-wallets/ambire.ts
polymutex: Filed https://github.com/walletbeat/walletbeat/issues/519 to have automated tests to catch this

data/software-wallets/ambire.ts
JIOjosBG: 1/
calldata copy:
tbh it wasn't clear from the comment on the property that a button is expected
```ts
	/* Can the user copy the raw hex code to the clipboard? */
	copyHexToClipboard: boolean
```
I can change it to

`Is there a button for copying the raw hex call data to the clipboard`

Wdyt?

Also i will set the property back to false for Ambire

JIOjosBG: 2/
For account abstraction wallets that support batching, do you expect the copy calldata button to be for each call or do you expect a single button that copies the hex that will be broadcasted to the wallet - wrapped in the multicall function?

JIOjosBG: 3/
formatted:
again I would argue that Ambire displays the data in a nicely parsed manner, but this is not you are looking for as a feature.

```ts
	/* Can display the calldata in some formatted output (e.g. JSON) */
	formatted: boolean
```

I suggest changing this to

`Can display **all arguments from** the calldata in some formatted output (e.g. JSON)`

JIOjosBG: 4/ 
We already do something similar for typed messages, is such visualization sufficient for the calldata arguments?

(the is a 712 message that the signer (in ambire) signed for a safe account (in the safe app))
image

data/software-wallets/ambire.ts
JIOjosBG: Sorry for that. I deleted the ref, because both changed properties should be false, which will make the ref kinda misleading

data/software-wallets/ambire.ts
JIOjosBG: I removed the ref altogether

JIOjosBG:
After @maykelxyz's [comment](https://github.com/walletbeat/walletbeat/pull/518#discussion_r2775286683) I removed the changes for transaction legibility

data/software-wallets/ambire.ts
maykelxyz: Thanks @JIOjosBG, we'll definitely take note of this. We're already working on improving guides and documentation #493, so this is definitely noted. For your questions:

> do you expect the copy calldata button to be for each call or do you expect a single button that copies the hex that will be broadcasted to the wallet - wrapped in the multicall function?

For batched transactions, a single button that copies the hex warpped in the multicall function is expected. After all, the purpose of a dedicated button for copying the calldata is to reduce user error.

> We already do something similar for typed messages, is such visualization sufficient for the calldata arguments?

In this example, data contains the calldata, which can be parsed into something like `depositETH(address,address,uint16)`. This helps users understand the actual action being performed, rather than only seeing execTransaction.

Something like:
```
Function: execTransaction(address,uint256,bytes,uint8,uint256,uint256,uint256,address,address,bytes)
Parameters:
  param0: 0x78e30497a3c7527d953c6B1E3541b021A98Ac43c
  param1: 0
  param2:
    Function: supply(address,uint256,address,uint16)
    Selector: 0x617ba037
    Parameters:
      param0: 0x5A7d6b2F92C77FAD6CCaBd7EE0624E64907Eaf3E
      param1: 50000000000000000000
      param2: 0x9467919138E36f0252886519f34a0f8016dDb3a3
      param3: 0
    Raw Data: 0x617ba0370000000000000000000000005a7d6b2f92c77fad6ccabd7ee0624e64907eaf3e000000000000000000000000000000000000000000000002b5e3af16b18800000000000000000000000000009467919138e36f0252886519f34a0f8016ddb3a30000000000000000000000000000000000000000000000000000000000000000
  param3: 0
  param4: 0
  param5: 0
  param6: 0
  param7: 0x0000000000000000000000000000000000000000
  param8: 0x0000000000000000000000000000000000000000
  param9: 0x000000000000000000000000f8cade19b26a2b970f2def5ea9eccf1bda3d1186000000000000000000000000000000000000000000000000000000000000000001
```

cc: @polymutex

polymutex: > tbh it wasn't clear from the comment on the property that a button is expected

You're right, will fix :saluting_face:

> 2/ For account abstraction wallets that support batching, do you expect the copy calldata button to be for each call or do you expect a single button that copies the hex that will be broadcasted to the wallet - wrapped in the multicall function?

The intent is to make it easy for the user to extract what the wallet is about to approve and put it into another tool for analysis. So from that perspective, only a button to copy the whole calldata (for the whole multicall) makes sense, since having multiple buttons would increase the potential for user errors (forgetting to copy one of the N transactions of the batch).

polymutex (approved):

data/software-wallets/ambire.ts
polymutex: (btw the changes for Safe support coming to Ambire in https://x.com/borislavItskovv/status/2020770734168273371 look sick)

JIOjosBG: Copy button:
> For batched transactions, a single button that copies the hex warpped in the multicall function is expected. After all, the purpose of a dedicated button for copying the calldata is to reduce user error.

thanks for the explanation, makes a lot of sense, we will probably not prioritize it for now.

Formatted calldata:
Again, is this on a per-call basis or needed to be on whole batch? if it is a per-call requirement we will probably prioritize it

cc: @maykelxyz

JIOjosBG: > Formatted calldata:
Again, is this on a per-call basis or needed to be on whole batch? if it is a per-call requirement we will probably prioritize it

a friendly ping on this question, imo makes more sense to be on a per-call basis, but asking just in case

@polymutex @maykelxyz

maykelxyz: Hey there sorry if it wasn't clear but I mentioned last time:

> For batched transactions, a single button that copies the hex warpped in the multicall function is expected. After all, the purpose of a dedicated button for copying the calldata is to reduce user error.

But yes, for short, the wrapped multicall function is expected

JIOjosBG: my question is about the formatted calldata, that can be displayed as json, not the copy button for the raw hex

maykelxyz: Oh, my bad. In that case you're right. Makes more sense to be on a per-call basis :)

polymutex: @maykelxyz We should make both of these questions (the hex copy button and the formatted calldata display) clear in the feature type comments. You're editing these files in #512 and I don't want to add merge conflicts, so maybe fold those in that PR

maykelxyz: Will do, thanks @polymutex

PR #522, opened 2026-02-09
Feat: Finalize tester contract

### LINKED ISSUE #478: Walletbeat Contract for testing
As discussed, deploying a dedicated test contract is the final step required to include this in the evaluation page. This contract is designed to exercise and validate wallet support for the following capabilities:

- Transaction batching
- Transaction simulation
- Transaction legibility

Based on prior discussions in #190 and follow-ups in DMs with @polymutex , two key requirements guided the design of this test:
- State-dependent simulation correctness
The test must surface whether a wallet performs actual execution-accurate simulation, rather than heuristic or selector-based assumptions. This is achieved by making the transaction outcome depend on current on-chain state (specifically block.number % 100), ensuring that a correct result can only be derived from a faithful simulation of the call.

- Non-transferable (soulbound) test assets
Any ERC20 or ERC721 tokens produced by the test must be non-transferable to avoid misinterpretation that Walletbeat is issuing or distributing real assets. Enforcing soulbound behavior ensures the tokens exist purely as test assets and cannot be moved or reused outside the context of the test.

### Few Notes:
- There already is an existing PR #442 to kickstart the development
- Currently uses Foundry for development and scripting

### Walletbeat Test Contract

```solidity
    /**
     * @notice Mints ERC20 and ERC721 tokens to the caller
     * @dev Calls mint() on both the ERC20 and ERC721 token contracts
     */
    function simulateFunction() external {
        (bool success,) = i_erc20Token.call(abi.encodeWithSignature("mint()"));
        if (!success) {
            revert WalletbeatTestContract__ERC20CallFailed();
        }
        (success,) = i_erc721Token.call(abi.encodeWithSignature("mint()"));
        if (!success) {
            revert WalletbeatTestContract__ERC721CallFailed();
        }
    }

    /**
     * @notice Identical to simulateFunction but uses the common transfer(address,uint256) selector
     * @dev This function has the same implementation as simulateFunction but uses a well-known
     * function signature to test how wallets handle and simulate transactions when presented
     * with a familiar selector like `transfer`. This helps evaluate whether wallets correctly
     * simulate the actual behavior or make assumptions based on the function name/selector.
     */
    function transfer(address, uint256) external {
        (bool success,) = i_erc20Token.call(abi.encodeWithSignature("mint()"));
        if (!success) {
            revert WalletbeatTestContract__ERC20CallFailed();
        }
        (success,) = i_erc721Token.call(abi.encodeWithSignature("mint()"));
        if (!success) {
            revert WalletbeatTestContract__ERC721CallFailed();
        }
    }
```

### WalletbeatTestErc20

```solidity
contract WalletbeatTestErc20 is ERC20, Ownable {
    error WalletbeatTestErc20__Soulbound();
    constructor(string memory name, string memory symbol) ERC20(name, symbol) {}

    function mint() external onlyOwner {
        uint256 tokensToMint = block.number % 100;
        if (tokensToMint == 0) {
            tokensToMint = 1;
        }
        super._mint(msg.sender, tokensToMint);
    }

    function _beforeTokenTransfer(address from, address to, uint256 amount) internal virtual override {
        super._beforeTokenTransfer(from, to, amount);

        if (from == address(0) || to == address(0)) {
            return;
        }

        revert WalletbeatTestErc20__Soulbound();
    }
}
```

### WalletbeatTestErc721

```solidity
contract WalletbeatTestErc721 is ERC721, Ownable {
    error WalletbeatTestErc20__Soulbound();
    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}

    function mint() external onlyOwner {
        uint256 tokensToMint = block.number % 100;
        if (tokensToMint == 0) {
            tokensToMint = 1;
        }
        super._mint(msg.sender, tokensToMint);
    }

    function _beforeTokenTransfer(address from, address to, uint256 firstTokenId, uint256 batchSize)
        internal
        virtual
        override
    {
        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);

        if (from == address(0) || to == address(0)) {
            return;
        }

        revert WalletbeatTestErc20__Soulbound();
    }
}
```

maykelxyz:
# PR: Feat: Contracts Finalizing

## Summary

A suite of Solidity smart contracts designed to evaluate how Ethereum wallets simulate, display, and warn users about transactions. The contracts intentionally implement deceptive and adversarial patterns to test wallet security UX.

### WalletbeatTestContract

The main entry point for wallet testing. All functions are permissionless.

| Function | Description |
|----------|-------------|
| `simulateFunctionV1()` | Mints ERC20 and ERC721 tokens to `msg.sender`. Tests basic multi-contract transaction simulation. |
| `simulateFunctionV2()` | Calls `claim()` on the ERC20 token, which mints on even blocks and burns the caller's entire balance on odd blocks. Tests wallet detection of unpredictable behavior. |
| `transfer(address to, uint256 amount)` | Intentionally misleading - uses the well-known `transfer` selector but actually mints ERC20 and ERC721 tokens to `msg.sender` (not `to`). Emits a fake `Transfer` event to further mislead simulations. |
| `claim()` | Simulates a malicious airdrop claim. Burns all of the caller's ERC20 tokens while emitting a fake `Transfer` event that looks like a mint of `1e18` tokens. |
| `alwaysFails()` | Always reverts with `WalletbeatTestContract__AlwaysFails`. Tests how wallets handle transactions guaranteed to fail. |

### WalletbeatTestErc20

| Function | Description |
|----------|-------------|
| `mint(address user)` | Mints `1 + (block.number % 100)` tokens to the specified address. The variable amount introduces unpredictability for transaction simulations. |
| `claim(address user)` | Mints tokens if `block.number` is even, burns the user's entire balance if odd. Simulates malicious drain behavior. |
| `burn(address user)` | Burns all tokens held by the specified address. |

Tokens are soulbound via `_beforeTokenTransfer` - any transfer between non-zero addresses reverts with `WalletbeatTestErc20__Soulbound`.

### WalletbeatTestErc721

| Function | Description |
|----------|-------------|
| `mint(address receiver)` | Mints `1 + (block.number % 4)` NFTs to the specified address using an incrementing `s_tokenId` counter. |

Tokens are soulbound via `_beforeTokenTransfer` - any transfer between non-zero addresses reverts with `WalletbeatTestErc721__Soulbound`.

## Other Changes
- Added Deploy Contract Scripts
  - Deploys all three contracts. The ERC20 and ERC721 are deployed first, then their addresses are passed to the WalletbeatTestContract contract constructor.
- Created testing suite
  - 35 unit tests and 4 stateless fuzz tests (256 runs each). All 40 tests passing.

## To clarify:

- Are there any other suggestions for the token name/symbol aside from "Walletbeat Testing" and "WBT"? We can just use this as this is just a contract intended for testing.
- `mint()` and `burn()` can be called by anyone on behalf of any address. This was based on a comment by @polymutex — is this the intended behavior, or did I misinterpret?

## To follow up
- imgUri for the ERC721 token

Closes #478


resources/contracts/src/WalletbeatTestErc20.sol
polymutex: How about making it mintable and burnable by anyone? I think that would help ensure it has no market value too (beyond the soulboundness), and help with ensuring users can have a non-zero starting balance when simulating the `claim` function.

polymutex (commented):
Can you rebase this on top of `beta`?

maykelxyz:
Updated the PR description

polymutex:
> Are there any other suggestions for the token name/symbol aside from "Walletbeat Testing" and "WBT"? We can just use this as this is just a contract intended for testing.

I'd suggestg `WBTEST` to make it even more obvious that it's a test token.

> `mint()` and `burn()` can be called by anyone on behalf of any address. This was based on a comment by @polymutex — is this the intended behavior, or did I misinterpret?

I don't have a strong opinion either way, but what's important to me is:

- The token has no financial value. The ability for anyone to burn anyone else's tokens ensures that this is always the case.
- The token can be minted at will, so that users (including those who aren't Walletbeat contributors) testing their wallets can start from a non-zero balance when they go on the wallet test playground. Being able to call `mint()` and grant yourself a huge amount of token is sufficient for this, so technically being able to mint tokens into other addresses' wallets isn't necessary, but is logically equivalent to self-mint followed by a transfer, so it's a distinction without a difference.

.cspell.json
```diff
@@ -331,7 +331,8 @@
 		"public/**/*.svg",
 		"resources/files/**",
 		"src/generated/**",
-		"src/legacy/**"
+		"src/legacy/**",
+		"resources/contracts/**"
```
polymutex: Why not spellcheck it?

polymutex (approved):

.cspell.json
maykelxyz: The majority of spell-check errors come from `resources/contracts/lib/**`, `resources/contracts/out/**`, and `resources/contracts/cache/**`, these are Foundry artifacts and other libraries (forge-std, OpenZeppelin), not code we author or maintain. 

If we ignored only those paths, the only remaining spell-check error would be `remappings`, which is actually in our own code. So it made more sense for me to just ignore the whole `resources/contracts/` path.

maykelxyz:
Update: Added the `tokenUri` and the image of the NFT that looks like this:
```json
{
  "name": "Walletbeat Testing ERC721",
  "description": "A test ERC721 token used solely for testing.",
  "attributes": [
    {
      "trait_type": "purpose", 
      "value": "testing"
    }
  ]
}
```


Walletbeat

polymutex (approved):

PR #527, opened 2026-02-10
Change address resolution

### LINKED ISSUE #507: Raise bar higher for chain specific addressing
# Raise bar higher for Chain Specific Addressing
As of now, ERC-7828 and ERC-7831 are not required for a wallet to pass address resolution. The original reasoning was that these standards are not yet widely implemented across wallets. While the statement is still true, we may want to evolve this to reflect Walletbeat's principle "Raising the bar higher". 

Proposed changes:
- Require ERC-7828 + ERC-7831 to pass address resolution.
- Require ENS solution to qualify as a stage 1 wallet, and treat ERC-7828 + ERC-7831 as a higher stage or advanced requirements.

johngrantuk:
Closes #507  

Address resolution changes:
- ENS is PARTIAL instead of PASS
- ERC-7828 and ERC-7831 were already PASS so keep the same
- Removed the “plain ENS onchain” example from ratingScale.pass
- Inserted that same example as the first item in ratingScale.partial

Split Address Resolution into Two Stage Requirements:
- Stage 1 (Basic): Requires basic human-readable address support (e.g., ENS)
  - Updated to allow PARTIAL ratings (allowPartial: true)
  - Wallets with basic ENS support can pass Stage 1
- Stage 2 (Advanced): Requires chain-specific human-readable addresses (ERC-7828, ERC-7831)
  - New criterion added: chain_specific_address_resolution
  - Requires PASS rating (allowPartial: false)
  - Only wallets supporting chain-specific formats like user@l2chain.eth or user.eth:l2chain pass Stage 2

New Wallet-Specific Stage Lookup Function:
Ensures the UI reflects the wallet's actual performance rather than just where the attribute is defined.
- Added getAttributeStagesForWallet() in stage-attributes.ts:
- Filters stages to those where the wallet actually passes the criterion
- Returns only the highest stage passed per ladder (e.g., if a wallet passes Stage 2, it doesn't also show Stage 1)
- Replaces the generic getAttributeStages() which showed all stages where an attribute is used, regardless of wallet performance

### DIFF
--- a/src/schema/attributes/ecosystem/address-resolution.ts
+++ b/src/schema/attributes/ecosystem/address-resolution.ts
@@ -155,15 +155,17 @@ function evaluateAddressResolution(
 	if (addressResolution.nonChainSpecificEnsResolution.medium === 'CHAIN_CLIENT') {
 		return {
 			value: {
-				id: 'support_basic_resolution_onchain',
-				rating: Rating.PASS,
-				displayName: 'Supports ENS addresses',
+				id: 'support_plain_ens_onchain',
+				rating: Rating.PARTIAL,
+				displayName: 'Supports plain ENS addresses only',
 				addressResolution,
-				shortExplanation: sentence('{{WALLET_NAME}} supports sending to ENS addresses.'),
+				shortExplanation: sentence(
+					'{{WALLET_NAME}} supports sending to plain ENS addresses but not chain-specific human-readable addresses.',
+				),
 				__brand: brand,
 			},
 			details: markdown(`
-				{{WALLET_NAME}} supports sending funds to human-readable ENS addresses such as \`username.eth\`.
+				{{WALLET_NAME}} supports sending funds to human-readable ENS addresses such as \`username.eth\`, but does not support chain-specific address formats that specify the destination chain.
 
 				It does so using onchain data sources using the same code as when interacting with the chain in general, inheriting its privacy and verifiability properties.
 			`),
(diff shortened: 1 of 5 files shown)


src/schema/attributes/ecosystem/address-resolution.ts
```diff
@@ -155,15 +155,17 @@ function evaluateAddressResolution(
 	if (addressResolution.nonChainSpecificEnsResolution.medium === 'CHAIN_CLIENT') {
 		return {
 			value: {
-				id: 'support_basic_resolution_onchain',
-				rating: Rating.PASS,
-				displayName: 'Supports ENS addresses',
+				id: 'support_plain_ens_onchain',
+				rating: Rating.PARTIAL,
+				displayName: 'Supports plain ENS addresses only',
```
polymutex: This is displayed by itself on the wallet comparison page's pie slices when hovering over the pie slice, so the use of "only" may be confusing in that context because it's not obviously comparable to anything else.

I suggest "Supports non-chain-specific ENS addresses" or something less ambiguous like that. The constraint is that it has to fit in a tooltip, so has to be short.

src/schema/attributes/ecosystem/address-resolution.ts
```diff
@@ -313,30 +319,33 @@ export const addressResolution: Attribute<AddressResolutionValue> = {
...
 					},
 					[],
 				),
 			),
+		],
+		partial: [
 			exampleRating(
 				mdSentence(
-					`The wallet resolves ${eipMarkdownLink(erc7828)} or ${eipMarkdownLink(erc7831)} addresses using an offchain provider in a verifiable and privacy-preserving manner.`,
+					'The wallet resolves plain ENS addresses (`username.eth`) when sending tokens, using onchain data for resolution.',
```
polymutex: Here it's less confusing to use "plain", because this will show in a list with the other possibilities just next to it, but for consistency it may still be better to use "non-chain-specific" instead of "plain".

polymutex (changes requested):
Logic looks perfect, just one wording change.

polymutex (approved):

polymutex:
It seems the build is failing ("Wallet did not match any of the expected variants: mobile | desktop | browser"). I will investigate...

polymutex:
> It seems the build is failing ("Wallet did not match any of the expected variants: mobile | desktop | browser"). I will investigate...

Fixed in e3d29681b49ba7bd0a677a69c9b58721e457fdbe.

PR #529, opened 2026-02-12
Schema: remove unused `__brand` types from rating values

darrylyeo:
We don't meaningfully use these for type discrimination.


polymutex (approved):
Yeah, in retrospect this was pretty useless.

PR #532, opened 2026-02-13
Add railgun support to private transfers 

johngrantuk:
Ref #192 

Adding Railgun support to private transfers. I think the features/evaluation makes sense but would definitely be worth a double check.

While testing I noticed an issue that WalletPage wasn't rendering correctly. Added PrivateTransfersDetails.  
To test I updated Ambire wallet with (can also change `defaultFungibleTokenTransferMode` to Railgun and see the change):
```
transactionPrivacy: {
...,
[PrivateTransferTechnology.RAILGUN]: supported({
	ref: refNotNecessary,
	broadcasterFee: comprehensiveFeesShownByDefault,
	crossContractCalls: featureSupported,
	privateTransfers: featureSupported,
	transactionSubmission: {
		type: 'BROADCASTER_ONLY',
		broadcasterLearnsUserIpAddress: false,
		customizableBroadcaster: featureSupported,
	},
	warnAboutSuccessiveOperations: featureSupported,
}),
}
```


src/views/attributes/privacy/PrivateTransfersDetails.svelte
polymutex, line 1: Wow, did this component really not exist? I guess no wallet has ever needed it so far (sigh)... Thanks for creating it.

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,77 @@ export type PrivacyPoolsSupport = WithRef<{
...
+			type: 'BROADCASTER_ONLY'
+
+			/**
+			 * Can the broadcaster endpoint be customized?
+			 */
+			customizableBroadcaster: Support
+
+			/**
+			 * Can the broadcaster learn the user's IP address?
+			 * Should be false if using Waku Network for IP protection.
```
polymutex: Waku renamed itself to Logos.

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,77 @@ export type PrivacyPoolsSupport = WithRef<{
...
+	 * in quick succession, potentially leading to time-based correlation?
+	 */
+	warnAboutSuccessiveOperations: Support
+
+	/**
+	 * How are transactions submitted? Broadcasters are required for transactions
+	 * FROM shielded addresses (private transfers, unshielding), but NOT for
+	 * shielding (depositing into Railgun).
+	 */
+	transactionSubmission: RailgunTransactionSubmission
```
polymutex: Maybe this nuance should be in the field name (`shieldedTransactionSubmission`) and type name (`RailgunShieldedTransactionSubmission`). Though maybe "shielded" can still be thought as including "shielding"...?

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,77 @@ export type PrivacyPoolsSupport = WithRef<{
...
+
+	/**
+	 * Is the fee taken by broadcasters displayed in the UI?
+	 */
+	broadcasterFee: FeeDisplay
```
polymutex: Having this only makes sense if the wallet supports broadcaster-based submission. Can a self-relayed-transaction-only wallet exist? Presumably.

Overall I think the `transactionSubmission` field can be split into two fields, like this:

```typescript
broadcasterBasedTransactionSubmission: Support<{
  customizableBroadcaster: Support
  broadcasterLearnsUserIpAddress: boolean
  broadcasterFee: FeeDisplay
}>
selfRelayedTransactionSubmission: Support
```

This way the broadcaster-based transaction fields only need to be specified if the wallet actually supports that, self-relayed transaction support is a separate dimension.

You can add a type constraint to `RailgunSupport` afterwards like `& ({broadcasterBasedTransactionSubmission: Supported<...>} | {selfTransactionSubmission: Supported})` to ensure that it's not possible to set both of those fields to "unsupported" state at once.

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,306 @@ function ratePrivacyPoolsSupport(
...
+				spendingPrivacy: PrivateTransfersPrivacyLevel.NOT_FULLY_IMPLEMENTED,
+				spendingDetails: mdParagraph(`
+					Private transfers are not supported; this means spending shielded funds
+					requires unshielding, which may expose transaction details.
+				`),
+				spendingImprovements: ['implement private transfers between Railgun wallets'],
+			}
+		}
+
+		if (railgun.transactionSubmission.type === 'BROADCASTER_OR_SELF_RELAY') {
```
polymutex: I think we should not penalize wallets for supporting a superset of wallet features. If it supports both broadcaster-based and self-relayed transactions, that should be at least as well-rated as supporting only broadcaster-based transactions.

What matters here is that for wallets that support both, which type is used by default? This can be added as an extra field in the feature type (`defaultTransactionSubmissionType` or something) and then you can use that here instead.

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,306 @@ function ratePrivacyPoolsSupport(
...
+			`),
+			sendingImprovements: [],
+		}
+	})()
+	const { receivingPrivacy, receivingDetails, receivingImprovements } = ((): {
+		receivingPrivacy: PrivateTransfersPrivacyLevel
+		receivingDetails: MarkdownParagraph<WalletNameStrings>
+		receivingImprovements: string[]
+	} => {
+		// Balance lookup is fully private (done locally, wallet syncs merkle tree)
```
polymutex: Can this code really make this assumption without any feature data to back this up?
Doesn't seem inconceivable to me that some wallet would be tempted to do server-side syncing of the merkle tree and offering wallet users a read API to it for example... (which would still be `CHAIN_DATA_PRIVATE`)

polymutex (changes requested):
Thank you, this is great for a first pass. I'm not completely familiar with how Railgun works (my mental model is basically "it's roughly Aztec") to a degree where I'd feel comfortable listing all of the privacy problems that Railgun integration implementations could have. Seeing the feature type here, it seems you have, but I am still concerned about the `receiving` part of the equation (sure the merkle tree sync can happen locally; in practice, will wallets actually do it?)

Curious to know if there are also any other privacy leak vector that could exist from a Railgun integration in general.

src/schema/features/privacy/transaction-privacy.ts
johngrantuk: Yeah, I left it as Waku because the Railgun docs all reference Waku and I thought a user would likely cross reference. I've updated to "Logos (previously Waku)" where appropriate but let me know if you prefer some alternative.

src/schema/features/privacy/transaction-privacy.ts
johngrantuk: This has been removed in favour of the changes below.

src/schema/features/privacy/transaction-privacy.ts
johngrantuk: I've refactored this now. Let me know if it makes sense or if it can be improved.

src/schema/attributes/privacy/private-transfers.ts
johngrantuk: Thanks for pointing this out, I agree. Have made the change.

src/schema/attributes/privacy/private-transfers.ts
johngrantuk: Thanks for pointing this out. I didn't fully appreciate the importance of the link between the protocol and the wallet integration. Now I think of it as the protocol could be fully private but a particular integration might not be. I updated:
* Added a new field to RailgunSupport type to track merkle tree syncing location - Wallet or External
* Updated the receiving privacy evaluation logic to check this field

johngrantuk:
> Thank you, this is great for a first pass. I'm not completely familiar with how Railgun works (my mental model is basically "it's roughly Aztec") to a degree where I'd feel comfortable listing all of the privacy problems that Railgun integration implementations could have. Seeing the feature type here, it seems you have, but I am still concerned about the `receiving` part of the equation (sure the merkle tree sync can happen locally; in practice, will wallets actually do it?)
> 
> Curious to know if there are also any other privacy leak vector that could exist from a Railgun integration in general.

I investigated a bit more and added some more privacy attributes. I think its fairly comprehensive now and covers:

- Broadcaster IP address leakage (`broadcasterLearnsUserIpAddress`)
- Self-relay IP exposure (`selfRelayedTransactionSubmission`, `defaultTransactionSubmissionType` via `RailgunTransactionSubmissionVariant`)
- Merkle tree sync privacy (`merkleTreeSync`)
- Time-based correlation (`warnAboutSuccessiveOperations`)
- Broadcaster customization (`customizableBroadcaster`)
- Shielding correlation warnings (`warnAboutShieldingCorrelation`)
- Unshielding destination correlation warnings (`warnAboutUnshieldingDestinationCorrelation`)
- Viewing key security warnings (`warnAboutViewingKeySharing`)

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,110 @@ export type PrivacyPoolsSupport = WithRef<{
...
+
+	/**
+	 * When scanning for received funds, is the Railgun UTXO merkle tree synced
+	 * and decrypted entirely by the wallet on the client side, or is it synced
+	 * by an external service? This matters for privacy: if syncing is done
+	 * server-side, the external provider can learn about received funds even
+	 * though the chain data itself doesn't reveal this information. This would
+	 * be CHAIN_DATA_PRIVATE rather than FULLY_PRIVATE.
+	 */
+	merkleTreeSync: 'WALLET_SIDE' | 'EXTERNAL'
```
polymutex: I suggest renaming `WALLET_SIDE` to `ON_USER_DEVICE`. We've had past contributions from wallet development teams arguing that the wallet developer's cloud infrastructure should count as "first party" (aka "wallet-side"), which isn't the intent of what we want to encode here. `ON_USER_DEVICE` makes this unambiguous.

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,110 @@ export type PrivacyPoolsSupport = WithRef<{
...
+	 */
+	broadcasterLearnsUserIpAddress: boolean
+
+	/**
+	 * Is the fee taken by broadcasters displayed in the UI?
+	 */
+	broadcasterFee: FeeDisplay
+}
+
+type RailgunTransactionSubmissionVariant =
```
polymutex: The word "Variant" has a specific meaning in Walletbeat's codebase, referring to the platform a wallet runs on (mobile/desktop/browser extension/hardware/embedded wallet). So I think it's better to not use it in contexts where it has a different meaning. Consider `RailgunTransactionSubmissionMethods` or something similar.

src/schema/features/privacy/transaction-privacy.ts
```diff
@@ -406,3 +414,110 @@ export type PrivacyPoolsSupport = WithRef<{
...
+	 */
+	warnAboutViewingKeySharing: Support
+
+	/**
+	 * When scanning for received funds, is the Railgun UTXO merkle tree synced
+	 * and decrypted entirely by the wallet on the client side, or is it synced
+	 * by an external service? This matters for privacy: if syncing is done
+	 * server-side, the external provider can learn about received funds even
+	 * though the chain data itself doesn't reveal this information. This would
+	 * be CHAIN_DATA_PRIVATE rather than FULLY_PRIVATE.
```
polymutex: While this is all useful and accurate information, the intent of feature types (`src/schema/features`) is to only represent wallet behavior, not the intent behind collecting this data about wallets or how that data is later interpreted by attributes' evaluation code in `src/schema/attributes`. So, while explaining what this wallet feature data is about and how that can impact user privacy is good, it's not the right place to refer to specifics of the evaluation code (e.g. `CHAIN_DATA_PRIVATE` vs `FULLY_PRIVATE` is an evaluation-side-only concern).

(Just a long-winded way of asking to remove the last "This would be CHAIN_DATA_PRIVATE rather than FULLY_PRIVATE" sentence.)

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,409 @@ function ratePrivacyPoolsSupport(
...
+					'default to broadcaster usage instead of self-relay for better privacy',
+				],
+			}
+		}
+
+		// Broadcaster is used (either broadcaster-only or both with broadcaster default)
+		if (broadcasterData.broadcasterLearnsUserIpAddress) {
+			if (!isSupported(broadcasterData.customizableBroadcaster)) {
+				return {
+					spendingPrivacy: PrivateTransfersPrivacyLevel.NOT_PRIVATE,
```
polymutex: This is still `CHAIN_DATA_PRIVATE`, right?

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,409 @@ function ratePrivacyPoolsSupport(
...
+			}
+		}
+
+		const broadcasterSupported = isSupported(railgun.broadcasterBasedTransactionSubmission)
+		const selfRelaySupported = isSupported(railgun.selfRelayedTransactionSubmission)
+
+		// If only self-relay is supported, this exposes IP address
+		if (!broadcasterSupported && selfRelaySupported) {
+			return {
+				spendingPrivacy: PrivateTransfersPrivacyLevel.NOT_PRIVATE,
```
polymutex: This is still `CHAIN_DATA_PRIVATE`, right?

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,409 @@ function ratePrivacyPoolsSupport(
...
+				spendingDetails: mdParagraph(`
+					Private transfers are not supported; this means spending shielded funds
+					requires unshielding, which may expose transaction details.
+				`),
+				spendingImprovements: ['implement private transfers between Railgun wallets'],
+			}
+		}
+
+		const broadcasterSupported = isSupported(railgun.broadcasterBasedTransactionSubmission)
+		const selfRelaySupported = isSupported(railgun.selfRelayedTransactionSubmission)
```
polymutex: I suggest not having these as `const` booleans, and instead just inline the `isSupported(...)` checks in the `if ()` blocks below. It's wordier, but `isSupported` is a type assertion function, so using it this way lets the TypeScript compiler understand the type implications better, without the need to have code that `throw`s errors.

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,409 @@ function ratePrivacyPoolsSupport(
...
+		// If both are supported and default is self-relay, this exposes IP address
+		if (
+			selfRelaySupported &&
+			'defaultTransactionSubmissionType' in railgun &&
+			railgun.defaultTransactionSubmissionType === 'SELF_RELAY'
+		) {
+			// Default is self-relay, which exposes IP address
+			if (broadcasterData.broadcasterLearnsUserIpAddress) {
+				return {
+					spendingPrivacy: PrivateTransfersPrivacyLevel.NOT_PRIVATE,
```
polymutex: Unless I'm missing something, I think this is still `CHAIN_DATA_PRIVATE`.

src/schema/attributes/privacy/private-transfers.ts
polymutex: Same here, I think this is `CHAIN_DATA_PRIVATE`.

src/schema/attributes/privacy/private-transfers.ts
```diff
@@ -1226,6 +1258,409 @@ function ratePrivacyPoolsSupport(
...
+					to censor or block transactions.
+				`),
+			)
+		}
+
+		if (!isSupported(railgun.warnAboutViewingKeySharing)) {
+			extraNotes.push(
+				mdParagraph(`
+					The wallet does not warn users about the privacy risks of sharing viewing
+					keys. Viewing keys are encoded in 0zk addresses and are irrevocable, meaning
```
polymutex: You can put backticks around "0zk", here and in other places where this shows up.

polymutex (approved):
Thanks for looking more deeply into Railgun. I think this looks great on methodology, the comments below are just rewordings.

src/schema/attributes/privacy/private-transfers.ts
johngrantuk: Yeah, this should be CHAIN_DATA_PRIVATE because the chain data itself is private and only an external provider learns the IP.  Changed.

polymutex (approved):

polymutex:
Thanks again for the PR, this is excellent work. Merged :heavy_check_mark:

PR #541, opened 2026-02-21
Feat: Add rainbow wallet data

maykelxyz:
# Add Rainbow wallet data
- Transaction legibility
- EIP supports
- Address resolution

## References

image

image

image

image

image

image

image

image

### DIFF
--- a/data/software-wallets/rainbow.ts
+++ b/data/software-wallets/rainbow.ts
@@ -8,7 +8,15 @@ import {
 	HardwareWalletType,
 	type SupportedHardwareWallet,
 } from '@/schema/features/security/hardware-wallet-support'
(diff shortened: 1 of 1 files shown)


polymutex (approved):
Maybe we should start putting these screenshots in the repo. We could show them similarly to how we should `ref`erences in attribute evaluations on the per-wallet page. Thoughts?

maykelxyz:
That's a good idea. Only problem I can think of is how to make it not look like clutter

polymutex:
> That's a good idea. Only problem I can think of is how to make it not look like clutter

Probably an image carousel in the attribute section, or a modal when clicking the references. Either way, we can figure the UI part out separately from the task of committing the screenshot to the repo.

maykelxyz:
Oh yeah, that's a good idea. Would love to see it! :)

PR #542, opened 2026-02-23
Fix: Update transaction details display for ambire

maykelxyz:
Ambire's `ComplexBenchmarkTransactions.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND` doesnt have any calldata decoded and transaction outcome is not explained too.

I’m also starting to think `transactionOutcome` shouldn’t apply to Safe transactions.
In Safe flows, the “transaction” the wallet is presenting to the signer is basically confirming a Safe tx / approving a hash, not executing the underlying calls. The actual approve/supply/transfer outcome happens when the Safe executes

cc: @polymutex

### DIFF
--- a/data/software-wallets/ambire.ts
+++ b/data/software-wallets/ambire.ts
@@ -638,8 +638,8 @@ export const ambire: SoftwareWallet = {
 					[ComplexBenchmarkTransactions.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND]:
 						{
 							...ambireTransactionDisplayDefault,
-							calldataDecoded: DataDisplayOptions.SHOWN_BY_DEFAULT,
-							transactionOutcome: TransactionOutcome.EXPLAINED,
+							calldataDecoded: DataDisplayOptions.NOT_IN_UI,
+							transactionOutcome: TransactionOutcome.NOT_EXPLAINED,
 						},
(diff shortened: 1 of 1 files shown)


polymutex (approved):

polymutex:
> I’m also starting to think `transactionOutcome` shouldn’t apply to Safe transactions. In Safe flows, the “transaction” the wallet is presenting to the signer is basically confirming a Safe tx / approving a hash, not executing the underlying calls. The actual approve/supply/transfer outcome happens when the Safe executes

If so, then how is the user supposed to understand the effect of the transaction they are about to add their signature to? Or is the idea that only the last user to sign (the one for whom adding the signature crosses the threshold) would see the transaction outcome? If so, wouldn't that mean that `k-1` signers are essentially blind signing (all they get is calldata)?

maykelxyz:
I agree blind signing isn’t acceptable. My concern is scope: `transactionOutcome` was originally about what the signer is doing. In Safe flows, the signer is just confirming a Safe tx, the actual execution happens from the Safe.

Safe does provide simulation in their app, but we shouldn’t rely on trust assumptions around UI/simulation alone. If we keep `transactionOutcome`, I think the solution is to explicitly define it for Safe as an “effect preview of what the Safe will execute,”.

So yeah, we can update the docs to clarify this distinction so the intent is unambiguous.

polymutex:
Sounds good.

PR #543, opened 2026-02-23
Feat: Add metamask data

maykelxyz:
### Changes
- Added metamask's transaction legibility data which updates metamask's transaction legibility to `PARTIAL`
- Also added image references for when I tested metamask's transaction legibility data.
  - This is still not being used in any references as our current structure doesn't support image as references. Will import it once it does.
- Added metamask's delegation data
- Added metamask's bug bounty program

Which makes metamask's remaining data to be filled:
- Wallet address privacy
- Multi address privacy
- App isolation
- Transaction inclusion

Issue #93 for pre-launch


polymutex (approved):

maykelxyz:
Merged, thanks @polymutex!

PR #545, opened 2026-02-24
Fix: Add comparison of wallets for sorting

### LINKED ISSUE #539: Stage system should take precedence over internal scoring for ordering purposes
It is currently technically possible for a stage 0 wallet to be sorted higher than a stage 1 wallet on the main comparison table, if it has great scores on the attributes that stage 1 doesn't look at.

We should make the score have a `stage number * 1000000` component added to it or similar, to ensure that stages always take precedence over individual attributes.

maykelxyz:
The default sort is by Overall Rating, but UNRATED attributes are excluded from the average rather than counting as zero. This meant a wallet with few rated attributes (all PASS) could outscore a well-documented Stage 1 wallet and appear above it.

Fixed by adding a custom `compare` to the Overall Rating column: wallets are sorted by highest cleared stage first, with the attribute score as a tiebreaker within the same stage.

A suggestion by @polymutex was adding `stageIndex * 1_000_000` to the score. The problem is this is stage 0 maps to `stageIndex = 0`, giving no bonus. Wallets that cleared Stage 0 would still mix with wallets that cleared no stage at all. A custom comparison avoids this.

Fixes #539

### DIFF
--- a/src/views/WalletTable.svelte
+++ b/src/views/WalletTable.svelte
@@ -575,6 +575,24 @@
 									sort: {
 										isDefault: true,
 										defaultDirection: SortDirection.Descending,
+										// Stages always take precedence over attribute scores when sorting:
+										// a stage 1 wallet should never appear below a stage 0 wallet
+										// regardless of how good its attribute scores are. Within the
+										// same stage, fall back to the attribute score as a tiebreaker.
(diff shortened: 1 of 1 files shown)


polymutex:
Can you just have "no stage" map to `-1`?

polymutex (approved):
Feel free to submit either way

PR #552, opened 2026-02-25
Docs: Improve docs guides

maykelxyz:
# Docs: Improve contributor guides and add Claude Code wallet-data skill

Most of this PR is about making it easier for contributors, both technical and non-technical, to fill in wallet feature data without guessing. Almost every feature schema file now has inline how-to guidance that explains what each field means, when to use `null`, and how to actually verify a wallet's behavior.

## Changes

- **Inline contributing guides** added across 18 feature schema files (privacy, security, self-sovereignty, transparency)
- **New Claude Code skill** — `/wallet-data-contributor` walks contributors through the full contribution flow interactively: entity setup, contributor file, skeleton generation, and populating every feature field. Great for first-timers

## Areas I'd love to get double-checked

A lot of the self-sovereignty and keys-handling logic was originally written by @polymutex and I want to make sure I documented it correctly. 

- `self-sovereignty/chain-configurability.ts`
- `self-sovereignty/transaction-submission.ts`
- `account-support.ts`
- `security/account-recovery.ts`
- `security/keys-handling.ts`

Issue #215, opened 2025-07-04
RESEARCH: Incorporate ideas from Coinspect


PatrickAlphaC:
https://www.coinspect.com/wallets/testing/

I'm not sure if we've read this, but their testing seems quite comprehensive as well. It might make sense to take some ideas and incorporate them here.

polymutex:
Agreed, it's a great set of security attributes. We should incorporate them into ours over time, and we should probably also link to Coinspect security ratings on per-wallet pages in the Security section.

juli:
Hi all Patrick and Polymutex! Juliano from Coinspect here 👋

We already maintain a **security-only, objective, transparent** wallet evaluation and want it to integrate straight into Walletbeat.

* **Ranking (per platform):** [https://www.coinspect.com/wallets/](https://www.coinspect.com/wallets/)

* **Methodology & checks:** [https://www.coinspect.com/wallets/testing/](https://www.coinspect.com/wallets/testing/)

* **WSVS checklist (MIT):** [https://github.com/coinspect/wallet-security-verification-standard](https://github.com/coinspect/wallet-security-verification-standard)

We also run a responsible-disclosure process with per-check reports (screenshots/notes) shared privately with wallet teams, and we use **[AHP](https://en.wikipedia.org/wiki/Analytic_hierarchy_process)** to weight checks with input from multiple experts.

This type of collaboration has been our intent from the start: **[WSVS]( https://github.com/coinspect/wallet-security-verification-standard)** was published two years ago, and in **June 2024** we proposed adding security filters to ethereum.org's "Find wallet" with results hosted neutrally. We've stayed consistent with that plan: [https://github.com/ethereum/ethereum-org-website/issues/13144](https://github.com/ethereum/ethereum-org-website/issues/13144)

### **What we’ll provide to Walletbeat**

* **JSON feed** of our public results per wallet/platform.

* **Versioned checks** and **timestamped scores**.

* **Licensing:** data **CC BY 4.0**, code/checks **MIT**.

* **Coverage updates:** we’ll prioritize adding any software wallets Walletbeat lists that we don’t test yet.

* Happy to open a PR with `docs/coinspect-integration.md` to define the JSON schema and mapping.

### **What we ask**

* **Display our security score** ("**Security (Coinspect)**: 82/100 • last updated Aug 2025") with a small link in each wallet page's Security section.

[...]

PR #568, opened 2026-03-04
JSON export of rated wallet data

### LINKED ISSUE #167: Structured JSON export of rated wallet data
We should provide JSON exports of fully-rated wallet data.

The wallet data in `/data/wallets` is good but only covers feature data. A full `RatedWallet` JSON export would include this, but also all the attribute rating data, rationale for why each attribute has each rating, etc.

Another difference is that it would also be fully expanded across wallet variants (mobile/desktop/etc) and would have no helper functions, whereas the data in `/data/wallets` is optimized for human readability/terseness/ease of filling out. The fully-expanded JSON blob is for machine consumption, so verbosity isn't as much of an issue.

johngrantuk:
Closes #167 

Adds a JSON export for each rated wallet and refactors shared content so the site and export stay in sync.

See example of JSON Export for Metamask [here](https://gist.github.com/johngrantuk/753ead94adf420400aa55afce4e03bee#file-example-metamask-json-L2).

## New

- **Per-wallet JSON endpoint**: `GET /[walletName]/index.json` returns a full export for that wallet (prerendered at build time).
- **Export shape**:
  - Base: `walletId`, `displayName`, `description`, `lastUpdated`, `types`, `variants`, `stage`, `stageBreakdown`, optional `website`/`repository`.
  - **`overall`**: attribute groups export aggregating evaluation across all variants (same structure as per-variant blocks).
  - **Per-variant attribute groups**: one key per variant (e.g. `BROWSER`, `MOBILE`) with full attribute-group evaluation (attribute metadata + ratings, methodology, references).
  - **Stage breakdown**: when a wallet has a stage ladder, `stageBreakdown` lists each stage with `stageId`, `label`, `passedCount`/`totalCount`, `status`, and criteria groups (criterion id, attribute, description, rating).
- **JSON schema**: `tests/fixtures/rated-wallet-export-schema.json` documents the export shape; tests validate every wallet's export against it.

## Refactors

- **Reusable content helpers** (used by both markdown and JSON):
  - `evaluation-content.ts`: shared strings for evaluations (e.g. criterion descriptions, group descriptions, "how is this evaluated").
  - `attribute-display.ts`: shared attribute display strings (e.g. "How is this evaluated?" heading).
  - `stage-attributes.ts`: stage/criteria breakdown (counts, status, criterion→attribute mapping) shared by `WalletStageOverview` and the JSON export.
- **Methodology / wallet page markdown**: `methodology-markdown.ts` and `wallet-page-markdown.ts` simplified to use these helpers; `WalletStageOverview.svelte` now uses `stage-attributes` for the stage breakdown instead of duplicating logic.

## Testing

- `tests/wallet-json-export.test.ts`: for every rated wallet, checks presence and consistency of base fields, `stage`/`stageBreakdown`, per-variant attribute groups, and schema validation via `assertValidJson` + the fixture schema.


src/utils/attribute-display.ts
```diff
@@ -0,0 +1,40 @@
...
+	if (wording.midSentenceName === null) {
+		return collapseToSingleLine(
+			renderTypographicContentToString(wording.whatCanWalletDoAboutIts, {
+				WALLET_NAME: walletName,
+			}),
+		)
+	}
+
+	return `What can ${walletName} do about its ${wording.midSentenceName}?`
+}
```
polymutex, line 40: These should be used on the HTML version too.

public/schemas/rated-wallet.schema.json
polymutex, line 1: Having a JSON schema is great, but relegating it to test-only purposes feels like wasted work. How about exporting that somewhere as a static file? The wallet JSON exports can also reference its URL under their `$schema`.

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,76 @@
...
+	isTypographicContent,
+	type Paragraph,
+	renderTypographicContentToString,
+} from '@/types/content'
+
+/**
+ * Build the wallet name and pseudonym strings used when rendering evaluation content.
+ * Single place to avoid drift when adding more placeholder keys.
+ */
+export function getWalletEvalStrings(wallet: RatedWallet): WalletNameAndPseudonymStrings {
```
polymutex: Since this is now a shared function moved out the the methodology-markdown.ts file, it should also start being used in all the other places where such objects are created (example: src/views/WalletPage.svelte, src/views/attributes/privacy/AddressCorrelationDetails.svelte, others; grep for WALLET_PSEUDONYM_SINGULAR:)

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Render stage criterion or group description content to plain text.
+ * Centralizes "content → string": typographic content is rendered with eval strings and trimmed;
+ * custom content yields empty string.
+ * Accepts Paragraph (e.g. from stage definitions) or Content<WalletNameAndPseudonymStrings>.
+ */
+function descriptionContentToText(
+	content: Content<WalletNameAndPseudonymStrings> | Paragraph,
+	evalStrings: WalletNameAndPseudonymStrings,
+): string {
```
polymutex: I don't really see the point of having these two functions simultaneously... Feels like just one would do the work of the other.

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Render stage criterion or group description content to plain text.
+ * Centralizes "content → string": typographic content is rendered with eval strings and trimmed;
+ * custom content yields empty string.
+ * Accepts Paragraph (e.g. from stage definitions) or Content<WalletNameAndPseudonymStrings>.
+ */
+function descriptionContentToText(
+	content: Content<WalletNameAndPseudonymStrings> | Paragraph,
```
polymutex: Weird argument type; isn't `Paragraph` already implied by `Content`?

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Criteria group description as plain text (for JSON export or further markdown processing).
+ */
+export function renderGroupDescriptionToText(
+	group: WalletStageGroup,
+	evalStrings: WalletNameAndPseudonymStrings,
+): string {
+	return descriptionContentToText(group.description, evalStrings)
+}
```
polymutex: These functions are trivial, IMO should just be inlined wherever they are used.

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+function descriptionContentToText(
+	content: Content<WalletNameAndPseudonymStrings> | Paragraph,
+	evalStrings: WalletNameAndPseudonymStrings,
+): string {
+	if (!isTypographicContent(content)) {
+		return ''
+	}
+
+	return renderTypographicContentToString(content, evalStrings).trim()
```
polymutex: `trimWhitespacePrefix` from `src/types/utils/text.ts` would be a more solid option here

src/utils/stage-attributes.ts
```diff
@@ -102,6 +153,29 @@ export const isAttributeUsedInStage = (
 export const getCriterionAttributeId = (criterion: WalletStageCriterion): string | null =>
 	getEvaluateFunctionAttributeId(criterion.evaluate)
 
+// TODO: https://github.com/walletbeat/walletbeat/issues/555
```
polymutex: This issue is now fixed. Please rebase and use the now-existing `displayName` attribute

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+				"criteria": {
+					"type": "array",
+					"items": { "$ref": "#/$defs/stageCriterionBreakdownItem" }
+				}
+			}
+		},
+		"stageCriterionBreakdownItem": {
+			"type": "object",
+			"additionalProperties": false,
+			"required": ["criterionId", "attributeId", "attributeDisplayName", "description", "rating"],
```
polymutex: Add `criterionDisplayName`. Or maybe replace `attributeDisplayName` with `displayName` (and set it to the criterion's display name, not encoding the attribute display name here)

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+			"description": "One criterion within a stage (mirrors markdown bullet with attribute link and rating).",
```
polymutex: Seems odd to refer to the markdown version of the wallet page in a JSON schema. (If anything references anything else, it should probably be the other way around...)

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+			"type": "object",
+			"description": "One attribute group: map of attribute id to attribute export block (attribute metadata + rating).",
+			"additionalProperties": {
+				"$ref": "#/$defs/attributeExportBlock"
+			}
+		},
+		"attributeExportBlock": {
+			"type": "object",
+			"additionalProperties": false,
+			"required": ["attribute", "rating"],
```
polymutex: I suggest replacing `rating` with `evaluation`, similar to the way the TypeScript type does. This keeps the word "rating" to be strictly associated to the PASS/PARTIAL/FAIL enum rather than overloading its meaning.

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+		"ratingBlock": {
+			"type": "object",
+			"additionalProperties": false,
+			"required": ["rating", "shortExplanation", "details"],
+			"description": "Wallet-specific evaluation result for one attribute (rating, explanation, details, optional impact, howToImprove, references).",
+			"properties": {
+				"rating": {
+					"type": "string",
+					"enum": ["PASS", "FAIL", "PARTIAL", "UNRATED", "EXEMPT"],
+					"description": "Attribute rating for this wallet."
```
polymutex: Add explanation of the values here

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+				},
+				"shortExplanation": {
+					"type": "string",
+					"description": "Brief explanation of why this rating was assigned."
+				},
+				"details": {
+					"type": "string",
+					"description": "Longer explanation or summary of the evaluation; may be \"See full details on the wallet page.\" when details are component-based."
```
polymutex: "when details are component-based" is an implementation detail. I think "when details are not easily expressed in textual format" would be a more helpful description.

public/schemas/rated-wallet.schema.json
```diff
@@ -0,0 +1,297 @@
...
+	"required": [
+		"walletId",
+		"types",
+		"variants",
+		"overall",
+		"displayName",
+		"description",
+		"lastUpdated",
+		"stage",
+		"stageBreakdown"
```
polymutex, line 15: Maybe this can be a follow-on PR, but I was assuming we'd want all the feature data here too. This is both because that's what we'd like to consume from other data sources (e.g. Coinspect, see #215) so we should reciprocate the level of detail on offer here, and also because it's useful for projects downstream of Walletbeat like https://walletradar.org/ that consume WB's feature data, not the attribute/rating/methodology data.

Practically speaking, I think it would be easier to implement this part dynamically (i.e. dynamically generating the `features` part of the JSON schema from the TypeScript type definitions) rather than hardcoding it. Note #552 which is implementing logic that reads through these TypeScript type definitions, which could probably be reused here for JSON schema auto-generation.

For now I'd recommend leaving a TODO and just having the feature data as an opaque `object` in the schema with no predefined named properties.

tests/fixtures/rated-wallet-export-schema.json
```diff
@@ -0,0 +1,297 @@
...
+			"stage",
+			"stageBreakdown",
+			"website",
+			"repository",
+			"overall",
+			"MOBILE",
+			"DESKTOP",
+			"BROWSER",
+			"EMBEDDED",
+			"HARDWARE"
```
polymutex: Can we have these under a `perVariant` object or something? It feels off to have them at the top level and following a different casing convention than the other fields.

tests/wallet-json-export.test.ts
polymutex, line 1: Aren't a lot of these tests redundant with the schema check?

src/utils/stage-attributes.ts
```diff
@@ -102,6 +153,29 @@ export const isAttributeUsedInStage = (
...
+ * Resolve display name for a stage criterion: attribute display name, then attribute ID, then criterion id formatted.
+ */
+export function getCriterionDisplayName(
+	criterion: WalletStageCriterion,
+	attributeId: string | null,
+	attribute: Attribute | null,
+): string {
+	return attribute?.displayName ?? attributeId ?? criterionIdToDisplayName(criterion.id)
+}
+
```
polymutex: There is now a `displayName` field, so this function can be removed.

src/utils/stage-attributes.ts
```diff
@@ -40,6 +40,57 @@ export const attributesById = new Map(
 const allCriteriaInStage = (stage: WalletStage): WalletStageCriterion[] =>
 	stage.criteriaGroups.flatMap(criteriaGroup => criteriaGroup.criteria)
 
+/**
+ * Aggregate status for a stage or criteria group based on applicable criterion ratings.
+ * Rule: all exempt → UNRATED; else all passed → PASS; else any passed → PARTIAL; else FAIL.
```
polymutex: "All exempt" should not be a case that ever happens, you can throw an error in this case. "UNRATED" seems like the wrong value either way for this case.

polymutex (changes requested):

public/schemas/rated-wallet.schema.json
johngrantuk: I've followed your recommendation. I added the features as part of the perVariant alongside attributes. Hopefully that makes sense.

tests/wallet-json-export.test.ts
johngrantuk: Sorry, I lost track of this. Have removed the redundant tests and left in some that check correctness.

src/utils/evaluation-content.ts
```diff
@@ -0,0 +1,56 @@
...
+	renderTypographicContentToString,
+} from '@/types/content'
+import { trimWhitespacePrefix } from '@/types/utils/text'
+
+/**
+ * Concrete wallet eval strings (all fields present, no null).
+ * Used as return type of getWalletEvalStrings so it's assignable to Typography's
+ * strings prop when content uses WalletNameAndPseudonymStrings (which is a union including null).
+ */
+export type ConcreteWalletEvalStrings = {
```
polymutex: Move this next to `WalletNameAndPseudonymStrings`

src/utils/wallet-page-markdown.ts
```diff
@@ -220,13 +115,12 @@ export function walletPageMarkdown(wallet: RatedWallet, siteUrl: string): string
...
-					) {
-						const groupDesc = normalizeMarkdownBlankLines(
-							renderTypographic(criteriaGroup.description, walletName),
-						).trim()
+					const groupDescRaw = renderContentToText(criteriaGroup.description, evalStrings, {
+						trim: true,
+					})
+
+					if (groupDescRaw !== '') {
+						const groupDesc = normalizeMarkdownBlankLines(groupDescRaw).trim()
```
polymutex: Is there still a need to trim this given the use of `trimWhitespacePrefix` inside `renderContentToText`?

src/utils/wallet-page-markdown.ts
```diff
@@ -331,17 +224,15 @@ export function walletPageMarkdown(wallet: RatedWallet, siteUrl: string): string
...
 					parts.push(shortExpl.trim(), '')
 
 					const details = normalizeMarkdownBlankLines(
-						renderEvaluationContentOrFallback(
-							evaluation.details,
-							evalStrings,
-							`[See full details for ${attribute.displayName}](${walletAttrUrl})`,
-						),
+						renderContentToText(evaluation.details, evalStrings, {
+							fallback: `[See full details for ${attribute.displayName}](${walletAttrUrl})`,
```
polymutex, line 251: `trim: true` might be good here too

src/utils/stage-attributes.ts
```diff
@@ -40,6 +40,59 @@ export const attributesById = new Map(
 const allCriteriaInStage = (stage: WalletStage): WalletStageCriterion[] =>
 	stage.criteriaGroups.flatMap(criteriaGroup => criteriaGroup.criteria)
 
+/**
+ * Aggregate status for a stage or criteria group based on applicable criterion ratings.
+ * Rule: all passed → PASS; else any passed → PARTIAL; else FAIL.
+ * Throws if every criterion is EXEMPT (stage must have at least one applicable criterion).
+ */
+export type StageCountsStatus = 'PASS' | 'PARTIAL' | 'FAIL'
```
polymutex: I think this can also be `UNRATED` in case there is any `UNRATED` criterion.

src/views/WalletStageOverview.svelte
```diff
@@ -30,11 +29,6 @@
 			label: 'All criteria failed',
 			color: 'var(--rating-fail)',
 		},
-		[StageStatus.UNRATED]: {
-			icon: '❔',
-			label: 'All criteria unrated',
-			color: 'var(--rating-unrated)',
-		},
```
polymutex, line 37: Hm, this should still be possible, right? If any criterion is unrated, then so is the stage status.

public/schemas/rated-wallet.schema.json
polymutex, line 1: How about `rated-wallet.schema.json`? (the "`-export`" doesn't seem relevant)

src/utils/stage-attributes.ts
```diff
@@ -40,6 +40,59 @@ export const attributesById = new Map(
...
+
+/**
+ * Compute passed/total counts and aggregate status for a stage.
+ */
+export function computeStageCountsAndStatus(
+	stage: WalletStage,
+	wallet: StageEvaluatableWallet,
+): StageCountsAndStatus {
+	return computeCountsAndStatus(allCriteriaInStage(stage), wallet)
+}
```
polymutex: Seems trivial, can inline.

tests/utils/assert-valid-json.ts
```diff
@@ -0,0 +1,59 @@
...
+const parsedSchema: unknown = JSON.parse(schemaJson)
+
+if (typeof parsedSchema !== 'object' || parsedSchema === null) {
+	throw new Error('Rated wallet export schema must be a JSON object')
+}
+
+const schema = parsedSchema
+
+const ajv = new Ajv2020({ allErrors: true })
+const validate = ajv.compile(schema)
```
polymutex: This appears to be a stateful object (`validate.errors` persists), yet it is global. This means successive calls to `assertValidJson` will probably have confusing results. Can you wrap all this in a class such that each instance has its own AJV validator object instead? Alternatively, shove all these variables inside `assertValidJson` so that they remain local to the function, not global.

src/utils/wallet-json-export.ts
```diff
@@ -0,0 +1,414 @@
...
+	if (typeof x === 'string' || Array.isArray(x)) {
+		return true
+	}
+
+	if (typeof x === 'object' && x !== null) {
+		return Object.hasOwn(x, 'urls') || Object.hasOwn(x, 'url')
+	}
+
+	return false
+}
```
polymutex: Move to `reference.ts`.

src/utils/wallet-json-export.ts
```diff
@@ -0,0 +1,414 @@
...
+		return false
+	}
+
+	if (typeof x === 'string' || Array.isArray(x)) {
+		return true
+	}
+
+	if (typeof x === 'object' && x !== null) {
+		return Object.hasOwn(x, 'urls') || Object.hasOwn(x, 'url')
+	}
```
polymutex: Please use the existing type predicates in `reference.ts` here, they are a bit more robust than these. I think something like checking if `x` is `undefined` or `null` or an object that passes `isNoRef(x)  || isReference(x)` (or an array of objects where each item passes this) is all.

src/utils/wallet-json-export.ts
```diff
@@ -0,0 +1,414 @@
...
+	description: string
+	lastUpdated: string
+	stage: string | null
+	stageBreakdown: StageBreakdownItemJsonExport[] | null
+	website?: string
+	repository?: string
+}
+
+/** Export shape: base fields with perVariant object keyed by variant (e.g. BROWSER, MOBILE). */
+export type RatedWalletJsonExport = RatedWalletJsonExportBase
```
polymutex: I don't understand having this type, if it's equal to the `Base` version?

src/utils/wallet-json-export.ts
```diff
@@ -0,0 +1,414 @@
...
+		urls: ref.urls.map(u => ({ label: u.label, url: u.url })),
+	}))
+}
+
+/**
+ * Recursively serialize a value from ResolvedFeatures (or nested feature data) to a
+ * JSON-serializable plain value. Normalizes `ref` properties: real references become
+ * ReferenceJsonExport[]; refTodo/refNotNecessary become null.
+ */
+function serializeFeatureValue(value: unknown): unknown {
```
polymutex: Nice, but maybe this should just be a function declared locally inside of `serializeResolvedFeatures`, since there's no use for it elsewhere and it's just used for recursion

src/utils/wallet-json-export.ts
```diff
@@ -0,0 +1,414 @@
...
+
+	for (const criterion of group.criteria) {
+		const evaluation = criterion.evaluate(stageEvaluatableWallet)
+		const attributeId = getCriterionAttributeId(criterion)
+		const descText = renderContentToText(criterion.description, evalStrings, { trim: true })
+
+		criteria.push({
+			criterionId: criterion.id,
+			attributeId,
+			// String() ensures a primitive string type for ESLint when it doesn't resolve criterion.displayName across files.
```
polymutex: Why doesn't it?

polymutex (changes requested):

src/utils/wallet-page-markdown.ts
johngrantuk: Apparently yes:

The .trim() is still needed; it does something different from trimWhitespacePrefix.

trimWhitespacePrefix (used when renderContentToText is called with trim: true) only strips the longest common leading whitespace from each line. So it normalizes indentation line-by-line and does not remove leading or trailing newlines (or other whitespace) around the whole string.

.trim() removes leading and trailing whitespace (including newlines) from the entire string.

So for something like "\n\n line one\n line two\n\n":

After trimWhitespacePrefix: "\n\nline one\nline two\n\n" (indentation removed per line, outer newlines kept).
After .trim(): "line one\nline two" (outer newlines removed).

src/utils/stage-attributes.ts
johngrantuk: Have updated. Now its consistent as any UNRATED → result is UNRATED.

tests/utils/assert-valid-json.ts
johngrantuk: Thanks for pointing that out! I've refactored to a class and changed the test so there’s one RatedWalletExportValidator per wallet instead of a single shared instance for the whole file. This should guarantee isolation at a very small performance cost.

src/utils/wallet-json-export.ts
johngrantuk: Removed.

src/utils/wallet-json-export.ts
johngrantuk: TBH I'm not 100% sure. The suggestion by AI is cross-file type resolution can be weaker in ESLint. I've refactored to use a local variable with an explicit type which doesn't hide any potential issues.

src/utils/wallet-page-markdown.ts
polymutex: This would still fail for strings like

```
  foo = `
    hello
  there
`
```

Because this double-trim would result in `hello\nthere`, whereas really it should be `  hello\nthere` to preserve the relative indentation of the two lines.

I think you can update `trimWhitespacePrefix` to remove empty leading/trailing lines, and then stop calling `.trim()`.

src/schema/attributes.ts
```diff
@@ -13,6 +13,17 @@ export type WalletNameAndPseudonymStrings =
 			WALLET_PSEUDONYM_PLURAL: string | null
 	  }
 
+/**
+ * Concrete wallet eval strings (all fields present, no null).
+ * Used as return type of getWalletEvalStrings so it's assignable to Typography's
+ * strings prop when content uses WalletNameAndPseudonymStrings (which is a union including null).
+ */
+export type ConcreteWalletEvalStrings = {
```
polymutex: Use `interface` rather than `type` when the type is describable using an `interface`.

src/schema/reference.ts
```diff
@@ -234,6 +234,26 @@ export function toFullyQualified(
...
+export function isReferenceInput(x: unknown): x is ReferenceInput {
+	if (x === undefined || x === null || isNoRef(x)) {
+		return false
+	}
+
+	if (isReference(x)) {
+		return true
+	}
+
+	if (Array.isArray(x) && x.every((item): item is Reference => isReference(item))) {
```
polymutex: I think `x.every(isReference)` is sufficient and should communicate the same thing to the typechecker.

src/utils/wallet-json-export.ts
polymutex: Still seems to be there.

src/utils/wallet-page-markdown.ts
polymutex: This would still fail for strings like

```
  foo = `
    hello
  there
`
```

Because this double-trim would result in `hello\nthere`, whereas really it should be `  hello\nthere` to preserve the relative indentation of the two lines.

I think you can update `trimWhitespacePrefix` to remove empty leading/trailing lines, and then stop calling `.trim()`.

Alternatives:
- Add an optional `options` argument to `trimWhitespacePrefix` that can toggle this on/off
- Make `normalizeMarkdownBlankLines` do the leading/trailing newlines trimming, and move that function to `text.ts`

tests/wallet-json-export.test.ts
```diff
@@ -0,0 +1,95 @@
...
+					expect(payload.stageBreakdown.length).toBe(ladderEvaluation?.ladder.stages.length ?? 0)
+
+					for (const item of payload.stageBreakdown) {
+						expect(item.passedCount).toBeLessThanOrEqual(item.totalCount)
+					}
+				}
+			})
+
+			it('payload.variants matches variant keys in payload.perVariant', () => {
+				const perVariantKeys = Object.keys(payload.perVariant).filter((k): k is Variant =>
```
polymutex: `.filter(variantEnum.is)` should be sufficient

johngrantuk: Was giving a lint error but updated to: `k => variantEnum.is(k)` which works

src/utils/wallet-json-export.ts
johngrantuk: Actually removed this time!

src/utils/wallet-page-markdown.ts
johngrantuk: `trimWhitespacePrefix` now strips only leading and trailing blank lines (not the whole string), so relative indentation is preserved. Also update other .trim() call sites to keep same behaviour.

polymutex (approved):
Good to merge, thanks again for working on this :)

PR #571, opened 2026-03-05
Fix: Account Recovery and Account Unruggability details not rendering

### LINKED ISSUE #570: Bug: Wallet attributes not showing
Image

Image

maykelxyz:
## Account Recovery

- Wallets without guardian recovery (e.g. Ambire) showed an empty details section.
AccountRecoveryDetails.svelte only rendered content inside a `value.outcomes !== null` block and had no fallback.

```ts
{#if value.outcomes === null}
	<Typography
		content={{
			contentType: ContentType.MARKDOWN,
			markdown: trimWhitespacePrefix(`
				{{WALLET_NAME}} does not implement guardian-based account recovery.
				The user will lose access to their account if they lose their seed phrase.
			`),
		}}
		strings={{ WALLET_NAME: wallet.metadata.displayName }}
	/>
{:else}
```

## Account Unruggability
- Same issue and we forgot to add `AccountUnruggabilityDetails` to `WalletPage.svelte`

closes #570

### DIFF
--- a/src/views/WalletPage.svelte
+++ b/src/views/WalletPage.svelte
@@ -199,6 +199,7 @@
 	import Tooltip from '@/components/Tooltip.svelte'
 	import Typography from '@/components/Typography.svelte'
 	import AccountRecoveryDetails from './attributes/security/AccountRecoveryDetails.svelte'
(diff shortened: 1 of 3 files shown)


polymutex (approved):
We need to have a test that would have caught this...

maykelxyz:
> We need to have a test that would have caught this...

Filed #574

Issue #572, opened 2026-03-06
Have policy for communications with wallet development teams


polymutex:
We should have a documented, published policy about the forms of communications that may take place between Walletbeat and wallet development teams, and any disclosure/transparency expectations of such communications. The intent would be to discourage the creation of back-channels.

polymutex:
Some rough ideas:

- Already-public communications (e.g. GitHub issues/PRs, tweets) do not need to be disclosed, since they are public.
- The *existence* of such communication channels should be publicly disclosed, naming at least the set of Walletbeat contributors involved, the name of the associated wallet, and the date the channel was established.
- Questions on methodology/clarifications are definitely OK, as it is helpful for wallet development teams to request clarifications on Walletbeat criteria. However, if any such discussion ends up resulting in a methodology change (even just a wording/clarification change without an underlying policy change), this discussion should be published, possibly with names redacted but the associated communication channel and wallet name should be identified.
- Offers for compensation should be rejected unless they match universal terms that Walletbeat already offers to any other wallet. Right now that set is empty, since Walletbeat does not have a business/funding model that involves taking money from wallet development teams, and there are no near-term plans to change this. However, if that _were_ to change in the future, it must change in a way where every wallet development team gets consistent terms. Hence the principle that any compensation offer that deviates from such universal terms should be rejected.
  - (Side-note: Actually Walletbeat already accepts compensation from wallet development teams, but only in the form of _labor_: data entry for wallets, and non-wallet-specific code changes like UI fixes.)

maykelxyz:
Agree with @polymutex for the most part. I’d add that conversations with wallet teams should be treated as open to the public.

Our purpose of having channels with wallet teams is to have a more direct communication and coordination with wallet teams, not to create private channels where discussions are hidden from the public.

Parties participating in these discussions should assume they may be publicly disclosed

ren2140eth:
also agree with @polymutex I also think it could be worth adding that members of walletbeat will actively be reaching out to wallet team members via DM's on platforms like (X, farcaster, discord, etc.) in order to establish more formal communications and get their help with filling out data about their respective wallets on the site.

maykelxyz:
Completed with #575

PR #575, opened 2026-03-08
Add Walletbeat communications policy.

polymutex:
Seed with all the group chats I am aware of.

Fixes issue #572.

@darrylyeo @lucilapastore @nconsigny @maykelxyz @ren2140eth Please review, and let me know if I am missing any group chat.


darrylyeo (approved):

ren2140eth:
The group chats list looks good to me

ren2140eth (approved):
I've reviewed and accept the policy

maykelxyz (approved):
It was mentioned that we could consider only applying this to conversations where Walletbeat gets brought up starting from now, rather than retroactively.

Though, I don’t think the document specifies that clause. It might be worth clarifying that in the doc

polymutex:
Added.

hesterbruikman:
could you add me as reviewer please @polymutex ?

johngrantuk (approved):

polymutex:
@lucilapastore @hesterbruikman Still looking for your approval here.

lucilapastore:
I've reviewed and accept the policy. I just have one question: is our marketing group chat not included? I mean the list only covers external comms but not internal working groups?

Thanks

lucilapastore (approved):

polymutex:
> I've reviewed and accept the policy. I just have one question: is our marketing group chat not included? I mean the list only covers external comms but not internal working groups?
> 
> Thanks

Correct, the marketing group chat is not included because it's an internal communications chat and doesn't have any external wallet representation on it. Same as the Walletbeat Farcaster group chat.

PR #582, opened 2026-03-13
WIP: Wallet Attribute: Security best practices

maykelxyz:
# WIP: Wallet attribute: add `securityBestPractices` feature schema

Adds the data schema for evaluating wallet security best practices. Opening this draft PR for early reviews.

## SecurityBestPracticesSupport

- `KeyStorageMechanism` - how the private key is stored on-device
- `SecureRngSource` - entropy source used during key generation
- `BrowserExtensionHardening` - manifest-level security posture
- `MobileAppHardening` - OS permission hygiene and hardware-backed key storage

## Things to clarify

**1. `KeyStorageMechanism` and `SecureRngSource` has no failing variants**
Looking at the mechanisms available, would be a `PASS` or `PARTIAL`. My idea is to either wrap in `Support<>` or add an `INSECURE_RNG` variant that evaluates to `FAIL`.

**2. `SecureRngSource` has no failing variants**
- Same as above. Might be worth adding `INSECURE_CUSTOM` for this.

**3. `minimalPermissions` to be more specific**
- Currently what I did to make `MobileAppHardening.minimalPermissions` and `BrowserExtensionHardening.minimalPermissions` to be both booleans. Will refactor it into a list of permissions and rate based on what permissions are allowed.


src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
+import type { WithRef } from '@/schema/reference'
+
+/**
+ * How the wallet stores the user's private key.
+ */
+export enum KeyStorageMechanism {
+	/**
+	 * The key is encrypted with a user-known secret before being stored on disk.
+	 */
+	ENCRYPTED_WITH_USER_SECRET = 'ENCRYPTED_WITH_USER_SECRET',
```
polymutex: There needs to be some more granularity here to recognize the strength the KDF. I'm not sure that we want to put a specific threshold on it, but we do want something that can distinguish between "I just applied 1 round of AES256" case from the "I ran standard PBKDF2". Maybe the distinction should be around standardization, i.e. have distinct enum entries for "encrypted with user secret with standardized KDF" and one for non-standardized.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * The key is stored in plaintext, but in OS-sandboxed app storage that
+	 * other apps and processes cannot read.
+	 */
+	OS_SANDBOXED_PLAINTEXT = 'OS_SANDBOXED_PLAINTEXT',
+
+	/**
+	 * No private key is stored on the device — the wallet uses passkey-managed
+	 * smart contract accounts where signing happens through the OS passkey API.
```
polymutex: This is just one of multiple possible ways that this option would be the correct one. Please reword the comment to not give this impression.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+	 * The app declares only the OS permissions it actually requires,
+	 * without requesting broad access to contacts, location, etc.
+	 */
+	minimalPermissions: boolean
+
+	/**
+	 * The app uses Android Keystore or iOS Secure Enclave for key operations,
+	 * hardware-backing cryptographic material where available.
+	 */
+	usesKeystoreOrEnclave: boolean
```
polymutex: This is already covered by KeyStorageMechanism.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+export interface SecurityBestPracticesSupport {
+	/** How the wallet stores the user's private key. */
+	keyStorageMechanism: KeyStorageMechanism
+
+	/** The entropy source used during key generation. */
+	secureRng: SecureRngSource
+
+	/**
+	 * Browser extension hardening details.
+	 * Set to null if the wallet does not have a browser extension variant.
```
polymutex: `null` is reserved for values where data is not known, please keep it this way. Same below. Consider `WalletIntegration.browser` for example, which uses a string when used in wallets that don't have a browser version.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	usesKeystoreOrEnclave: boolean
+}
+
+/**
+ * Security best-practices data for a wallet.
+ * Identifies how keys are stored, how entropy is sourced, and whether
+ * deployment-environment hardening is applied.
+ */
+export interface SecurityBestPracticesSupport {
```
polymutex: Why have `Support` in the name of this interface?

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+	 * its functionality, avoiding overbroad `<all_urls>` or similar grants.
+	 */
+	minimalPermissions: boolean
+
+	/**
+	 * The `web_accessible_resources` manifest field is either absent or
+	 * restricted to specific origins, preventing other web pages from
+	 * loading internal extension resources.
+	 */
+	lockedDownAccessibleResources: boolean
```
polymutex: I think we should retrieve these programmatically. It'll make this automatically updated. For the purpose of feature data, this interface can be replaced by something representing the Manifest v3 spec. https://developer.chrome.com/docs/extensions/reference/manifest

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,108 @@
...
+	 * without requesting broad access to contacts, location, etc.
+	 */
+	minimalPermissions: boolean
+
+	/**
+	 * The app uses Android Keystore or iOS Secure Enclave for key operations,
+	 * hardware-backing cryptographic material where available.
+	 */
+	usesKeystoreOrEnclave: boolean
+}
```
polymutex: Similarly, let's just have the `AndroidManifest.xml` fields as schema here. Don't need to encode all the fields but all the ones we may want to care about, like the permissions set (`<uses-permission>`). We can expand it later at no cost since we'd be able to re-fetch the manifest on demand. Same for iOS apps (`Info.plist` → 'NS*UsageDescription's).
This won't always be possible for non-source-visible apps, and that's OK, we can fail those since that would not affect their stage rating and having no visible source is also a terrible security practice.

src/schema/features.ts
```diff
@@ -103,6 +104,9 @@ export interface WalletBaseFeatures {
 
 		/** How are secret keys handled? */
 		keysHandling: VariantFeature<WithRef<KeysHandlingSupport>>
+
+		/** Security best practices (key storage, RNG, hardening). */
+		securityBestPractices: VariantFeature<SecurityBestPracticesData>
```
polymutex: This should not be a VariantFeature since it has variant-specific fields. It would end up looking like this:

```typescript
securityBestPractices: {
  [Variant.BROWSER]: {
    // ...
    browserExtensionHardening: // ...
	mobileAppHardening: // Useless here
  },
  [Variant.MOBILE]: {
    // ...
    browserExtensionHardening: // Repeated and useless here
	mobileAppHardening: // Repeated here
  },
}
```

Consider instead making `SecurityBestPracticesData` having variant-specific subfields. I'm not even sure that there would be non-variant-specific behavior here, because things like RNG access and OS storage are OS-specific by definition, so the convenience optional "automatic merging" across variants that VariantFeature provides is useless for this set of fields.

polymutex (changes requested):
Good thing to keep this PR scoped to feature fields only.

src/schema/features/security/security-best-practices.ts
maykelxyz: Kept it simple and straightforward:

```ts
export enum KeyStorageMechanism {
	ENCRYPTED_WITH_USER_SECRET_STANDARDIZED_KDF = 'ENCRYPTED_WITH_USER_SECRET_STANDARDIZED_KDF',
	ENCRYPTED_WITH_USER_SECRET_WEAK_KDF = 'ENCRYPTED_WITH_USER_SECRET_WEAK_KDF'
}
```

src/schema/features/security/security-best-practices.ts
maykelxyz: These are the fields I added:

- **`hostPermissions`** — what pages the background service worker can access.
- **`contentScripts`** — what pages the wallet auto-injects into.
- **`externallyConnectable`** — who can send messages directly to the wallet.

src/schema/features/security/security-best-practices.ts
maykelxyz: Added field for MobileAppHardening:
- **`usesPermissions` (Android)** / **`usageDescriptions` (iOS)** — OS-level permissions the app declares.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Scope of a Chrome Extension URL match pattern grant, expressed in terms
+ * relevant to wallet extensions. Ordered from least to most permissive.
+ */
+export enum HostPermissionScope {
+	/** No host permissions declared; the extension does not access web pages. */
+	NONE = 'NONE',
+
+	/**
+	 * HTTPS origins only — covers all legitimate dApps without touching
```
polymutex: apps

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,260 @@
...
+	 * insecure pages, local files, or WebSocket connections.
+	 */
+	HTTPS_ONLY = 'HTTPS_ONLY',
+
+	/** HTTP and HTTPS origins — includes insecure web pages in addition to dApps. */
```
polymutex: apps

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,260 @@
...
+
+	/** Any installed extension may connect (`"ids": ["*"]`). */
+	ANY = 'ANY',
+}
+
+/**
+ * Security-relevant fields from the Chrome Extension Manifest.
+ * Values should be extracted directly from the published manifest.json.
+ */
+export interface BrowserExtensionManifest {
```
polymutex, line 203: We'll want `web_accessible_resources` and `permissions`. See https://developer.chrome.com/docs/extensions/develop/concepts/declare-permissions for the permissions enum.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,260 @@
...
+export interface SecurityBestPracticesData {
+	/** Browser extension variant. Set to 'NOT_A_BROWSER_EXTENSION' if absent. */
+	browser: WithRef<BrowserSecurityBestPractices> | 'NOT_A_BROWSER_EXTENSION'
+
+	/** Mobile app variant. Set to 'NOT_A_MOBILE_APP' if absent. */
+	mobile: WithRef<MobileSecurityBestPractices> | 'NOT_A_MOBILE_APP'
+
+	/** Desktop app variant. Set to 'NOT_A_DESKTOP_APP' if absent. */
+	desktop: WithRef<SecurityBestPracticesBase> | 'NOT_A_DESKTOP_APP'
+}
```
polymutex, line 382: Perhaps in a separate PR, but please add a unit test that verifies that the use of these strings (`NOT_A_BROWSER_EXTENSION`, `NOT_A_MOBILE_APP` etc) is coherent with the wallet's `variants` fields.

polymutex (approved):

src/schema/features/security/security-best-practices.ts
maykelxyz: Filed #597

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,382 @@
...
+
+	/**
+	 * All origins including non-web schemes (file://, ws://, wss://, etc.) —
+	 * the extension can read and modify local files and raw socket traffic.
+	 */
+	UNRESTRICTED = 'UNRESTRICTED',
+}
+
+/**
+ * Security-sensitive Chrome Extension permission strings declared in the
```
polymutex: "Browser" not "Chrome" to avoid putting vendor-specific names in our codebase. Manifest v3 works across browsers so it's not a chrome-specific enum.

src/schema/features/security/security-best-practices.ts
```diff
@@ -0,0 +1,382 @@
...
+ * `permissions` manifest field. Values match the manifest string exactly.
+ */
+export enum ChromeExtensionPermission {
```
polymutex: BrowserExtensionPermissions

polymutex (approved):
Looks good but please follow up with a PR shortly afterwards to do automated population for the manifests, at least starting with browser extensions, to avoid wallet development teams coming in and populating this manually

PR #588, opened 2026-03-14
GridPlus updates

ren2140eth:


ren2140eth:
Must use `redtodo` as the gridplus.io domain is blocking automated requests from the CI environment

polymutex:
> Must use `redtodo` as the gridplus.io domain is blocking automated requests from the CI environment

Can you add `gridplus.io` to the `URLS_TO_SKIP` list in `tests/url-check.test.ts`?
After that you should be able to add the gridplus URL instead of the `refTodo` and it should work.

ren2140eth:
should be all set here

polymutex (approved):
Perfect, thanks 👍

PR #593, opened 2026-03-16
Feat: Delegated spending control attribute

### LINKED ISSUE #579: Wallet attribute: Permissions management
Taken from [1TS security benchmark](https://github.com/walletbeat/walletbeat/discussions/407), a wallet attribute that encompasses the following:
- ERC-20 outstanding approvals management: Can the user easily control and revoke ERC-20 approvals?
- Spending rate-limits (for smart accounts): Is it possible to set different spending policies for different levels of spend, so that large transactions require timelocks or more signatures? (This helps with duress resistance but is a good practice even outside of such contexts)
- Preapproved transaction revocation, e.g. being able to cancel preapproved subscription payments.

The first version of this attribute should be very simple and just focus on the ERC-20 approvals part of it to keep things simple and for launch. The rest can be added on later as Account Abstraction matures.

maykelxyz:
# Delegated Spending Control
Background: #570 & #407 

## Why

ERC-20 approvals grant other addresses, contracts or accounts, indefinite permission to spend tokens on a user's behalf. Without visibility and management, users have no way to manage their token approvals.

## Changes
Created new `DelegatedSpendingControl.erc20Approvals` which asks two questions:

```ts
export interface Erc20ApprovalsControl {
	/**
	 * Can the user inspect existing token approvals through the wallet?
	 */
	canInspectTokenApprovals: boolean

	/**
	 * Can the user revoke an existing token approval through the wallet?
	 */
	canRevokeTokenApprovals: boolean
}

```

## Rating

| Rating | Condition |
|--------|-----------|
| **PASS** | Can inspect and revoke |
| **PARTIAL** | Can inspect but not revoke |
| **FAIL** | No approval management |

Inspect-only is PARTIAL, not PASS. Awareness without control is not enough.

## Future Considerations

Add spending policies and pre authorized payment revocations post - launch. We can do this on another issue to move forward. Here's my suggestion: 

```ts
spendingPolicies: Support | 'SMART_ACCOUNT_NOT_SUPPORTED'
preAuthorizedPaymentRevocation: Support
```

Closes #579


polymutex:
> Background: #570 & #407

I'm not seeing the relation with #570. Did you mean #579?

data/software-wallets/ambire.ts
polymutex, line 1: As always, please keep Ambire fully populated.

src/schema/features/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * In v1, only ERC-20 approvals may be rated. Additional mechanisms such as
+ * smart-account spending limits and revocable pre-authorized payments can be
+ * added over time.
+ */
+export interface DelegatedSpendingControl {
+	/**
+	 * ERC-20 token approvals granted to contracts.
+	 */
+	erc20Approvals: Erc20ApprovalsControl
```
polymutex: Maybe we should add ERC-721 and ERC-1155 too...

maykelxyz:
> I'm not seeing the relation with #570. Did you mean #579?

Ah, yeah. I meant #579

src/schema/features/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * 	TODO: Add spending policies and pre authorized payment revocations post - launch
+	 * 	spendingPolicies: Support | 'SMART_ACCOUNT_NOT_SUPPORTED'
+	 *  preAuthorizedPaymentRevocation: Support
+	 */
```
polymutex: We can figure this out later, the code isn't the best place to put these future plans. Remove

src/schema/features/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface Erc20ApprovalsControl {
+	/**
+	 * Can the user inspect existing token approvals through the wallet?
+	 */
+	canInspectTokenApprovals: boolean
+
+	/**
+	 * Can the user revoke an existing token approval through the wallet?
+	 */
+	canRevokeTokenApprovals: boolean
```
polymutex: This allows the combination of `canInspectTokenApprovals: false && canRevokeTokenApprovals: true`, which seems impossible.
Consider instead a 3-state enum like

```typescript
export enum SpendingApprovalsControl {
  CANNOT_INSPECT,
  CAN_INSPECT_BUT_NOT_REVOKE,
  CAN_INSPECT_AND_REVOKE,
}
```

This would also be easily reusable for ERC-721 and ERC-1155.

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should add the ability to view and revoke ERC-20 token approvals.',
+		),
+	})
+}
+
+export const delegatedSpendingControl: Attribute<DelegatedSpendingControlValue> = {
+	id: 'delegatedSpendingControl',
+	icon: '\u{1f511}', // Key
+	displayName: 'Spending control',
```
polymutex: IMO the name on #579 ("Permissions management") is a better long-term name. Yes right now that is just about token approvals permissions, but the intent is to expand beyond that.
"Delegation" is a good word too, but IMO "Permission" is more widely understood and less jargon-y.

"Permissions control" would work perhaps, I still prefer "Permissions management" over it but that's a weak preference at that point.

OK to keep the feature field as more "close to the ground" as they currently are (and they're easier to rename); this comment is specifically about the _attribute_ name. Expanding feature types is easy, but it's better to lessen the surprise of expanding the scope of an attribute by making its scope upfront broader.

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+	wording: {
+		midSentenceName: 'delegated spending control',
+	},
+	question: sentence(
+		'Can you inspect and revoke ERC-20 token approvals granted to other addresses, directly from within {{WALLET_NAME}}?',
```
polymutex: Similarly, the question should be a bit broader here. For example: `Does {{WALLET_NAME}} let you inspect and manage the spending permissions you've granted to other contracts and apps?` (Could also remove "spending" in there as well optionally, or keep it for clarity)

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+
+import { pickWorstRating, unrated } from '../common'
+
+export type DelegatedSpendingControlValue = Value
+
+function evaluate(
+	ctx: EvaluationContext<DelegatedSpendingControlValue>,
+	control: Support<DelegatedSpendingControlSupport>,
+): Evaluation<DelegatedSpendingControlValue> {
+	if (!isSupported<DelegatedSpendingControlSupport>(control)) {
```
polymutex: The type parameter should not be needed here.

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+		Being able to inspect and revoke approvals is an important tool for protecting
+		your assets from unnecessary or dangerous delegated spending authority.
+	`),
+	methodology: markdown(`
+		Wallets are rated based on whether they allow users to inspect existing ERC-20
+		token approvals and revoke them directly from within the wallet interface.
+
+		* **Pass**: The wallet lets users both inspect and revoke ERC-20 approvals.
+		* **Partial**: The wallet lets users inspect approvals but not revoke them.
+		* **Fail**: The wallet provides no approval inspection or revocation functionality.
```
polymutex: This is already explained by the rating scale. Delete

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+		permission to spend tokens on your behalf.
+		Malicious or compromised contracts with existing approvals can drain your wallet,
+		and approvals to other accounts carry the same risk.
+
+		Being able to inspect and revoke approvals is an important tool for protecting
+		your assets from unnecessary or dangerous delegated spending authority.
+	`),
+	methodology: markdown(`
+		Wallets are rated based on whether they allow users to inspect existing ERC-20
+		token approvals and revoke them directly from within the wallet interface.
```
polymutex: Add something about future plans for this attribute here, to lessen surprise when it does. Something like "As Account Abstraction becomes more prevalent, this methodology will also grow to encompass the management of more complex account permissions."

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
```diff
@@ -0,0 +1,175 @@
...
+		),
+		howToImprove: paragraph(
+			'{{WALLET_NAME}} should add the ability to view and revoke ERC-20 token approvals.',
+		),
+	})
+}
+
+export const delegatedSpendingControl: Attribute<DelegatedSpendingControlValue> = {
+	id: 'delegatedSpendingControl',
+	icon: '\u{1f511}', // Key
```
polymutex: Feel like there may be a better option here. Maybe it doesn't matter since we're so close to having icons. But should poke Teresa to actually make one!

polymutex (changes requested):
Note that our timeline had @johngrantuk actually work on this one. Hopefully he hasn't started so that there's no duplicate work here :)

src/schema/features/self-sovereignty/delegated-spending-control.ts
maykelxyz:
```ts
permissionsManagement: supported({
ref: refTodo,
erc1155Approvals: SpendingApprovalsControl.CAN_INSPECT_AND_REVOKE,
erc20Approvals: SpendingApprovalsControl.CAN_INSPECT_AND_REVOKE,
erc721Approvals: SpendingApprovalsControl.CAN_INSPECT_AND_REVOKE,
}),
```

src/schema/attributes/self-sovereignty/delegated-spending-control.ts
maykelxyz: Hmm, I think it would be better to make BOTH attribute and feature `Permissions Management` and rename it down the line.

polymutex: Sure, that's fine too

polymutex (approved):
Looks good, just remember to populate Ambire.

PR #594, opened 2026-03-17
Chore: Add a fileverse version of the software wallets questionnaire

maykelxyz:


polymutex (commented):
Not sure I understand the purpose of this PR. This file is meant to be the template that we can use to import into Fileverse when making copies for each wallet team we send this out to. So adding a link here would mean all the copies kind of link to another central doc... Am I missing something?

maykelxyz:
I was having a conversation with @ren2140eth and he mentioned he was having trouble with importing the MD to Fileverse

just created the fileverse version in case someone's also having the same issues. It'd be easier to copy paste or duplicate from Fileverse

polymutex:
OK. Then maybe we should instead have a folder rather than a top-level file (probably best anyway for this to not be at the top level), and then split this:

- `resources/docs/questionnaire/software-wallet-questionnaire.md`: The current markdown file
- `resources/docs/questionnaire/README.md`: An explanation of what the questionnaire is and a link to the Fileverse version

resources/docs/questionnaire/software-wallet-questionnaire.md
```diff
@@ -194,3 +194,5 @@ Thank you for taking the time to fill this out. Your answers help us rate your w
 ## Anything Else you want to specifically mention?
 
 Is there anything about your wallet's security, privacy, or self-sovereignty features that you'd like us to know, or that the questions above didn't cover?
+
+_Alternatively, here's a link to this questionnaire on Fileverse [Link here](https://docs.fileverse.io/0xbf80c3d83c7f1ed9938e677fae318bc5d950a2a3/4#key=bEht60AIHK7mSSOH9aY-BKU1I-RX2bCYYdR0IgUbWT8exByOvkctcHfTA7apE5aP)_
```
polymutex: Now that this link exists in the README, it can be deleted from here.

polymutex (approved):

PR #601, opened 2026-03-18
Add general privacy hygiene attribute

### LINKED ISSUE #559: Wallet attribute: User privacy
See thread here: https://x.com/PryvitKyle/status/2028550003577516121

We should have an attribute that captures whether the wallet respect the user's privacy. Things like:

- Not leaking your browsing history
- Asking for consent before using any analytics (Matomo/Sentry/etc)

Essentially "don't send any data that the user hasn't agreed to being sent".

Pre-launch because this is part of software methodology, which we need to finalize for launch. Thankfully this should not require much new feature data since the bulk of it is already captured in `privacy.dataCollection`, but we do need to add a "does the wallet ask for user consent for analytics" feature type too.

johngrantuk:
Closes #559 

* Add analyticsConsent privacy feature
  * Configure Ambire as NEVER:
	```
	Ambire privacy policy (ambire.com/privacy-policy): States the wallet does not require or store cookies (except Cloudflare on the hosted version), does not store identifiable personal data for normal use, and that non-personal data is only optionally stored on a backend. It does not mention in‑wallet analytics, crash reporting, or telemetry.
	GitHub (AmbireTech/wallet): A code search for “sentry”, “analytics”, or “telemetry” returned no matches in the wallet repo.
	```
* Add new attribute: General privacy hygiene
  * What it evaluates:
    * Fail: Sending browsing history or wallet-connected domains to an external service by default without consent, or using analytics (e.g. Matomo, Sentry) without prior user consent.
    * Pass: No such data sent by default and analytics only used after consent (e.g. opt-in, prompted, or never).
  * Inputs: Existing feature data for privacy data collection and analytics consent
  * Exemption: Hardware wallets are exempt (attribute not applicable).
  * Scale: Fail–pass only (no partial); aggregation takes the worst rating across variants.
	* The new attribute is wired into the Privacy attribute group.
* Add General Privacy Hygiene to software Stage 1
  * baseline expectation that the wallet doesn’t send sensitive data or enable analytics without user consent.


data/software-wallets/completed.tmpl.ts
```diff
@@ -347,6 +346,7 @@ export const completedTemplate: SoftwareWallet = {
 		},
 		multiAddress: featureSupported,
 		privacy: {
+			analyticsConsent: null,
```
polymutex: Please keep this file fully-populated as well, it's meant to be a "fully passing" fictitious wallet.

src/schema/stages/software-wallet-stages.ts
polymutex, line 1: Please keep this new attribute out of the ladder for now. The whole ladder is due for a refresh anyway, which we can do in one shot; let's keep this PR to just adding the attribute.

src/schema/features.ts
```diff
@@ -119,6 +118,13 @@ export interface WalletBaseFeatures {
 
 		/** Transaction privacy features. */
 		transactionPrivacy: VariantFeature<TransactionPrivacy>
+
+		/**
+		 * Collection policy for analytics and telemetry (product analytics and
+		 * crash/error reporting, e.g. Matomo, Sentry). NEVER = no analytics;
+		 * OPT_IN / PROMPTED = consent before first use; BY_DEFAULT / ALWAYS = no consent.
+		 */
+		analyticsConsent: VariantFeature<CollectionPolicy | null>
```
polymutex: `| null` is already implied by `VariantFeature`, so no need to put it here.

src/schema/features.ts
```diff
@@ -119,6 +118,13 @@ export interface WalletBaseFeatures {
 
 		/** Transaction privacy features. */
 		transactionPrivacy: VariantFeature<TransactionPrivacy>
+
+		/**
+		 * Collection policy for analytics and telemetry (product analytics and
+		 * crash/error reporting, e.g. Matomo, Sentry). NEVER = no analytics;
+		 * OPT_IN / PROMPTED = consent before first use; BY_DEFAULT / ALWAYS = no consent.
+		 */
```
polymutex, line 135: I think it would be worthwhile to separate the two main types of analytics (crash reporting and general UI tracking). This has three advantages:

Forces both questions to be asked as part of the wallet data entry process (otherwise someone filling this in may assume this is just about the general UI tracking)
The wallet may choose to have a different policy for each, e.g. ask for user consent for UI tracking but unconditionally send crash reports, and it's useful to be able to capture that distinction.
We can apply different policies for both. For example, maybe it's a PARTIAL rating if the wallet has no usage tracking and keeps the crash reporting as BY_DEFAULT, since the user can presumably disable it before the wallet first crashes.

polymutex, line 135: Another aspect that may be useful to capture here is the entity collecting the data. There's a meaningful distinction between a wallet capturing crash reports for themselves vs a wallet selling every click to whatever "web3 ads" "attribution tracking" engine.

Something like:

```typescript
interface WalletAnalytics {
  /** Where the analytics data goes. */
  entity: Entity

  /**
   * Analytics data collection policy.
   * Cannot be "NEVER"; use a `NO_*_ANALYTICS` string in the parent field instead when the wallet doesn't do this type of analytics.
   */
  policy: Exclude<CollectionPolicy, CollectionPolicy.NEVER>
}


//... and in the `features.privacy` block:
analytics: {
  usageAnalytics: WithRef<WalletAnalytics> | 'NO_USAGE_ANALYTICS'
  crashReports: WithRef<WalletAnalytics> | 'NO_CRASH_REPORT_ANALYTICS'
},
```

Maybe there's a better way (using `Support<...>` perhaps?) but the idea being to populate the `Entity` field only if the wallet actually has such analytics.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+} from '@/schema/features/privacy/data-collection'
+import type { AtLeastOneVariant } from '@/schema/variants'
+import { verifiabilityRequiresSourceCodeAccess } from '@/schema/verifiability'
+import type { WalletMetadata } from '@/schema/wallet'
+import { WalletType } from '@/schema/wallet-types'
+import { markdown, paragraph, sentence } from '@/types/content'
+
+import { exempt, pickWorstRating, unrated } from '../common'
+
+/** If this list is expanded, the attribute's methodology text must be updated to name the new data types. */
```
polymutex: In order to ensure this is always done and can't be forgotten, please introduce a `switch`-y function similar to `userInfoName` in `src/schema/features/privacy/data-collection.ts` but that returns a boolean. The compiler will ensure that it must have a branch for every value in `UserInfo`.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+	PersonalInfo.BROWSING_HISTORY_URLS,
+	WalletInfo.WALLET_CONNECTED_DOMAINS,
+]
+
+export type GeneralPrivacyHygieneValue = Value
+
+function doesNotCollectAnalyticsWithoutConsent(policy: CollectionPolicy | null): boolean {
+	if (policy === null) {
+		return false
+	}
```
polymutex: This is not a valid assumption, and having this logic here may lead to a misleading rating. I think it is safer to remove the `| null` from the function signature, as this function is never called when `policy` is `null` anyway, and then this `if` can be removed too.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+		details: sentence(
+			'{{WALLET_NAME}} does not send browsing history or wallet-connected domains without consent, and asks for consent before using analytics.',
+		),
+	})
+}
+
+export const generalPrivacyHygiene: Attribute<GeneralPrivacyHygieneValue> = {
+	id: 'generalPrivacyHygiene',
+	icon: '\u{1f50f}', // lock
+	displayName: 'General privacy hygiene',
```
polymutex: I kind of like the "User privacy" name but in retrospect it's perhaps not distinguishable enough from the other privacy attribute names. "General privacy hygiene" is better in that regard, but also feels overly wordy.

Maybe just "Privacy hygiene", or maybe "User analytics"?

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+		},
+		details: sentence(
+			'{{WALLET_NAME}} does not send browsing history or wallet-connected domains without consent, and asks for consent before using analytics.',
+		),
+	})
+}
+
+export const generalPrivacyHygiene: Attribute<GeneralPrivacyHygieneValue> = {
+	id: 'generalPrivacyHygiene',
+	icon: '\u{1f50f}', // lock
```
polymutex: Too generic of an icon. Perhaps a  soap (🧼) for hygiene?

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+	displayName: 'General privacy hygiene',
+	wording: {
+		midSentenceName: 'general privacy hygiene',
+	},
+	question: sentence(
+		'Does {{WALLET_NAME}} avoid sending data that should never be shared without your consent?',
```
polymutex: Triple negative is confusing ("avoid" + "should never" + "without"). Consider:

`'Does {{WALLET_NAME}} only send sensitive data with your explicit consent?'`

Or maybe we can leave it as a single negative (and imply that the better answer is "no"): `'Does {{WALLET_NAME}} send sensitive data without your explicit consent?'`

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+	),
+	why: markdown(
+		'Users expect that data like browsing history and which sites they connect their wallet to is never sent without consent. Analytics (Matomo, Sentry, etc.) should only be used after the user has agreed. This attribute encodes that baseline.',
```
polymutex: Maybe add an analogy to browsers here. Something like:
"Much like users wouldn't expect their web browsers to leak their browsing history for analytics, they wouldn't expect their wallets to track every click they make."

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+	),
+	methodology: markdown(
+		'We consider default behavior; we look at network requests and data-collection policies; we fail if browsing history or wallet-connected domains are sent by default, or if analytics are used without prior consent.',
```
polymutex: Distinguish crash report analytics and user tracking analytics here.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+			),
+			exampleRating(
+				sentence(
+					'The wallet uses analytics (e.g. Matomo or Sentry) without asking for user consent first.',
+				),
+				analyticsWithoutConsent(EvaluationContext.forTest(() => generalPrivacyHygiene)),
+			),
+		],
+		partial: [],
+		pass: exampleRating(
```
polymutex: There's an important other `pass` case here, which is a wallet that simply doesn't track the user.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+function doesNotCollectAnalyticsWithoutConsent(policy: CollectionPolicy | null): boolean {
+	if (policy === null) {
+		return false
+	}
+
+	return (
+		policy === CollectionPolicy.NEVER ||
+		policy === CollectionPolicy.OPT_IN ||
+		policy === CollectionPolicy.PROMPTED
+	)
```
polymutex: You can use the `collectedByDefault` function from `src/schema/features/privacy/data-collection.ts` instead. (Maybe this whole function can just be replaced by that)

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+function forbiddenDataByDefault(
+	ctx: EvaluationContext<GeneralPrivacyHygieneValue>,
+): Evaluation<GeneralPrivacyHygieneValue> {
+	return ctx.build({
+		value: {
+			id: 'forbidden_data_by_default',
+			rating: Rating.FAIL,
+			displayName: 'Browsing history or wallet-connected domains sent by default',
```
polymutex: Please add distinct ratings between these cases; otherwise a wallet that only does one of the two may perceive this phrasing as an unnecessary accusation.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+function analyticsWithoutConsent(
+	ctx: EvaluationContext<GeneralPrivacyHygieneValue>,
+): Evaluation<GeneralPrivacyHygieneValue> {
+	return ctx.build({
+		value: {
+			id: 'analytics_without_consent',
+			rating: Rating.FAIL,
+			displayName: 'Analytics without prior consent',
+			shortExplanation: sentence(
+				'{{WALLET_NAME}} uses analytics (such as Matomo or Sentry) without asking for user consent first.',
```
polymutex: With the `Entity` in the feature field, wording like "Matomo or Sentry" can be customized to the wallet's actual case.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+		const allDataCollection = dataCollectionForAllSupportedFlows(dataCollection)
+
+		if (dataCollection === null || allDataCollection === null) {
+			return unrated(ctx, null)
+		}
+
+		let hasAnalyticsInSomeFlow = false
+
+		for (const collected of allDataCollection) {
+			ctx.addRef(collected)
```
polymutex: Should only call `addRef` if `collected.purposes.includes(DataCollectionPurpose.ANALYTICS)` or if the data collected is part of the forbidden data set.

src/schema/attributes/privacy/general-privacy-hygiene.ts
```diff
@@ -0,0 +1,207 @@
...
+			hasAnalyticsInSomeFlow =
+				hasAnalyticsInSomeFlow || collected.purposes.includes(DataCollectionPurpose.ANALYTICS)
+
+			const qualified = qualifiedDataCollection(collected.dataCollection)
+
+			for (const info of DATA_NEVER_WITHOUT_CONSENT) {
+				if (collectedByDefault(qualified[info])) {
+					return forbiddenDataByDefault(ctx)
```
polymutex: Return at the end of the `allDataCollection` loop to make sure we catch all the relevant `ref`s, not just the first one.

polymutex (changes requested):
Thanks for taking this on!

data/software-wallets/ambire.ts
```diff
@@ -394,6 +394,7 @@ export const ambire: SoftwareWallet = {
 		},
 		multiAddress: featureSupported,
 		privacy: {
+			analyticsConsent: CollectionPolicy.NEVER,
```
polymutex: Are you sure about this btw? https://github.com/search?q=repo%3AAmbireTech%2Fextension%20sentry&type=code suggests they may be using Sentry.

src/schema/features.ts
johngrantuk: Now it uses `notSupported` if a type of analytics is not used by the wallet, e.g.:
```
analytics: {
  crashReports: notSupported,
  usage: notSupported,
}
```
Not sure if that feels a bit weird so let me know if there's a better alternative.

data/software-wallets/ambire.ts
johngrantuk: I've updated this now. I added a new Sentry entity but not sure if thats the correct approach. The reasoning is - the crash telemetry is sent via Sentry SDK so the actual external crash-report collector/processor is Sentry, not Ambire itself.
Also not 100%R sure about `usage` analytics. I can't see anything obvious but would be good to have another set of eyes to confirm.

src/schema/features.ts
polymutex: Yeah, "support" feels like the wrong word here, but at least in terms of the type structure it makes sense. Let's keep it

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -0,0 +1,557 @@
...
+	return entities.filter(e => {
+		if (seen.has(e.id)) {
+			return false
+		}
+
+		seen.add(e.id)
+
+		return true
+	})
+}
```
polymutex: Feel free to move these helper functions somewhere more generic. For example next to the other helper functions in `src/schema/entity.ts`.

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -0,0 +1,557 @@
...
+
+function noForbiddenDataByDefault(
+	ctx: EvaluationContext<PrivacyHygieneValue>,
+): Evaluation<PrivacyHygieneValue> {
+	return ctx.build({
+		value: {
+			id: 'no_forbidden_data_by_default',
+			rating: Rating.PASS,
+			displayName: 'No forbidden data sent by default',
```
polymutex: This `displayName` has to make sense on its own even when seen out of immediate context of the attribute name (e.g. on tooltip in the combined chart). The phrase "forbidden data" here is ambiguous (and also a bit scary).
I think this can be something like "Asks for analytics consent" to keep it short. Still short, and maybe only covers both the analytics part, but the part about forbidden data being off by default seems like a given when the wallet is already careful enough to ask for consent for more benign data (by comparison) like usage analytics.

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -0,0 +1,557 @@
...
+			shortExplanation: sentence(
+				'{{WALLET_NAME}} does not send browsing history or wallet-connected domains without consent, and asks for consent before using product analytics or crash/error reporting.',
```
polymutex: "wallet-connected domains" is already implied by "browsing history", since you cannot connect to a domain without first browsing to it. So I think you can remove the "wallet-connected domains" part here and everywhere where browsing history is already mentioned.

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -0,0 +1,557 @@
...
+	const crashReportsAnalytics = ctx.features.privacy.analytics.crashReports
+	const evaluations: Array<Evaluation<PrivacyHygieneValue>> = []
+
+	if (usageAnalytics !== null && isSupported(usageAnalytics)) {
+		ctx.addRef(usageAnalytics)
+
+		if (
+			usageAnalytics.policy === CollectionPolicy.BY_DEFAULT ||
+			usageAnalytics.policy === CollectionPolicy.ALWAYS
+		) {
```
polymutex: Use `collectedByDefault`

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -0,0 +1,557 @@
...
+					}
+
+					if (
+						info === WalletInfo.WALLET_CONNECTED_DOMAINS &&
+						!walletConnectedDomainsByDefaultEntities.some(e => e.id === collected.byEntity.id)
+					) {
+						walletConnectedDomainsByDefaultEntities.push(collected.byEntity)
+					}
+				}
+			}
```
polymutex, line 374: The intent behind having `isForbiddenWithoutPriorConsentUserInfo` is enforcing that the typechecker warns in case we add a new type of user info to ensure that it is classified in this attribute's logic as well. In the same spirit, this code also has the potential to be forgotten to be updated in such a pass, which we can avoid. I think this can be added by structuring it as follows:

```typescript
for (const info of forbiddenInfos) {
  if (!collectedByDefault(qualifiedCollection[info])) {
    continue
  }
  switch (info) {
    case PersonalInfo.BROWSING_HISTORY_URLS:
      if (!browsingHistoryByDefaultEntities.some(e => e.id === collected.byEntity.id) {
        browsingHistoryByDefaultEntities.push(collected.byEntity)
      }
      break;
    case PersonalInfo.WALLET_CONNECTED_DOMAINS:
      // ...
      break;
    default:
      throw new Error(`unexpected forbidden info ${info}`)
  }
}
```

This way the code will throw an error if it's not also updated.

polymutex (approved):
Looks great, just some code structure comments. Feel free to merge after this

Issue #519, opened 2026-02-06
Add test enforcing commit hashes in `ref` links to GitHub


polymutex:
`ref` links to GitHub are common and very valuable, but by default the GitHub URL has the branch name in it, rather than the commit hash. This means the file can be changed in the future, line numbers may become wrong, or the file may get deleted/renamed and the link breaks. Additionally, commit hashes are a way to encode the time the review was done, as any new commit that has touched the file since is now clearly a change since the review was done and this is now easy to tell.

We should enforce that GitHub links in `ref`s have explicit commit hashes, i.e. the format `https://github.com/[username]/[reponame]/blob/[commithash]/[path]`.

pheobeayo:
Can I take up this issue?

maykelxyz:
I think no one's still done this, so yes :)

pheobeayo:
@maykelxyz On it

PR #603, opened 2026-03-20
Attribute: security best practices attribute

### LINKED ISSUE #218: Wallet attribute: Security best practices
Great idea by @PatrickAlphaC on the wallet stage system issue #173:

> Key is stored in encrypted form

This seems like an important security attribute, for which we should have feature data. Currently we don't have anything.

I'd generalize this attribute to "is the private key securely accessed?", as this is a less-implementation-prescriptive way to express the same idea. Namely, this generalizes to also allow:

- Keys stored in hardware enclaves only that can't get out of it, without the wallet software having to implement encryption itself. Example: a mobile wallet that stores keys in enclaves.
- Keys stored in plaintext in confines of the filesystem that can't be accessed by other applications, as enforced by the operating system. For example, I believe Android apps' application-internal storage is confined to that application, and such enforcement is done by sandboxing other applications away from having access to these files. So the keys are safe from being read by any other application, even if the keys are stored in plaintext.
- Wallets that don't store keys at all. Example: Smart wallets for which the account is secured by a passkey managed outside of the wallet, and for which the passkey verification is done onchain (so not in the wallet software itself).

maykelxyz:
# Security Best Practices Attribute

###  `securityBestPractices` Attribute

Adds a new `securityBestPractices` attribute that evaluates whether a wallet follows security best practices across up to four independently scored sub-evaluations. The worst sub-evaluation rating determines the overall score.

---

##### Key Storage
Checks how private keys are stored on-device.
- **Pass**: Keys stored in a hardware security module/secure enclave, or no key stored at all (passkey-managed accounts).
- **Partial**: Keys encrypted with a user secret via standardized KDF, or stored in OS-sandboxed plaintext (no additional encryption at rest).
- **Fail**: Keys encrypted with a weak/non-standard KDF, or unrecognized/insecure storage mechanism.

##### Secure RNG
Checks the entropy source used for key generation.
- **Pass**: OS CSPRNG or dedicated hardware entropy source.
- **Partial**: Library-provided RNG whose cryptographic quality is unverified.
- **Fail**: Unknown or insecure entropy source.

##### Deployment Hardening
Split into browser extension and mobile app checks.

- **Browser extension**: evaluates host permissions scope, `web_accessible_resources`, content script injection scope, `externally_connectable` config, and declared browser API permissions (e.g. `debugger`, `proxy`, `nativeMessaging`).
- **Mobile app**: checks for unnecessary Android (`SYSTEM_ALERT_WINDOW`, `RECORD_AUDIO`, etc.) and iOS (`NSMicrophoneUsageDescription`, `NSPhotoLibraryUsageDescription`, etc.) permission declarations.

Ratings:
- **Pass**: Minimal permissions, restricted resources, no dangerous API access.
- **Partial**: Some elevated-risk permissions or partial hardening gaps.
- **Fail**: Overbroad permissions, unrestricted resources, or dangerous browser API permissions declared.

##### Passkey Verification *(only evaluated if the wallet supports passkey verification)*
Checks the on-chain P-256 verification library used.
- **Pass**: Well-audited library (Smooth Crypto Lib, Daimo P256 Verifier, OpenZeppelin P256 Verifier, WebAuthn.sol).
- **Partial**: Less-audited library (Fresh Crypto Lib or other).
- **Fail**: No recognized passkey verification library used.

## Other changes
- Added words to cspell 
	- deanonymize
	- KDF
	- OS
	- overbroad
- Added `securityBestPractices` to wallets
- Added a passing `securityBestPractices` feature to completed template
- Added a `passkeyLibraryName` function to translate `PasskeyVerificationLibrary` enum to human readable name

Closes #218


src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+	BrowserExtensionPermission.NATIVE_MESSAGING,
+	// Reads clipboard contents without a user gesture.
+	BrowserExtensionPermission.CLIPBOARD_READ,
+	// Can synchronously block or modify all HTTP/S requests.
+	BrowserExtensionPermission.WEB_REQUEST_BLOCKING,
+	// Reads and modifies browser-wide privacy settings.
+	BrowserExtensionPermission.PRIVACY,
+	// Saves a full MHTML snapshot of any tab.
+	BrowserExtensionPermission.PAGE_CAPTURE,
+]
```
polymutex: How did you come up with these lists?
I am worried that some of these may be incorrect, in that an extension or an app may use them for legitimate purposes, and I am also worried about this list being non-comprehensive. Would prefer a `Record<BrowserExtensionPermission, ExplicitRating>` because that enforces completeness at the typechecker level as permissions are added/updated.

Also, looking at real extensions' manifests would be informative as to the accuracy of these lists, probably more than our own judgement would. For example, permissions like `USER_SCRIPTS` and `BROWSING_DATA` may legitimately be required for building a wallet that injects `window.ethereum`. I don't know if that's truly the case, but it's plausible. Worth verifying in practice.

polymutex (changes requested):
I would like to see a real wallet before we go further with the attribute here. This means building the pipeline for converting manifests to the `SecurityBestPracticesData` now, before building the attribute. Can be its own PR. I'd recommend following an approach like #558, basically having a checked-in data file (JSON in the data collection case, but could be any format, like XML for the Android Manifest), a TypeScript tool to automatically download a fresh manifest for a given browser extension ID or APK URL, and later down the line a cron job to do it every so often (can run by walletbeat-bot similar to the Helios checkpoint update today).

(Also this order will ensure we can keep Ambire fully-rated)

src/schema/attributes/security/security-best-practices.ts
maykelxyz: Yeah, I agree on that point. A `Record<BrowserExtensionPermission, ExplicitRating>` type could work.

The idea I have is to rate each permission as FAIL, PARTIAL, or PASS depending on how necessary it actually is. PASS = clearly needed for the wallet to function, while PARTIAL and FAIL are more on the “this might be risky / not really needed” side.

Thoughts?

src/schema/url.ts
```diff
@@ -16,6 +16,33 @@ type WithDomain<D extends string> = `${'http' | 'https'}://${D | `www.${D}`}${''
 export type DomainUrl<D extends string> = Url &
 	(WithDomain<D> | (LabeledUrl & { url: WithDomain<D> }))
 
+/**
+ * Format: https://chromewebstore.google.com/detail/[slug]/[32-char-extension-id]
```
polymutex: This seems like a narrower type than `DomainUrl<'chromewebstore.google.com'>`. Also I think it is confusing "slug" with "32-char extension id" (aren't they the same slug?)

I think you should make a separate file for the `ExtensionUrl` type, and have the type be more narrowly-defined (e.g. include the `/detail/${string}/` part and restrict to `https` only).

This should be able to extend `DomainUrl<'chromewebstore.google.com'>` as well; you can enforce this by defining it like
```typescript
export type ExtensionUrl = DomainUrl<'chromewebstore.google.com'> & `https://chromewebstore.google.com/detail/${string}`
```

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+} {
+	const known = new Map<string, AndroidPermission>(
+		Object.values(AndroidPermission).map(v => [v, v]),
+	)
+
+	const permissions: AndroidPermission[] = []
+	const unrecognized: string[] = []
+
+	// Match <uses-permission android:name="..."> (single or double quotes)
+	const pattern = /<uses-permission\s[^>]*android:name=["']([^"']+)["']/g
```
polymutex: Can we use a real XML parser? This is brittle, and a malicious wallet could format their XML differently just to avoid being detected as using a specific permission by Walletbeat.

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+	let match: RegExpExecArray | null
+
+	while ((match = pattern.exec(xmlText)) !== null) {
+		const name = match[1]
+		const known_perm = known.get(name)
```
polymutex: We've been using `lowerCamelCase` for variants in TypeScript so far so let's continue to do that.

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+} {
+	const known = new Map<string, IosUsageDescription>(
+		Object.values(IosUsageDescription).map(v => [v, v]),
+	)
+
+	const usageDescriptions: IosUsageDescription[] = []
+	const unrecognized: string[] = []
+
+	// Match <key>NS*UsageDescription</key>
+	const pattern = /<key>(NS\w+UsageDescription)<\/key>/g
```
polymutex: Same, use XML parser

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+		const known_perm = known.get(name)
+
+		if (known_perm !== undefined) {
+			if (!permissions.includes(known_perm)) {
+				permissions.push(known_perm)
+			}
+		} else {
+			if (!unrecognized.includes(name)) {
+				unrecognized.push(name)
+			}
```
polymutex: Should this just `throw` an error instead? Have it as a forcing function to categorize unknown permission names.

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Parses an AndroidManifest.xml string and returns the declared permissions.
+ *
+ * Extracts all `<uses-permission android:name="...">` entries and maps them
+ * to `AndroidPermission` enum values. Unrecognized permissions are returned
+ * separately so they can be reviewed and added to the enum if needed.
+ */
+export function parseAndroidManifest(xmlText: string): {
+	permissions: AndroidPermission[]
```
polymutex: Use `Set<AndroidPermission>` to avoid the repeated `permissions.includes` checks.

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * Parses an iOS Info.plist XML string and returns the declared usage descriptions.
+ *
+ * Extracts `NS*UsageDescription` keys and maps them to `IosUsageDescription`
+ * enum values. Unrecognized keys are returned separately.
+ */
+export function parseIosPlist(plistText: string): {
+	usageDescriptions: IosUsageDescription[]
```
polymutex: `Set<IosUsageDescription>`

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+		const known_desc = known.get(key)
+
+		if (known_desc !== undefined) {
+			if (!usageDescriptions.includes(known_desc)) {
+				usageDescriptions.push(known_desc)
+			}
+		} else {
+			if (!unrecognized.includes(key)) {
+				unrecognized.push(key)
+			}
```
polymutex: throw

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+	const usageDescriptions: IosUsageDescription[] = []
+	const unrecognized: string[] = []
+
+	// Match <key>NS*UsageDescription</key>
+	const pattern = /<key>(NS\w+UsageDescription)<\/key>/g
+	let match: RegExpExecArray | null
+
+	while ((match = pattern.exec(plistText)) !== null) {
+		const key = match[1]
+		const known_desc = known.get(key)
```
polymutex: lowerCamelCase

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+ * to `AndroidPermission` enum values. Unrecognized permissions are returned
+ * separately so they can be reviewed and added to the enum if needed.
+ */
+export function parseAndroidManifest(xmlText: string): {
+	permissions: AndroidPermission[]
+	unrecognized: string[]
+} {
+	const known = new Map<string, AndroidPermission>(
+		Object.values(AndroidPermission).map(v => [v, v]),
+	)
```
polymutex: Define an `Enum` in `schema/features/security/security-best-practices` (using `src/utils/enum.ts`) and this code is no longer necessary

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Extracts `NS*UsageDescription` keys and maps them to `IosUsageDescription`
+ * enum values. Unrecognized keys are returned separately.
+ */
+export function parseIosPlist(plistText: string): {
+	usageDescriptions: IosUsageDescription[]
+	unrecognized: string[]
+} {
+	const known = new Map<string, IosUsageDescription>(
+		Object.values(IosUsageDescription).map(v => [v, v]),
+	)
```
polymutex: Define an `Enum`

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,129 @@
...
+			if (!unrecognized.includes(key)) {
+				unrecognized.push(key)
+			}
+		}
+	}
+
+	return { usageDescriptions, unrecognized }
+}
+
+function enumKey<T extends Record<string, string>>(enumObj: T, value: string): string {
```
polymutex: You can add this as a helper function on `Enum` if you want.

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,223 @@
...
+} from './android-manifest-parser'
+import {
+	parseBrowserExtensionManifest,
+	renderBrowserExtensionManifest,
+} from './browser-ext-manifest-parser'
+import { fetchBrowserExtensionManifest, fetchText } from './crx-downloader'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const REPO_ROOT = path.resolve(__dirname, '../../..')
```
polymutex: Use the helpers in `tests/utils/codebase.ts`

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,223 @@
...
+Optional (requires --id):
+  --android-manifest-url <url>     URL to a raw AndroidManifest.xml (e.g. from
+                                   GitHub) for the wallet's mobile app.
+  --ios-plist-url <url>            URL to a raw Info.plist XML (e.g. from
+                                   GitHub) for the wallet's iOS app.
+
+Data is saved to:
+  data/manifests/<wallet-id>/<extension-id>.manifest.json
+  data/manifests/<wallet-id>/android-manifest.xml
+  data/manifests/<wallet-id>/ios-info.plist.xml
```
polymutex: Save under `data/software-wallet/manifests/<wallet-id>` since these only make sense for software wallets.

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,223 @@
...
+if (allMode && (androidManifestUrl !== undefined || iosPlistUrl !== undefined)) {
+	process.stderr.write('Error: --android-manifest-url and --ios-plist-url require --id\n')
+	usage()
+}
+
+type WalletEntry = { id: string; extensionUrls: string[] }
+
+function isWalletId(id: string): id is keyof typeof allWallets {
+	return Object.hasOwn(allWallets, id)
+}
```
polymutex: I already have a helper for this in the data collection CLI, maybe I should merge it to avoid code duplication...

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,223 @@
...
+		.filter(entry => entry.extensionUrls.length > 0)
+
+	if (targets.length === 0) {
+		process.stderr.write('No wallets with Chrome extension URLs found.\n')
+		process.exit(0)
+	}
+
+	process.stderr.write(`Found ${targets.length} wallet(s) with extension URLs.\n`)
+} else {
+	const extensionUrls = getExtensionUrls(walletId!)
```
polymutex: You can move the walletId validity check from `getExtensionUrls` to here to avoid the need for the `!` here and below. Just change the signature type of `getExtensionUrls` to take in a WalletId.

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,223 @@
...
+				'\nUnrecognized Android permissions (review; add to AndroidPermission enum if security-relevant):\n',
+			)
+
+			for (const p of unrecognized) {
+				process.stderr.write(`  ${p}\n`)
+			}
+		}
+
+		process.stderr.write('\n--- MobileAppManifest.android (paste into wallet file) ---\n\n')
+		process.stdout.write(renderAndroidPermissions(permissions) + '\n')
```
polymutex: This seems unnecessary, we already have the data we need accessible in a programmatic way. So there is no need to require the developer to manual copy-paste this. You can write to JSON in the same directory, then import as JSON in the wallet file. Check the data collection CLI thing which works the same way.
This also makes it so that if the android manifest updates, then the only thing needed is to run the tool, no need to edit the wallet files manually.

src/tools/manifest-collector/zip-reader.ts
polymutex, line 1: Can we use a pre-existing TypeScript zip library instead?

src/tools/manifest-collector/crx-downloader.ts
```diff
@@ -0,0 +1,123 @@
...
+	// The `x` parameter is a URL-encoded query string for the extension update check.
+	const innerQuery = encodeURIComponent(`id=${extensionId}&installsource=ondemand&uc`)
+	const url =
+		'https://clients2.google.com/service/update2/crx' +
+		'?response=redirect' +
+		'&prodversion=120.0' +
+		'&acceptformat=crx3' +
+		`&x=${innerQuery}`
+
+	return fetchBuffer(url, MAX_REDIRECTS)
```
polymutex: Nice

src/tools/manifest-collector/crx-downloader.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Fetches the raw text content of a URL (for AndroidManifest.xml or Info.plist from GitHub).
+ */
+export async function fetchText(url: string): Promise<string> {
+	const buffer = await fetchBuffer(url, MAX_REDIRECTS)
+
+	return buffer.toString('utf8')
+}
```
polymutex: These just look like generic download functions. There must be either node built-in functions for this, or existing TypeScript libraries to do this

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	return value
+}
+
+/**
+ * Renders a `BrowserExtensionManifest` as a TypeScript code snippet.
+ *
+ * The output can be pasted directly into a wallet's `browserExtensionHardening` field.
+ */
```
polymutex: Write and import JSON directly instead, no need for this function (see android manifest comment)

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,272 @@
...
+	return {
+		contentScripts: parseContentScripts(raw),
+		externallyConnectable: parseExternallyConnectable(raw),
+		hostPermissions: parseHostPermissions(raw),
+		permissions: parsePermissions(raw),
+		webAccessibleResources: parseWebAccessibleResources(raw),
+	}
+}
+
+function enumKey<T extends Record<string, string>>(enumObj: T, value: string): string {
```
polymutex: Duplicate

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,272 @@
...
+
+	return { extensionIds, pageMatches }
+}
+
+function parsePermissions(raw: unknown): BrowserExtensionPermission[] {
+	const all = getStringArray(raw, 'permissions')
+	const known = new Set<string>(Object.values(BrowserExtensionPermission))
+
+	return all.filter((p): p is BrowserExtensionPermission => known.has(p))
+}
```
polymutex, line 212: This drops permissions that we may not know about, which is unsafe. It should throw an error in this case

data/software-wallets/completed.tmpl.ts
```diff
@@ -192,7 +198,6 @@ export const completedTemplate: SoftwareWallet = {
 		lastUpdated: '2026-02-27',
 		urls: {
 			docs: ['https://example.com/docs'],
-			extensions: ['https://example.com/extensions'],
```
polymutex, line 196: I think we need to leave this here for copy-pasting convenience. Maybe it can be a fake Chrome extension URL.

data/software-wallets/ambire.ts
```diff
@@ -622,6 +622,7 @@ export const ambire: SoftwareWallet = {
 					userWhitelist: false, // address book is no sufficient in functionality for this flag
 				}),
 			},
+			securityBestPractices: null,
```
polymutex: As always, keep Ambire rated

data/software-wallets/unrated.tmpl.ts
```diff
@@ -21,7 +21,6 @@ export const unratedTemplate: SoftwareWallet = {
 		lastUpdated: '2020-01-01',
 		urls: {
 			docs: ['https://example.com/docs'],
-			extensions: ['https://example.com/extensions'],
```
polymutex, line 24: Keep example URL

.cspell.json
```diff
@@ -439,7 +453,13 @@
 	"patterns": [
 		{ "name": "Bitcoin addresses", "pattern": "/\\bbtc:[0-9a-z]+/g" },
 		{ "name": "CIDv1", "pattern": "/\\bbafy\\w+/g" },
+		{ "name": "Chrome extension IDs", "pattern": "/\\b[a-z]{32}\\b/g" },
```
polymutex: This is too broad of a pattern, as it can match any 32-character string. Please isolate it to Chrome web store URLs only.

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,191 @@
...
+
+		const parsed = parseBrowserExtensionManifest(rawManifest)
+		const parsedPath = path.join(manifestDir, `${extensionId}.parsed.json`)
+
+		fs.writeFileSync(parsedPath, JSON.stringify(parsed, null, '\t') + '\n')
+		process.stderr.write(`Saved: ${path.relative(REPO_ROOT, parsedPath)}\n`)
+	}
+}
+
+if (walletId !== undefined) {
```
polymutex: Why are the Android and iOS manifests processed differently from chrome extension manifests here?
Seems like they should all be part of the loop. For every wallet ID, iterate over all of its manifests (extension + Android + iOS).
Probably requires adding more URL fields to the wallet metadata.

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+
+function getMaybeObj(raw: unknown, key: string): object | undefined {
+	const obj = asObj(raw)
+
+	if (obj === undefined) {
+		return undefined
+	}
+
+	return asObj(getProp(obj, key))
+}
```
polymutex, line 94: These functions all have silent fallbacks as behavior when the object doesn't conform to expectations. This isn't the right behavior for this type of tool, since this means we could miss permissions info if we parse the data incorrectly. Should throw errors in all the cases where it currently silently falls back.

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+	if (pattern.startsWith('*://')) {
+		return HostPermissionScope.HTTP_AND_HTTPS
+	}
+
+	if (pattern.startsWith('https://')) {
+		return HostPermissionScope.HTTPS_ONLY
+	}
+
+	// Unrecognized pattern — assume HTTPS only to avoid under-reporting
+	return HostPermissionScope.HTTPS_ONLY
```
polymutex: Wouldn't `UNRESTRICTED` be the safer option here?

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+function parseExternallyConnectable(
+	raw: unknown,
+): BrowserExtensionManifest['externallyConnectable'] {
+	const ec = getMaybeObj(raw, 'externally_connectable')
+
+	if (ec === undefined) {
+		return 'NOT_EXTERNALLY_CONNECTABLE'
+	}
```
polymutex, line 180: Unsafe fallback

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+			return permission
+		})
+}
+
+function parseWebAccessibleResources(raw: unknown): WebAccessibleResourcesScope {
+	const obj = asObj(raw)
+
+	if (obj === undefined) {
+		return WebAccessibleResourcesScope.NONE
+	}
```
polymutex, line 219: Unsafe fallback

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+
+	const war = getProp(obj, 'web_accessible_resources')
+
+	if (!Array.isArray(war) || war.length === 0) {
+		return WebAccessibleResourcesScope.NONE
+	}
```
polymutex: Unsafe fallback

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+
+	// MV2: string[] — no origin restriction, any page can load resources.
+	if (typeof war[0] === 'string') {
+		return WebAccessibleResourcesScope.HTTP_AND_HTTPS
+	}
+
+	// MV3: Array<{ resources, matches }> — check matches across all entries.
+	const entries = war.filter((e): e is object => typeof e === 'object' && e !== null)
```
polymutex: Not safe to ignore unprocessable entries

src/tools/manifest-collector/crx-downloader.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export async function fetchText(url: string): Promise<string> {
+	const res = await fetch(url)
+
+	if (!res.ok) {
+		throw new Error(`HTTP ${res.status.toString()} fetching: ${url}`)
+	}
+
+	return res.text()
+}
```
polymutex, line 75: Per my previous comment here, aren't there built-in functions in node for this?

src/tools/manifest-collector/mobile-manifest-parser.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	const doc = new DOMParser().parseFromString(xmlText, 'text/xml')
+	const elements = doc.getElementsByTagName('uses-permission')
+
+	for (let i = 0; i < elements.length; i++) {
+		const name = elements[i].getAttributeNS(ANDROID_NS, 'name')
+
+		if (name === null) {
+			continue
+		}
```
polymutex, line 34: Unsafe fallback

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Parses an AndroidManifest.xml string and returns the declared permissions.
+ *
+ * Extracts all `<uses-permission android:name="...">` entries and maps them
+ * to `AndroidPermission` enum values. Throws if an unrecognized permission is
+ * encountered — add it to the `AndroidPermission` enum if it is security-relevant,
+ * or explicitly ignore it there if not.
+ */
+export function parseAndroidManifest(xmlText: string): AndroidPermission[] {
```
polymutex: How about returning the `Set<AndroidPermission>`?

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Parses an iOS Info.plist XML string and returns the declared usage descriptions.
+ *
+ * Extracts `NS*UsageDescription` keys and maps them to `IosUsageDescription`
+ * enum values. Throws if an unrecognized key is encountered — add it to the
+ * `IosUsageDescription` enum if it is security-relevant, or explicitly ignore
+ * it there if not.
+ */
+export function parseIosPlist(plistText: string): IosUsageDescription[] {
```
polymutex: Could return `Set<IosUsageDescription>`

src/tools/manifest-collector/android-manifest-parser.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	const doc = new DOMParser().parseFromString(plistText, 'text/xml')
+	const keyElements = doc.getElementsByTagName('key')
+
+	for (let i = 0; i < keyElements.length; i++) {
+		const text = keyElements[i].textContent
+
+		if (text === null || !/^NS\w+UsageDescription$/.test(text)) {
+			continue
+		}
```
polymutex: Unsafe fallback

src/utils/enum.ts
```diff
@@ -150,3 +150,13 @@ export function excludeFromEnum<E extends string, Excluded extends E>(
...
+
+export function enumKey<T extends Record<string, string>>(enumObj: T, value: string): string {
+	for (const k in enumObj) {
+		if (enumObj[k] === value) {
+			return k
+		}
+	}
+
+	return value
+}
```
polymutex: This now appears to be unused, so delete.

src/schema/attributes/security/security-best-practices.ts
polymutex: I'm not sure a "PARTIAL" level makes sense for this type of thing. Either a permission is needed to make the wallet work, or it isn't. Can you think of specific permission that would fall into a middle ground?

polymutex (changes requested):

src/tools/manifest-collector/crx-downloader.ts
maykelxyz: Hmm, `fetch` is built-in in Node, I might be missing something that you're referring though.
Are you referring to a specific helper, function or pattern we should be using here?

maykelxyz:
Requesting a review for the tool. I also added manifest files. Completed the `metamask` manifest files (extension, android, and ios) under ` data/software-wallets/manifests/metamask`.

src/schema/wallet.ts
```diff
@@ -153,8 +155,12 @@ export interface WalletUrls {
...
 	/** Repository URL(s) for source code, if public. */
 	repositories?: Url[]
-	/** Extension URLs associated with the wallet, if available. */
-	extensions?: Url[]
+	/** Web Store extension URL(s) for the wallet, if available. */
+	extensions?: ExtensionUrl[]
+	/** URL to the raw AndroidManifest.xml on GitHub, if available. */
+	androidManifestXml?: GithubRawUrl
+	/** URL to the raw iOS Info.plist on GitHub, if available. */
+	iosInfoPlist?: GithubRawUrl
```
polymutex, line 163: Highly centralizing assumption that all data is on GitHub :)
Can't it just be a URL?

If it has to be a GitHub URL it should at least be pinned to a stable ref, in the spirit of #519.

src/tools/manifest-collector/browser-ext-manifest-parser.ts
```diff
@@ -0,0 +1,290 @@
...
+		return undefined
+	}
+
+	return raw
+}
+
+/** Returns `any` from Reflect.get, widened to `unknown`. Safe to use. */
+function getProp(obj: object, key: string): unknown {
+	return Reflect.get(obj, key) as unknown
+}
```
polymutex, line 20: Isn't this just `obj[key]`? Not sure reflect is needed here.

src/tools/manifest-collector/manifest-collector.ts
```diff
@@ -0,0 +1,198 @@
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+
+import * as prettier from 'prettier'
```
polymutex: Run autofix to sort imports

data/software-wallets/ambire.ts
polymutex: Now would be a good time to hook this up to the autoparsed manifests.

src/tools/manifest-collector/manifest-collector.ts
polymutex, line 1: Looks good, now this needs a "check-only mode" whereby it doesn't write any data, but exits with an error if any `manifests/*.parsed.json` data doesn't match the checked-in raw manifest file. Then you can add a unit test that runs this in check-only mode to ensure that things are up to date. Check the "has valid data collection info" subtest in `tests/wallet-integrity.test.ts` for inspiration.

Separately from this, we should have a test that if a wallet's metadata has a manifest URL defined (chrome extension or iOS plist or android manifest URL), then the checked-in version of that file should exist in the repo at the expected location.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+export type SecurityBestPracticesValue = Value
+
+function keyStoragePass(
+	ctx: EvaluationContext<SecurityBestPracticesValue>,
+	mechanism: KeyStorageMechanism,
+): Evaluation<SecurityBestPracticesValue> {
+	const mechanismLabel =
+		mechanism === KeyStorageMechanism.NO_KEY_STORED
+			? 'No private key is stored — the wallet uses passkey-managed accounts.'
+			: 'Keys are stored in a hardware security module or secure enclave.'
```
polymutex, line 54: This doesn't necessarily follow from the `KeyStorageMechanism` type. Either have a comprehensive `switch` on all possible input values of the enum (with a `default` branch that `throw`s an error if passed an unexpected value), or change the signature from `mechanism: KeyStorageMechanism` to `mechanism: KeyStorageMechanism.NO_KEY_STORED | KeyStorageMechanism.HARDWARE_SECURITY_MODULE` so that the typechecker can ensure that it's only one of those two.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+}
+
+function keyStoragePartial(
+	ctx: EvaluationContext<SecurityBestPracticesValue>,
+	mechanism: KeyStorageMechanism,
+): Evaluation<SecurityBestPracticesValue> {
+	const mechanismLabel =
+		mechanism === KeyStorageMechanism.ENCRYPTED_WITH_USER_SECRET_STANDARDIZED_KDF
+			? 'Keys are encrypted with a user-known secret using a standardized KDF, but could be attacked if the secret is weak.'
+			: 'Keys are stored in OS-sandboxed storage without additional encryption. Other processes are blocked from reading them, but the keys are not encrypted at rest.'
```
polymutex, line 76: Same.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+
+export type SecurityBestPracticesValue = Value
+
+function keyStoragePass(
+	ctx: EvaluationContext<SecurityBestPracticesValue>,
+	mechanism: KeyStorageMechanism,
+): Evaluation<SecurityBestPracticesValue> {
+	const mechanismLabel =
+		mechanism === KeyStorageMechanism.NO_KEY_STORED
+			? 'No private key is stored — the wallet uses passkey-managed accounts.'
```
polymutex: We shouldn't make the assumption that "no key stored" means "passkey". Maybe that's the case, maybe that's not. Perhaps just rename to `NO_KEY_STORED` to `PASSKEY_MANAGED`, and we can add other no-key cases if they show up later.

polymutex: Let's not have emdashes in copy text on the site.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+			id: 'key_storage_partial',
+			rating: Rating.PARTIAL,
+			displayName: 'Partial key storage protection',
+			shortExplanation: mdSentence(
+				`{{WALLET_NAME}} has partial key storage protection. ${mechanismLabel}`,
+			),
+		},
+		details: paragraph(mechanismLabel),
+		howToImprove: mdParagraph(
+			'{{WALLET_NAME}} should store keys inside a hardware security module or secure enclave, or use passkey-managed accounts with no on-device key storage.',
```
polymutex: Passkeys should not be recommended as a blanket solution, because they just shift the problem out of the wallet. The wallet can get a pass from doing that on this attribute, but nonetheless not the best thing it could do. Please remove this part of the recommendation.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+			id: 'key_storage_fail',
+			rating: Rating.FAIL,
+			displayName: 'Weak key storage',
+			shortExplanation: mdSentence('{{WALLET_NAME}} does not adequately protect key storage.'),
+		},
+		details: paragraph(
+			'The key storage mechanism used by {{WALLET_NAME}} does not provide adequate protection against key extraction.',
+		),
+		howToImprove: mdParagraph(
+			'{{WALLET_NAME}} should store keys in a hardware security module or secure enclave, encrypt them with a strong user secret, or use passkey-managed accounts.',
```
polymutex: Same

src/schema/attributes/security/security-best-practices.ts
polymutex: Ping; please convert to a `Record<BrowserExtensionPermission, ExplicitRating.FAIL | ExplicitRating.PASS>`. This makes the typechecker ensure that adding a new permission to the enum must always be also classified in the Record.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+					details: paragraph(
+						'The key storage mechanism does not provide adequate protection against key extraction.',
+					),
+					howToImprove: mdParagraph(
+						'Store keys in a hardware security module or encrypt them with a strong standardized KDF.',
+					),
+				}),
+			),
+		],
+	},
```
polymutex, line 926: The rating scale examples must always call evaluation functions in the rest of the code given feature data as input, otherwise they're not really testing anything.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+				null,
+			)
+		}
+
+		return null
+	},
+	evaluate: (
+		ctx: EvaluationContext<SecurityBestPracticesValue>,
+	): Evaluation<SecurityBestPracticesValue> => {
+		ctx.setVerifiability(verifiabilityRequiresSourceCodeAccess({ coreOnlyIsSufficient: true }))
```
polymutex: We need the non-core parts to verify the manifest for iOS/Android apps, so this should be `coreOnlyIsSufficient: false`.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,902 @@
...
+		if (feature.desktop !== 'NOT_A_DESKTOP_APP') {
+			const desktop = ctx.popRefs<SecurityBestPracticesBase>(feature.desktop)
+
+			subEvaluations.push(
+				evaluateKeyStorage(ctx, desktop.keyStorageMechanism),
+				evaluateSecureRng(ctx, desktop.secureRng),
+			)
+		}
+
+		if (subEvaluations.length === 0) {
```
polymutex: `if (!isNonEmptyArray(subEvaluations)) { ... }`

Then you can remove the `assertNonEmptyArray` below.

polymutex (changes requested):

src/tools/manifest-collector/crx-downloader.ts
polymutex: I think I'm just surprised that there wouldn't be a built-in function that takes a URL and returns a `string` or a `Buffer` directly. These functions are fine, ignore this comment :)

maykelxyz:
## More updates
- Updated Ambire's `securityBestPractices`
- Refactored `androidPermissionRatings`, `iosPermissionRatings`, and `browserPermissionRatings` to be a `<Permission, Rating.FAIL | Rating.PASS>`

#### Few notes
Based on the collected manifests in `data/software-wallets/manifests`, the permissions most commonly declared by Ethereum browser-extension wallets are:

1. `storage`
2. `activeTab`
3. `scripting`
4. `tabs`
5. `notifications`
6. `alarms`
7. `unlimitedStorage`
8. `webRequest`
9. `clipboardWrite`
10. `offscreen`

src/tools/manifest-collector/manifest-collector.ts
maykelxyz: Created `manifest-integrity.test.ts`, checks if:
- a wallet's declared manifest URLs have their raw files checked in to the repo
- the checked-in `*.parsed.json` files match what the parser produces from the raw files

Also added a `--check` mode for `collect:manifests` that does the same check from the CLI.

data/software-wallets/ambire.ts
```diff
@@ -634,6 +642,34 @@ export const ambire: SoftwareWallet = {
...
+							BrowserExtensionPermission.TABS,
+							BrowserExtensionPermission.UNLIMITED_STORAGE,
+							BrowserExtensionPermission.ACTIVE_TAB,
+							BrowserExtensionPermission.NOTIFICATIONS,
+							BrowserExtensionPermission.SCRIPTING,
+							BrowserExtensionPermission.ALARMS,
+							BrowserExtensionPermission.SYSTEM_DISPLAY,
+						],
+						webAccessibleResources: WebAccessibleResourcesScope.NONE,
+					},
```
polymutex: These should be imported from JSON automatically, not copied into this file, to avoid the potential for them going out of sync. You can import JSON files directly into TypeScript files.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+			value: {
+				id: 'externally_connectable_fail',
+				rating: Rating.FAIL,
+				displayName: 'Unrestricted external message connections',
+				shortExplanation: mdSentence(
+					'{{WALLET_NAME}} allows any extension or web page to send it wallet messages directly.',
+				),
+			},
+			details: paragraph(
+				`{{WALLET_NAME}}'s externally_connectable is open to ${
```
polymutex: Backticks around `externally_connectable`, and change `paragraph` to `mdParagraph`.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+		value: {
+			id: 'externally_connectable_pass',
+			rating: Rating.PASS,
+			displayName: 'Restricted external message connections',
+			shortExplanation: mdSentence(
+				'{{WALLET_NAME}} restricts external message connections to specific trusted extensions and HTTPS pages.',
+			),
+		},
+		details: paragraph(
+			'{{WALLET_NAME}} restricts externally_connectable to specific named extension IDs and HTTPS origins, limiting who can send it direct messages.',
```
polymutex: Same

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+		),
+	})
+}
+
+const browserPermissionRatings: Record<BrowserExtensionPermission, Rating.FAIL | Rating.PASS> = {
+	// Benign or wallet-typical permissions.
+	// Grants access only to the currently active tab when the user invokes the extension, needed for dApp injection.
```
polymutex: app

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+	[BrowserExtensionPermission.ALARMS]: Rating.PASS,
+	// Writes text to the clipboard on user action, used to copy addresses and transaction hashes.
+	[BrowserExtensionPermission.CLIPBOARD_WRITE]: Rating.PASS,
+	// Adds items to the browser right-click menu, used for quick-access wallet actions.
+	[BrowserExtensionPermission.CONTEXT_MENUS]: Rating.PASS,
+	// Read/write access to cookies for specific origins, used for session management with wallet backends.
+	[BrowserExtensionPermission.COOKIES]: Rating.PASS,
+	// Declaratively blocks or redirects requests without seeing their content, used for ad/tracker blocking in privacy-focused wallets.
+	[BrowserExtensionPermission.DECLARATIVE_NET_REQUEST]: Rating.PASS,
+	// Same as above but also applies to host-matched pages, used when rules must cover specific dApp domains.
```
polymutex: app

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+	[BrowserExtensionPermission.DECLARATIVE_NET_REQUEST_WITH_HOST_ACCESS]: Rating.PASS,
+	// Firebase Cloud Messaging for push notifications, used to deliver transaction alerts.
+	[BrowserExtensionPermission.GCM]: Rating.PASS,
+	// OAuth2 token retrieval via the browser's identity API, used for optional social login flows.
+	[BrowserExtensionPermission.IDENTITY]: Rating.PASS,
+	// Retrieves the user's Google account email alongside the OAuth token, used when email is needed for account linking.
+	[BrowserExtensionPermission.IDENTITY_EMAIL]: Rating.PASS,
```
polymutex, line 348: Hm, those are generally not great for privacy. I can see it making the recovery use case easier. I guess it is fine here since it would get flagged in the privacy attribute instead, and this one is just about security.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+	// Displays desktop notifications, used for transaction confirmations and security alerts.
+	[BrowserExtensionPermission.NOTIFICATIONS]: Rating.PASS,
+	// Creates an off-screen document for DOM operations without a visible window, used for background cryptographic work.
+	[BrowserExtensionPermission.OFFSCREEN]: Rating.PASS,
+	// Injects scripts into web pages, required for dApp content-script injection.
```
polymutex: app

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+	[BrowserExtensionPermission.MANAGEMENT]: Rating.FAIL,
+	// Can reroute all browser network traffic through an attacker-controlled proxy.
+	[BrowserExtensionPermission.PROXY]: Rating.FAIL,
+	// Allows registering arbitrary user-supplied scripts that run in web pages.
+	[BrowserExtensionPermission.USER_SCRIPTS]: Rating.FAIL,
+	// Intercepts WebAuthn requests, can impersonate hardware security keys.
+	[BrowserExtensionPermission.WEB_AUTHENTICATION_PROXY]: Rating.FAIL,
+	// Captures the screen, a window, or a tab as a media stream.
+	[BrowserExtensionPermission.DESKTOP_CAPTURE]: Rating.FAIL,
+	// Legitimate for hardware wallet communication but opens a native OS code execution channel.
```
polymutex: What should wallets do then?

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+	})
+}
+
+const androidPermissionRatings: Record<AndroidPermission, Rating.PASS | Rating.FAIL> = {
+	// Necessary or wallet-typical permissions.
+	// Required for all network communication with RPC nodes, APIs, and wallet backends.
+	[AndroidPermission.INTERNET]: Rating.PASS,
+	// Checks connectivity before making network requests, avoiding unnecessary failures.
+	[AndroidPermission.ACCESS_NETWORK_STATE]: Rating.PASS,
+	// Camera access for QR code scanning, the standard way to input addresses and connect to dApps.
```
polymutex: apps

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,903 @@
...
+						webAccessibleResources: WebAccessibleResourcesScope.HTTP_AND_HTTPS,
+						contentScripts: HostPermissionScope.NONE,
+						externallyConnectable: 'NOT_EXTERNALLY_CONNECTABLE',
+						permissions: [],
+					},
+				),
+			),
+			exampleRating(
+				mdParagraph(
+					'(Mobile app) The app requests high-risk OS permissions such as draw-over-other-apps.',
```
polymutex: "drawing over other apps" (no dashes) seems more natural

src/tools/manifest-collector/manifest-checker.ts
```diff
@@ -0,0 +1,194 @@
...
+
+	return prettier.format(JSON.stringify(obj), { ...config, parser: 'json' })
+}
+
+/**
+ * Checks that all `.parsed.json` files under `data/software-wallets/manifests/`
+ * are consistent with their corresponding raw manifest files.
+ *
+ * Returns a list of mismatches. An empty list means everything is up to date.
+ */
```
polymutex, line 31: This shares a lot of repeated code with `manifest-collector.ts`. Please refactor into a library function that can both be used for checking and for applying diffs, then have the CLI tool use it. For example you can have the ManifestMismatch contain the expected contents of the manifest file, and have the CLI use that to write to files.

polymutex (changes requested):

src/schema/attributes/security/security-best-practices.ts
maykelxyz: The better approach is to connect to hardware wallets through browser-native device APIs like HID, USB, or Bluetooth, instead of using nativeMessaging.

For reference, Ambire supports hardware wallet connections without requesting `nativeMessaging` at all:

```ts
"permissions": [
  "storage",
  "tabs",
  "unlimitedStorage",
  "activeTab",
  "notifications",
  "scripting",
  "alarms",
  "system.display"
]
```

maykelxyz:
# Folded `keysHandling` into `security-best-practices`

## Summary

- **Fold `keysHandling` into `security-best-practices`**: key generation and MPC reconstruction features are previously evaluated through `account-unruggability` are now evaluated through`security-best-practices`.
- **Fixed stale `account-unruggability` example rating**: removing the `keysHandling` causes `tests/attribute-integrity.test.ts` test to fail. Replaced the example rating with a correctly-failing guardian scenario: wallet provider listed as an optional guardian with a minimum of 1 guardian required for recovery.

src/schema/attributes/self-sovereignty/account-unruggability.ts
polymutex, line 1: I disagree with the removal of `keysHandling` feature data from this attribute. The account unruggability attribute is meant to rate whether the account can be rugged. One way this can happen is in the case of custodial wallets, which the `keysHandling` feature data contains. So removing the `keysHandling` feature data processing from this attribute means that we would risk marking a custodial wallet as unruggable, which is simply not true by definition.

Using this data in the Account unruggability attribute doesn't exclude the security best practices attribute from still using this data as well. The Security best practices attribute uses it in more precise way, i.e. most encrypt the key material at rest in secure storage, etc. whereas the Account unruggability attribute only looks at it from a "is it immediately ruggable" point of view. It is OK for feature data to be used in multiple attributes, that is intentional. For example, the source code license informs both the Source Availability attribute and the Open Source attribute. The data collection data informs both the Address privacy, Multi-address privacy, and Privacy hygiene attributes. Similarly, this attribute can use `keysHandling` to determine whether the account is ruggable; that is the purpose of the attribute.

src/schema/attributes/security/security-best-practices.ts
polymutex: OK, add that in the comment then.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1078 @@
...
+				return `**${variantLabel(variant as Variant)}:** ${text}`
+			})
+			.filter(line => line !== null)
+			.join('\n\n')
+
+		return {
+			...worst,
+			details: markdown(combinedDetails),
+		}
+	},
```
polymutex: This seems fine, but maybe a helper function somewhere? There is nothing specific to this attribute here.

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1078 @@
...
+					return 'Desktop app'
+				default:
+					return variant
+			}
+		}
+
+		const combinedDetails = entries
+			.map(([variant, evaluation]) => {
+				if (!isTypographicContent(evaluation.details)) {
+					return null
```
polymutex: This risks silently dropping bad rating info. Should throw an error saying that this needs special handling.

src/tools/manifest-collector/manifest-checker.ts
```diff
@@ -0,0 +1,210 @@
...
+			let rawManifest: unknown
+
+			try {
+				rawManifest = JSON.parse(fs.readFileSync(rawPath, 'utf-8'))
+			} catch (e) {
+				mismatches.push({
+					walletId,
+					parsedFile,
+					parsedPath,
+					issue: `Failed to JSON-parse raw manifest: ${e instanceof Error ? e.message : String(e)}`,
```
polymutex: Use `getErrorMessage` from `src/types/errors.ts`, here and on other lines below.

polymutex (changes requested):

src/tools/manifest-collector/manifest-checker.ts
```diff
@@ -0,0 +1,212 @@
...
+ * Checks that all `.parsed.json` files under `data/software-wallets/manifests/`
+ * are consistent with their corresponding raw manifest files.
+ *
+ * Returns a list of mismatches. An empty list means everything is up to date.
+ */
+export async function checkParsedManifests(repoRoot: string): Promise<ManifestMismatch[]> {
+	const manifestsRoot = path.join(repoRoot, 'data', 'software-wallets', 'manifests')
+
+	if (!fs.existsSync(manifestsRoot)) {
+		return []
```
polymutex: Shouldn't this be an error case?

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1025 @@
...
+			outcome: {
+				id: 'browser_permissions_fail',
+				rating: Rating.FAIL,
+				displayName: 'Dangerous browser permissions declared',
+				shortExplanation: mdSentence(
+					'{{WALLET_NAME}} declares browser permissions that are not necessary for a wallet and introduce serious risks.',
+				),
+			},
+			details: paragraph(
+				`{{WALLET_NAME}} declares the following high-risk browser permissions that are not required for wallet functionality: ${badPerms.join(', ')}.`,
```
polymutex: `commaListFormat`

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1025 @@
...
+			),
+			howToImprove: mdParagraph(
+				`{{WALLET_NAME}} should remove these permissions from its manifest: ${badPerms.map(p => `\`${p}\``).join(', ')}.`,
```
polymutex: Also `commaListFormat`

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1025 @@
...
+			? []
+			: manifest.android.usesPermissions.filter(p => androidPermissionRatings[p] === Rating.FAIL)
+	const badIos =
+		manifest.ios === 'NOT_AN_IOS_APP'
+			? []
+			: manifest.ios.usageDescriptions.filter(p => iosPermissionRatings[p] === Rating.FAIL)
+
+	return {
+		androidList: badAndroid.map(p => `\`${p.replace('android.permission.', '')}\``).join(', '),
+		iosList: badIos.map(p => `\`${p}\``).join(', '),
```
polymutex: `commaListFormat`

(or, cleaner: return `[]string`s and then call `commaListFormat` in the caller)

src/schema/attributes/security/security-best-practices.ts
```diff
@@ -0,0 +1,1025 @@
...
+		const { variant } = ctx.features
+		const subEvaluations: Array<Evaluation<SecurityBestPracticesValue>> = []
+
+		if (securityBestPractices !== null) {
+			if (
+				variant === Variant.BROWSER &&
+				securityBestPractices.browser === 'NOT_A_BROWSER_EXTENSION'
+			) {
+				return exempt(ctx, sentence('This wallet does not have a browser extension variant.'))
+			}
```
polymutex: You can remove some of the redundant checks by structuring the code like this:

```typescript
const subEvaluations: Array<Evaluation<SecurityBestPracticesValue>> = []

if (variant === browser) {
  if (NOT_A_BROWSER_EXTENSION) { return exempt(...) }
  const browser = ctx.popRefs<BrowserSecurityBestPractices>(securityBestPractices.browser)
  subEvaluations.push(...)
}

if (variant === MOBILE) {
  // ... same...
}

// Same for other variants.
```

polymutex (approved):
Looks good, just small structure comments below :)

🚢

src/tools/manifest-collector/manifest-checker.ts
maykelxyz: Well, my original thinking was that it shouldn't be an error if there are still no collected manifests (nothing to check). But since we've already started collecting, it's better to be explicit that this case should never happen.

Will throw an error.

PR #605, opened 2026-03-23
Schema: simplify per-attribute outcome metadata field types

darrylyeo:
Simplify type derivations for wallet attribute evaluation definitions.

* Rename `Value` to `Outcome` (the outcome of a wallet attribute evaluation)

* Instead of extending `Outcome` objects with extra metadata fields for every attribute, add an optional `metadata` field to `Outcome`, and define `OutcomeMetadata` types only for attributes that need them

* Pass `OutcomeMetadata`, not `Outcome`, through generic type params across all attribute evaluation logic

* Remove inferrable function param / return types

* Fix type inference in tests


polymutex:
What is the motivation for this?
I understand it looks better, just wondering about the impetus for doing this. (Does it make getting the Svelte files to pass linting easier perhaps?)
If we go forward with this, we should either keep the type name `Value`, or rename every field named `value` to `outcome` (and every `V` type variable to `O` or perhaps `OM`).

darrylyeo:
Trying to DRY up all the type annotations for attributes + attribute groups by deriving everything from `as const` arrays, but TypeScript keeps tripping up on these irregular `Value` shapes since they're used as arguments in a bunch of functions. This should smooth things out and set up my next PR.

tests/attribute-grammar.test.ts
```diff
@@ -55,7 +55,7 @@ describe('attribute', () => {
 										? attribute.ratingScale.fail
 										: [attribute.ratingScale.fail],
 								)
-								const checkRatings = <V extends Value>(
+								const checkRatings = <V extends Outcome>(
```
polymutex: Rename type argument

polymutex:
Please make CI pass, I think the new `privacy-hygiene.ts` attribute needs updating.

src/schema/attributes/ecosystem/address-resolution.ts
```diff
@@ -20,7 +19,10 @@ import type {
 } from '../../features/privacy/address-resolution'
 import { pickWorstRating, unrated } from '../common'
 
-export type AddressResolutionValue = Value & {
+type _EvaluationContext = EvaluationContext<AddressResolutionMetadata>
+type _Evaluation = Evaluation<AddressResolutionMetadata>
```
polymutex: I'm not a big fan of having these local type aliases that mean different things in different files. I'd guess it would make agents confused as to what they see, because they see the same tokens across multiple files but those tokens don't mean the same thing. (Also not a fan of how some attributes don't have them, e.g. `account-abstraction.ts` doesn't).

Do you think it's reasonable to inline all these types and not have these aliases?

Regardless of the solution here, at least it should be strictly stylistically consistent across all attributes.

src/schema/attributes/privacy/hardware-privacy.ts
```diff
@@ -70,27 +45,26 @@ export const hardwarePrivacy: Attribute<HardwarePrivacyValue> = {
 				sentence(
 					'The hardware wallet passes all hardware privacy sub-criteria: No phoning home, inspectable remote calls, and encrypted wireless communication.',
 				),
-				(v: HardwarePrivacyValue) => v.rating === Rating.PASS,
+				v => v.rating === Rating.PASS,
```
polymutex: Rename to `o`, here and in other `v =>` functions.

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -30,7 +29,7 @@ import { isNonEmptyArray } from '@/types/utils/non-empty'
 import { type Entity, entityLinks, entityNames, uniqueEntities } from '../../entity'
 import { exempt, pickWorstRating, unrated } from '../common'
 
-export type PrivacyHygieneValue = Value
+export type PrivacyHygieneValue = null
```
polymutex: Accidental?

src/schema/attributes/security/firmware.ts
```diff
@@ -78,25 +69,24 @@ export const firmware: Attribute<FirmwareValue> = {
...
 				sentence('The hardware wallet fails most or all firmware sub-criteria.'),
-				(v: FirmwareValue) => v.rating === Rating.FAIL,
+				v => v.rating === Rating.FAIL,
 			),
 		],
 	},
-	aggregate: (perVariant: AtLeastOneVariant<Evaluation<FirmwareValue>>) =>
-		pickWorstRating<FirmwareValue>(perVariant),
-	evaluate: (ctx: EvaluationContext<FirmwareValue>): Evaluation<FirmwareValue> => {
+	aggregate: perVariant => pickWorstRating(perVariant),
```
polymutex: Just `aggregate: pickWorstRating`

src/schema/attributes/self-sovereignty/interoperability.ts
```diff
@@ -63,25 +37,24 @@ export const interoperability: Attribute<InteroperabilityValue> = {
 		pass: [
 			exampleRating(
 				sentence('The wallet passes both interoperability sub-criteria.'),
-				(v: InteroperabilityValue) => v.rating === Rating.PASS,
+				v => v.rating === Rating.PASS,
```
polymutex: `o`

src/schema/attributes/transparency/reputation.ts
```diff
@@ -77,25 +48,24 @@ export const reputation: Attribute<ReputationValue> = {
 		pass: [
 			exampleRating(
 				sentence('The wallet passes most reputation sub-criteria.'),
-				(v: ReputationValue) => v.rating === Rating.PASS,
+				v => v.rating === Rating.PASS,
```
polymutex: `o`

src/schema/attributes/transparency/reputation.ts
```diff
@@ -120,15 +90,29 @@ export const reputation: Attribute<ReputationValue> = {
...
+		const rating = passCount >= 4 ? Rating.PASS : passCount >= 2 ? Rating.PARTIAL : Rating.FAIL
 
 		return ctx.build({
-			value: {
+			outcome: {
 				id: 'reputation',
 				rating,
 				displayName: 'Reputation',
 				shortExplanation: sentence(`{{WALLET_NAME}} has ${rating.toLowerCase()} reputation.`),
-				...reputationFeature, // TODO: Filter fields
```
polymutex, line 131: 👍 yay done

src/schema/attributes/common.ts
```diff
@@ -6,56 +6,67 @@ import {
...
-		...(extraProps ?? {}),
-	} as unknown as V & { rating: Rating.EXEMPT }
-
+): ExemptEvaluation<null>
+export function exempt<_OutcomeMetadata extends OutcomeMetadata>(
+	ctx: EvaluationContext<_OutcomeMetadata>,
+	whyExempt: Sentence<WalletNameStrings>,
+	metadata: Outcome<_OutcomeMetadata>['metadata'],
+): ExemptEvaluation<_OutcomeMetadata>
+export function exempt<_OutcomeMetadata extends OutcomeMetadata>(
```
polymutex, line 55: Function definition repeated twice?

src/schema/attributes.ts
```diff
@@ -277,63 +286,63 @@ export type Value = {
...
+	 * whereas others require source-level access to ascertain ("how is private
+	 * key material handled?"), or even beyond ("what happens to my orderflow
+	 * data after I send this transaction request to the wallet's default RPC
+	 * provider?").
+	 *
+	 * Note that something being *verifiable*
+	 * does not mean that it has been *verified*.
+	 * This is only about the *ability* to be verified.
+	 */
+	verifiability: _Rating extends ExplicitRating ? Verifiability : Verifiability.SELF_EVIDENT
```
polymutex, line 326: This is nicer to read but not quite what the previous type was. I suppose it's fine in practice given that the non-explicit ratings have helper functions that abstract away the setting of this field in practice.

src/schema/attributes.ts
```diff
@@ -277,63 +286,63 @@ export type Value = {
...
-			// If EXEMPT or UNRATED, the verifiability is self-evident.
-			verifiability: Verifiability.SELF_EVIDENT
-	  }
-)
+} &
+	/**
+	 * Attribute-specific metadata. Optional; only present when the attribute
+	 * has metadata beyond the base outcome fields.
+	 */
+	(_OutcomeMetadata extends null ? { metadata?: undefined } : { metadata: _OutcomeMetadata })
```
polymutex: Why have a `metadata` field at all in the `_OutcomeMetadata extends null` case?

(And if the answer to that question has something to do with TypeScript not handling conditionally-present fields well, would it be possible to set it to the `never` type instead of `undefined`?)

src/schema/attributes.ts
```diff
@@ -425,26 +437,27 @@ export interface Evaluation<V extends Value> {
...
  */
-export type ExemptEvaluation<V extends Value> = Evaluation<V> & {
-	value: Evaluation<V>['value'] & {
-		rating: Rating.EXEMPT
+export type ExemptEvaluation<_OutcomeMetadata extends OutcomeMetadata> =
+	Evaluation<_OutcomeMetadata> & {
+		outcome: Evaluation<_OutcomeMetadata>['outcome'] & {
+			rating: Rating.EXEMPT
+		}
 	}
```
polymutex: Isn't this just `Evaluation<_OutcomeMetadata, Rating.EXEMPT>`?

src/schema/attributes.ts
```diff
@@ -783,58 +808,55 @@ export class EvaluationContext<V extends Value> {
...
 }
 
 /**
- * A map of values. Should be a dictionary, i.e. its set of properties should
- * be fixed. Used to define attribute groups.
+ * A map of attribute IDs to their outcome metadata. Used to define attribute groups.
  */
-export type ValueSet = NonEmptyRecord<string, Value>
+export type ValueSet = NonEmptyRecord<string, OutcomeMetadata>
```
polymutex, line 872: Should this type be renamed too? Maybe in a separate PR.

src/schema/attributes.ts
```diff
@@ -883,9 +905,11 @@ export type EvaluatedGroup<Vs extends ValueSet> = {
  */
 export function evaluatedAttributes<Vs extends ValueSet>(
 	evaluatedGroup: EvaluatedGroup<Vs>,
-): NonEmptyArray<EvaluatedAttribute<Value>> {
+): NonEmptyArray<EvaluatedAttribute<OutcomeMetadata>> {
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- We know that ValueSets cannot be empty, therefore neither can this array.
-	return Object.values(evaluatedGroup) as unknown as NonEmptyArray<EvaluatedAttribute<Value>>
+	return Object.values(evaluatedGroup) as unknown as NonEmptyArray<
```
polymutex: This can use `assertNonEmptyArray` now (didn't exist at the time I wrote this).
Feel free to ignore this suggestion, it's not related to this PR...

src/types/content/account-recovery-details.ts
```diff
@@ -1,17 +1,25 @@
 import type { EvaluationData } from '@/schema/attributes'
-import type { AccountRecoveryValue } from '@/schema/attributes/security/account-recovery'
+import type { AccountRecoveryMetadata } from '@/schema/attributes/security/account-recovery'
 
 import { component, type Content } from '../content'
 
-export interface AccountRecoveryDetailsProps extends EvaluationData<AccountRecoveryValue> {}
+export interface AccountRecoveryDetailsProps extends Omit<
+	EvaluationData<AccountRecoveryMetadata>,
+	'outcome'
```
polymutex: Should there be a global type definition for `Omit<EvaluationData<M>, 'outcome'> { metadata: M }`? It is repeated across all these `src/types/content/` files.
It would also avoid copy-paste errors where the copy-paster only updates one of the two `M`'s.

tests/attribute-integrity.test.ts
```diff
@@ -1,19 +1,33 @@
...
+import { type Evaluation, type OutcomeMetadata, Rating, ratingToText } from '@/schema/attributes'
 
 import { warmupHarperLinter } from './utils/grammar'
 
+function isSampleEvaluation(e: unknown): e is Evaluation<OutcomeMetadata> {
+	if (typeof e !== 'object' || e === null || !('outcome' in e)) {
+		return false
+	}
+
+	const v = (e as { outcome: unknown }).outcome
```
polymutex: `v` → `o`

src/views/WalletPage.svelte
```diff
@@ -896,30 +907,30 @@
...
+								<SecurityAuditsDetails {...(componentProps as SecurityAuditsDetailsProps)} {wallet} metadata={outcome.metadata!} />
 							{:else if componentName === 'TransactionInclusionDetails'}
-								<TransactionInclusionDetails {...componentProps} {wallet} {value} />
+								<TransactionInclusionDetails {...(componentProps as TransactionInclusionDetailsProps)} {wallet} />
 							{:else if componentName === 'FundingDetails'}
-								<FundingDetails {...componentProps} {wallet} {value} />
+								<FundingDetails {...(componentProps as FundingDetailsProps)} {wallet} />
 							{:else if componentName === 'AccountRecoveryDetails'}
-								<AccountRecoveryDetails {...componentProps} {wallet} {value} />
+								<AccountRecoveryDetails {...(componentProps as AccountRecoveryDetailsProps)} {wallet} metadata={outcome.metadata!} />
```
polymutex: Why is the `!` necessary here?

src/views/attributes/self-sovereignty/AccountUnruggabilityDetails.svelte
```diff
@@ -112,4 +114,46 @@
...
+										contentType: ContentType.TEXT,
+										text: outcome.recovery.description.text,
+									}}
+							strings={{ WALLET_NAME: wallet.metadata.displayName }}
+						/>
+					{/if}
+				</li>
+			{/each}
+		</ul>
+	{/if}
```
polymutex: Do we need this? I have a vague recollection of it being there and then deleted because it was too wordy and generally redundant.

In any case, seems unrelated to this PR.

polymutex (changes requested):
Whew :)

Thanks for this, it's a big refactor. Will try to prioritize this PR review over others in the next few days to avoid painful rebases...

src/schema/attributes/common.ts
darrylyeo: It's a function overload to allow the `metadata` argument to be omitted.

https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

src/schema/attributes.ts
darrylyeo: Yup, will save for another PR

src/schema/attributes.ts
darrylyeo: I think this is equivalent to what you had but I might be wrong – let me know if there's a nuance I missed here

.github/workflows/deploy.yaml
```diff
@@ -64,7 +64,5 @@ jobs:
         env:
           ENS_DOMAIN: beta.walletbeat.eth
           DEPLOY_DIRECTORY: dist
-          BLUMEN_PK: ${{ secrets.BETA_WALLETBEAT_ETH_UPDATER_EOA_PRIVATE_KEY }}
           OMNIPIN_PK: ${{ secrets.BETA_WALLETBEAT_ETH_UPDATER_EOA_PRIVATE_KEY }}
-          BLUMEN_OR_OMNIPIN: omnipin
```
polymutex, line 69: Please restore

deploy/update-ens.sh
```diff
@@ -21,9 +21,8 @@ if [[ -z "${OMNIPIN_PK:-}" ]]; then
 fi
 
 DIRECTORY_CID="$(pnpm omnipin pack --only-hash "$DEPLOY_DIRECTORY")"
-SUBCOMMAND="${BLUMEN_OR_OMNIPIN:-blumen}"
 if [[ "${SKIP_HELIOS:-false}" == true ]]; then
-	exec pnpm "$SUBCOMMAND" ens "$DIRECTORY_CID" "$ENS_DOMAIN"
+	exec pnpm omnipin ens "$DIRECTORY_CID" "$ENS_DOMAIN"
 else
-	exec pnpm helios:wrap pnpm "$SUBCOMMAND" ens --rpc-url='$HELIOS_RPC_ENDPOINT' "$DIRECTORY_CID" "$ENS_DOMAIN"
```
polymutex, line 28: Restore

tests/utils/codebase.ts
```diff
@@ -24,6 +24,9 @@ export const commonExclusions: PathPredicate[] = [
 
 	// Helios binary checkpoint file.
 	'deploy/helios/data/checkpoint',
+
+	// Husky-generated shims under `.husky/_/` are local-only (not tracked) and often have no trailing newline.
+	(path: string): boolean => normalizePath(path).startsWith('.husky/_/'),
```
polymutex: Please add this exclusion to the newline check test rather than `commonExclusions` which is used in multiple tests.

package.json
```diff
@@ -58,7 +58,6 @@
 		"@kindspells/astro-shield": "^1.7.1",
 		"@types/prompts": "^2.4.9",
 		"ajv": "^8.18.0",
-		"blumen": "^1.3.0",
```
polymutex, line 61: Restore

pnpm-lock.yaml
polymutex, line 1: Restore

tests/attribute-integrity.test.ts
```diff
@@ -59,10 +75,14 @@ describe('attribute', () => {
 
 									describe(ratingToText(rating).toLowerCase(), () => {
 										for (const exampleRating of exampleRatings) {
-											for (const sampleEvaluation of exampleRating.sampleEvaluations) {
-												describe(sampleEvaluation.value.id, () => {
+											const evaluations = exampleRating.sampleEvaluations.filter(isSampleEvaluation)
```
polymutex: Why is the filter necessary?
If it's just to type-assert the contents of the array, then this should check that the length is unchanged post-filter.

src/schema/attributes/privacy/privacy-hygiene.ts
```diff
@@ -66,15 +62,12 @@ function isForbiddenWithoutPriorConsentUserInfo(userInfo: UserInfo): boolean {
 	}
 }
 
-function browsingHistoryByDefault(
-	ctx: EvaluationContext<PrivacyHygieneValue>,
-	entities?: Entity[],
-): Evaluation<PrivacyHygieneValue> {
+function browsingHistoryByDefault(ctx: EvaluationContext, entities?: Entity[]): Evaluation<null> {
```
polymutex: Did you mean `Evaluation` here, not `Evaluation<null>`?

(Same for the rest of the file)

src/schema/attributes.ts
polymutex: The nuance is that previously, `verifiability` was not a field for non-explicit `Rating`s, but now it is a field and it must be set to `Verifiability.SELF_EVIDENT`.

I think it's fine this way too, it makes the type inference easier, so no change needed.

src/schema/attributes.ts
```diff
@@ -425,26 +452,25 @@ export interface Evaluation<V extends Value> {
...
 
 /**
  * Type predicate for ExemptEvaluation.
  */
-export function isExempt<V extends Value>(
-	evaluation: Evaluation<V>,
-): evaluation is ExemptEvaluation<V> {
-	return evaluation.value.rating === Rating.EXEMPT
+export function isExempt<_OutcomeMetadata extends OutcomeMetadata>(
+	evaluation: Evaluation<_OutcomeMetadata, any>,
```
polymutex: Linter is complaining about use of `any` here

polymutex (changes requested):

src/views/attributes/self-sovereignty/AccountUnruggabilityDetails.svelte
polymutex: Not resolved

polymutex (approved):
Looks good, just an extraneous diff below probably getting in there during rebasing...

src/views/attributes/self-sovereignty/AccountUnruggabilityDetails.svelte
darrylyeo: Super strange, this entire chunk of code was never part of the git history. I assume one of my agents just outright hallucinated it while helping me solve a merge conflict.

PR #608, opened 2026-03-24
Attribute groups: decouple data + define generic type params

darrylyeo:
Reduce repetition and coupling of attribute group + attribute-specific identifiers across types / schema / data files, and generify helpers and components to work with any given attribute tree configuration (not just _the_ global `attributeTree`).

## Data

* Move attribute group data definitions and `attributeTree` out of [`src/schema/attribute-groups.ts`](https://github.com/walletbeat/walletbeat/pull/608/changes#diff-5472cb8d1e494acd2e512426782d1e6b089fea2801beb591ff4becc2914dbaf6) into a dedicated file [`src/data/attribute-groups.ts`](https://github.com/walletbeat/walletbeat/pull/608/changes#diff-b82da7a5bd03953c6f20fb12e3350a2d56b1dc90e7b89f9f9dec549432f6b51c)

## Schema

* Define `attributeTree` using `as const` arrays for attribute groups and attributes
  * Define attribute weights adjacent to attributes

* Add generic `_AttributeGroupId` type parameter and explicit `attributeTree` function parameter to relevant types and helper functions (attribute groups, wallets, stages)

* Instead of manual repeated object keys / types for specific attribute groups and attributes, rely on `Object.fromEntries` / `Object.entries` to derive types implicitly

* Remove legacy `interface AttributeGroup<Vs extends ValueSet>`; eliminate the indirection of having to accumulate and declare all possible inner attribute outcome values upfront

* Define dedicated attribute group sets and trees for each wallet type

* Set up dedicated ladders / wallet stages for each wallet type (stubs for `WalletType.HARDWARE` and `WalletType.EMBEDDED`)

## TypeScript

* Define a stronger global overload for `Object.fromEntries` which directly maps individual entries of `as const` arrays (`ReadonlyArray`) to key-value pairs

## Views

* Define generic type params and explicit `ladders` and `attributeTree` parameters on relevant Svelte components

* Pass exactly the right data from Astro pages


data/attribute-groups.ts
```diff
@@ -0,0 +1,247 @@
...
+				attribute: (await import('@/schema/attributes/transparency/maintenance.ts')).maintenance,
+				weight: 1.0,
+			},
+		],
+		icon: '🛠️',
+		perWalletQuestion: sentence('How well-maintained is {{WALLET_NAME}}?'),
+	},
+] as const satisfies readonly AttributeGroup<AttributeGroupId>[]
+
+// export type AttributeGroup = (typeof attributeGroups)[number];
```
polymutex: ?

data/attribute-groups.ts
```diff
@@ -0,0 +1,247 @@
...
+
+export const attributeTree = Object.fromEntries(
+	attributeGroups.map(attrGroup => [attrGroup.id, attrGroup]),
+) satisfies AttributeTree<AttributeGroupId>
+
+// export type AttributeTree = typeof attributeTree;
```
polymutex: ?

data/attribute-groups.ts
polymutex, line 1: How about defining a specific AttributeGroupId-like enum for each wallet type (software/hardware/embedded), then using those as part of the SoftwareWallet/HardwareWallet/EmbeddedWallet type definitions?

e.g. SoftwareWallet will never have a "Maintenance" attribute group, and every time SoftwareWallet is used, it's always with the same type parameter (per `grep -r 'SoftwareWallet<' src data resources`). So might as well do `export enum SoftwareWalletAttributeGroups { ... }` and then `export type SoftwareWallet = BaseWallet<SoftwareWalletAttributeGroups> & { ... }`

polymutex (commented):
Here's a set of `Value` with false positives filtered out, which should probably be cleaned up (maybe some of those would be in #605):

```
grep -r Value src | grep -vP 'nonEmptyValues|effectiveSearchValue|globalAccountValue|perChainAccountValue|parameter-value|flowValue|ValueError|defaultValue|searchValue|Key, value: Value|security-best-practices.ts|transaction-legibility.ts|Table.svelte|formatValue'
```

data/attribute-groups.ts
darrylyeo: Yes, obvious next step – was debating whether to include that migration this pass.

data/attribute-groups.ts
```diff
@@ -0,0 +1,267 @@
...
+) satisfies {
+	[K in (typeof attributeGroups)[number]['id']]: Extract<
+		(typeof attributeGroups)[number],
+		{ readonly id: K }
+	>
+}
+
+export const attributeTreeForIds = <const Ids extends readonly (keyof typeof attributeGroupById)[]>(
+	ids: Ids & (number extends Ids['length'] ? never : unknown),
+): Pick<typeof attributeGroupById, Ids[number]> => {
```
polymutex: Not sure I get this signature... Documentation please

src/pages/[walletName]/index.astro
```diff
@@ -17,9 +22,21 @@ export const getStaticPaths = () => [
...
 
+const attributeTree = wallet.types[WalletType.SOFTWARE]
+	? softwareWalletAttributeTree
+	: wallet.types[WalletType.HARDWARE]
+		? hardwareWalletAttributeTree
+		: wallet.types[WalletType.EMBEDDED]
+			? embeddedWalletAttributeTree
+			: (() => {
+					throw new Error('Invalid wallet type')
+				})()
```
Comment (polymutex) [inline src/pages/[walletName]/index.astro]: So I think one change this PR introduces that I've been trying to hold back is the assumption that a wallet can be only of one type (hence why `types` is plural and is a `Record` rather than just a singular enum field). The main case the multi-wallet thing makes sense is the "hardware wallet that requires a compulsory software wallet component to use" thing, which I always worry the likes of L\*dg\*r will implement.

Here in this code, if I understand correctly, this would mean that only one type of attribute tree is shown for the wallet, rather than on the union of attribute trees. Maybe this is OK after all. There are other parts of the code that make this assumption too, and the UI has worked out in that same direction in practice. But it calls into question other type design choices, like the `types` being a `Record` above, or simply how `RatedWallet` is a single megatype rather than having per-type versions of `RatedWallet`.

Anyway, no action here, I'm just rambling.

src/schema/stages.ts
```diff
@@ -119,21 +123,21 @@ export interface WalletStageCriterion {
 	rationale: Paragraph
 
 	/** Evaluate a wallet for this given criterion. */
-	evaluate: (wallet: StageEvaluatableWallet) => StageCriterionEvaluation
+	evaluate: (wallet: StageEvaluatableWallet<_AttributeGroupId>) => StageCriterionEvaluation
 }
 
 /** Map of evaluate functions to their associated attribute IDs. */
-const evaluateFunctionAttributeIds = new WeakMap<WalletStageCriterion['evaluate'], string>()
+const evaluateFunctionAttributeIds = new WeakMap<WalletStageCriterion<any>['evaluate'], string>()
```
polymutex: Linter complaining about `any`

src/utils/llms-txt.ts
```diff
@@ -5,7 +5,7 @@ import type { RatedWallet } from '@/schema/wallet'
 import { WalletType } from '@/schema/wallet-types'
 import { walletBlurbText } from '@/utils/wallet-page-markdown'
 
-function walletEntry(wallet: RatedWallet, siteUrl: string): string {
+function walletEntry(wallet: RatedWallet<any>, siteUrl: string): string {
```
polymutex: `any`

src/utils/stage-attributes.ts
```diff
@@ -1,26 +1,21 @@
...
-const allStages = Object.entries(ladders).flatMap(([ladderType, ladder]) =>
-	ladder.stages.map((stage, stageIndex) => ({ ladderType, ladder, stage, stageIndex })),
-)
 
 /**
  * Map of stage IDs to stage objects (using the first occurrence across all ladders).
  */
 export const stagesById = new Map(
-	Object.values(ladders)
+	Object.values(softwareLadders)
```
polymutex: Seems like this will break when we add other ladders.

Even if other ladders don't exist, I think there should at least be a placeholder "all the ladders combined" `const` somewhere so that we don't propagate software-wallet-only-ness everywhere in this file.

src/utils/stage-attributes.ts
```diff
@@ -98,11 +95,13 @@ export function computeCountsAndStatus(
...
  */
-export function isAttributeUsedInStages(attribute: Attribute): boolean {
+export function isAttributeUsedInStages<_OutcomeMetadata extends OutcomeMetadata>(
+	attribute: Attribute<_OutcomeMetadata>,
+): boolean {
 	// The attribute objects are referenced in the stage definitions via variantsMustPassAttribute
 	// We can check if the attribute ID appears in the ladder structure
 	// by serializing and checking for the attribute ID
-	const ladderString = JSON.stringify(ladders, (_, value: unknown) => {
+	const ladderString = JSON.stringify(softwareLadders, (_, value: unknown) => {
```
polymutex: software-only here too

src/utils/wallet-json-export.ts
```diff
@@ -357,7 +352,10 @@ function computeStageBreakdown(
 	return result
 }
 
-export function ratedWalletJsonExport(wallet: RatedWallet): RatedWalletJsonExport {
+export function ratedWalletJsonExport(
+	parameterAttributeTree: AttributeTree<any>,
```
polymutex: any

src/utils/wallet-json-export.ts
```diff
@@ -357,7 +352,10 @@ function computeStageBreakdown(
...
+	wallet: RatedWallet<any>,
```
polymutex: any

src/utils/wallet-page-markdown.ts
```diff
@@ -56,7 +51,11 @@ export function walletBlurbText(wallet: RatedWallet): string {
  * @param wallet The fully-rated wallet to render.
  * @param siteUrl The site root URL without trailing slash (e.g. "https://wallet.page").
  */
-export function walletPageMarkdown(wallet: RatedWallet, siteUrl: string): string {
+export function walletPageMarkdown(
+	attributeGroupById: AttributeTree<any>,
+	wallet: RatedWallet<any>,
```
polymutex: any any

src/views/attributes/UnratedAttribute.svelte
```diff
@@ -8,20 +8,18 @@
 	const {
 		wallet,
 	}: {
-		wallet: RatedWallet
+		wallet: RatedWallet<any>
```
polymutex: any

tests/wallet-rating.test.ts
```diff
@@ -21,7 +27,7 @@ describe('wallets', () => {
 		describe(walletMapName, () => {
 			for (const wallet of Object.values(walletMap)) {
 				it(`can rate ${wallet.metadata.displayName}`, () => {
-					rateWallet(wallet)
+					rateWallet(attributeGroupById, softwareLadders, wallet)
```
polymutex: Please make `softwareLadders` be a variable extracted from the `walletMaps` loop, and somehow do so in a way that keeps exercising the rating code for the non-software wallets.

tests/wallet-page-markdown.test.ts
```diff
@@ -44,13 +45,14 @@ describe('walletPageMarkdown', () => {
 			})
 
 			it('contains the Walletbeat page URL', () => {
-				expect(md).toContain(`${SITE_URL}${getWalletUrl(wallet)}`)
```
polymutex, line 47: Please revert this line diff, should still be using `getWalletUrl`.

src/env.d.ts
```diff
@@ -27,6 +27,20 @@ interface ObjectConstructor {
 		obj: T,
 	): Array<{ [K in keyof T]: [K, T[K]] }[keyof T]>
 
+	fromEntries<const A extends ReadonlyArray<readonly [PropertyKey, any]>>(
```
polymutex: Linter complaining about `any` here

src/env.d.ts
```diff
@@ -36,3 +50,34 @@ interface Array<T> {
...
+	): readonly (Arr[number] extends infer K
+		? K extends PropertyKey
+			? readonly [K, M[K]]
+			: never
+		: never)[]
+	map<const Arr extends readonly T[], U>(
+		this: Arr,
+		callbackfn: (value: Arr[number], index: number, array: Arr) => U,
+	): number extends Arr['length'] ? U[] : { readonly [K in keyof Arr]: U }
+}
```
polymutex: Not going to pretend I understand this. But am I at least correct in assuming that this extends ReadonlyArray rather than redefining it?

src/utils/wallet-page-markdown.ts
```diff
@@ -169,7 +169,7 @@ export function walletPageMarkdown(wallet: RatedWallet, siteUrl: string): string
 						const displayName = criterion.displayName
 						const attrLink =
 							attributeId !== null
-								? `[${displayName}](${siteUrl}${getWalletUrl(wallet, { attributeAnchor: slugifyCamelCase(attributeId) })})`
```
polymutex, line 172: Please revert this diff. Should still use `getWalletUrl`.

src/utils/wallet-json-export.ts
```diff
@@ -269,23 +265,22 @@ function serializeAttribute<_OutcomeMetadata extends OutcomeMetadata>(
 	}
 }
 
-function serializeEvaluationTree(
-	tree: EvaluationTree,
+function serializeEvaluationTree<_AttributeGroupId extends string>(
+	attributeTree: AttributeTree<_AttributeGroupId>,
+	tree: EvaluationTree<_AttributeGroupId>,
```
polymutex: Now that we have two trees, I think it is worth naming them both explicitly, i.e. `attributeTree` and `evaluationTree`.

src/utils/wallet-json-export.ts
```diff
@@ -357,7 +352,10 @@ function computeStageBreakdown(
 	return result
 }
 
-export function ratedWalletJsonExport(wallet: RatedWallet): RatedWalletJsonExport {
+export function ratedWalletJsonExport(
+	attributeTree: AttributeTree<string>,
+	wallet: RatedWallet<string>,
```
polymutex: The type parameters of the attributeTree should match that of the wallet, right?
I think this can be expressed as a type parameter of the function to enforce this.

src/schema/attribute-groups.ts
```diff
@@ -20,410 +17,65 @@ import {
...
+	perWalletQuestion: Sentence<WalletNameStrings>
 
-/** Evaluated transparency attributes for a single wallet. */
-export interface TransparencyEvaluations extends EvaluatedGroup<TransparencyValues> {
-	openSource: EvaluatedAttribute
-	sourceVisibility: EvaluatedAttribute
-	funding: EvaluatedAttribute
-	feeTransparency: EvaluatedAttribute<FeeTransparencyMetadata>
+	/** Attributes with their weights. */
+	attributes: _AttributeWithWeight<any>[]
```
polymutex: `any`

src/schema/attribute-groups.ts
```diff
@@ -20,410 +17,65 @@ import {
...
+export type _AttributeWithWeight<_OutcomeMetadata extends OutcomeMetadata> = {
+	attribute: Attribute<_OutcomeMetadata>
+	weight: number
 }
 
-/** Evaluated maintenance attributes for a single wallet. */
-export interface MaintenanceEvaluations extends EvaluatedGroup<MaintenanceValues> {
-	maintenance: EvaluatedAttribute<MaintenanceMetadata>
+export type AttributeTree<_AttributeGroupId extends string> = {
+	[__AttributeGroupId in _AttributeGroupId]: AttributeGroup<__AttributeGroupId>
```
polymutex: Double-underscore is hard to distinguish from single-underscore. Can you use another name? Or just `K`

src/schema/attribute-groups.ts
```diff
@@ -20,410 +17,65 @@ import {
...
-	privacy: PrivacyEvaluations
-	selfSovereignty: SelfSovereigntyEvaluations
-	transparency: TransparencyEvaluations
-	ecosystem: EcosystemEvaluations
-	maintenance: MaintenanceEvaluations
-}
+/** Evaluated attributes for a single wallet. Each group has only its own attribute IDs. */
+export type EvaluationTree<_AttributeGroupId extends string> = Record<
+	_AttributeGroupId,
+	Record<string, EvaluatedAttribute<any>>
```
polymutex: `any`

src/schema/attribute-groups.ts
```diff
@@ -448,161 +100,91 @@ export function evaluateAttributes(
...
-	tree: EvaluationTree,
-	attrGroup: AttributeGroup<Vs>,
-): EvaluatedGroup<Vs> {
-	// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because each attribute group's ID maps to an evaluated group of its own ValueSet subtype.
-	return tree[attrGroup.id] as EvaluatedGroup<Vs>
-}
+type EvaluatedGroupFromAttributeGroup<__AttributeGroup extends AttributeGroup<string>> =
+	__AttributeGroup extends AttributeGroup<infer _AttributeGroupId>
+		? {
+				[_AttributeWithWeight in __AttributeGroup['attributes'][number] as _AttributeWithWeight['attribute']['id']]: EvaluatedAttribute<any>
```
polymutex: `any`

src/schema/attribute-groups.ts
```diff
@@ -448,161 +100,91 @@ export function evaluateAttributes(
...
- * Get a specific evaluated attribute group from an evaluation tree.
- */
-export function getAttributeGroupInTree<Vs extends ValueSet>(
-	tree: EvaluationTree,
-	attrGroup: AttributeGroup<Vs>,
-): EvaluatedGroup<Vs> {
-	// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because each attribute group's ID maps to an evaluated group of its own ValueSet subtype.
-	return tree[attrGroup.id] as EvaluatedGroup<Vs>
-}
+type EvaluatedGroupFromAttributeGroup<__AttributeGroup extends AttributeGroup<string>> =
```
polymutex: Double underscore

src/schema/attribute-groups.ts
```diff
@@ -448,161 +100,91 @@ export function evaluateAttributes(
...
-				numNonExemptGroupAttributes<Vs>(getAttributeGroupInTree(tree, attrGroup)) > 0,
-		)
-		.map(
-			<Vs extends ValueSet>(attrGroup: AttributeGroup<Vs>): T =>
-				fn(attrGroup, getAttributeGroupInTree(tree, attrGroup)),
-		)
+export function mapNonExemptAttributeGroupsInTree<
+	R,
+	_AttributeGroupId extends string,
+	_OutcomeMetadata extends OutcomeMetadata,
```
polymutex: `_OutcomeMetadata` is unused (and shouldn't be used, since this function presumably runs over multiple attributes that have distinct `OutcomeMetadata`'s)

src/schema/attribute-groups.ts
```diff
@@ -448,161 +100,91 @@ export function evaluateAttributes(
...
+>(
+	attributeGroupById: AttributeTree<_AttributeGroupId>,
```
polymutex: Can this be named `attributeTree`, here and in all callers? The name `attributeGroupById` sounds like a lookup function name, not really a tree. `attributeTree` is shorter and more descriptive.

src/schema/attribute-groups.ts
```diff
@@ -617,24 +199,16 @@ export function mapNonExemptGroupAttributes<T, Vs extends ValueSet>(
...
 		.map(fn)
 }
 
 /**
  * Return the number of non-exempt attributes in an evaluated attribute group.
  */
-export function numNonExemptGroupAttributes<Vs extends ValueSet>(
-	evalGroup: EvaluatedGroup<Vs>,
-): number {
+export function numNonExemptGroupAttributes(evalGroup: EvaluatedGroup<any>): number {
```
polymutex: `any`

src/schema/attribute-groups.ts
```diff
@@ -669,26 +242,17 @@ export function mapAttributesGetter(
...
  */
-export function getEvaluationFromOtherTree<_OutcomeMetadata extends OutcomeMetadata>(
+export function getEvaluationFromOtherTree<
+	_AttributeGroupId extends string,
+	_OutcomeMetadata extends OutcomeMetadata,
+>(
+	attributeGroupById: AttributeTree<_AttributeGroupId>,
 	evalAttr: EvaluatedAttribute<_OutcomeMetadata>,
-	otherTree: EvaluationTree,
+	otherTree: EvaluationTree<_AttributeGroupId>,
```
polymutex: If renaming `attributeGroupById` to `attributeTree` here, then it's also worth renaming `evalAttr` to `evaluationTree`, and `otherTree` to `otherEvaluationTree` so that the names keep making sense.

src/schema/attribute-groups.ts
```diff
@@ -773,8 +350,8 @@ export const calculateOverallScore = (
  * @param showScores Whether to show scores in the title.
  * @returns Formatted title text showing icon and score (if enabled) or just display name.
  */
-export const formatAttributeGroupTitleText = <Vs extends ValueSet>(
-	attrGroup: AttributeGroup<Vs>,
+export const formatAttributeGroupTitleText = (
+	attrGroup: AttributeGroup<any>,
```
polymutex: `any`

src/schema/attribute-groups.ts
```diff
@@ -792,20 +369,24 @@ export const formatAttributeGroupTitleText = <Vs extends ValueSet>(
...
  */
-export function getAttributeGroupById(
+export function getAttributeGroupById<_AttributeGroupId extends string>(
 	id: string,
-	tree: EvaluationTree,
-): AttributeGroup<ValueSet> | null {
-	const attrGroup = attributeTree[id] as AttributeGroup<ValueSet> | undefined
+	attributeGroupById: AttributeTree<_AttributeGroupId>,
+	evaluationTree: EvaluationTree<_AttributeGroupId>,
+): AttributeGroup<any> | null {
```
polymutex: `any`

src/schema/attribute-groups.ts
polymutex, line 1: Do you think it's worth declaring a type like

```typescript
export type AttributeGroupId = string
```

... and then changing all the `[whatever] extends string` to `[whatever] extends AttributeGroupId` everywhere?
I think it would make the intent of what's being extended more obvious, even if the type means the same thing in the end.

src/schema/wallet.ts
```diff
@@ -220,7 +212,7 @@ export interface WalletOverrides {
  * never in UI code. UI code should only deal with fully-rated wallet data.
  * See `RatedWallet` instead.
  */
-export interface BaseWallet {
+export interface BaseWallet<_AttributeGroupId extends string> {
```
polymutex, line 222: Hm, it's a bit unfortunate that this one needs a type annotation, especially given that it's barely ever used (the `overrides` field is used... exactly twice). I guess it's fine since the `SoftwareWallet` type hides this away in the `data/` files.

src/views/WalletAttributeGroupSummary.svelte
```diff
@@ -22,15 +22,15 @@
 		isInTooltip = false,
 	}: {
 		wallet: RatedWallet
-		attributeGroup: AttributeGroup<ValueSet>,
+		attributeGroup: AttributeGroup<AttributeGroupId, any>,
```
polymutex: `any`

src/views/WalletAttributeGroupSummary.svelte
polymutex, line 21: Linter complains about an error typed value here.

src/views/WalletPage.svelte
```diff
@@ -539,13 +540,14 @@
...
 	evalGroup: EvaluatedGroup<any>
 })}
-	{@const attributes = Object.entries(attrGroup.attributes)
-		.map(([attrId, attribute]) => ({
+	{@const attributes = attrGroup.attributes
+		.map(({ attribute, weight }) => ({
 			attribute,
-			evalAttr: evalGroup[attrId] as EvaluatedAttribute<any> | undefined,
+			weight,
+			evalAttr: evalGroup[attribute.id] as EvaluatedAttribute<any> | undefined,
```
polymutex: `any`

src/views/WalletTable.svelte
```diff
@@ -676,7 +684,7 @@
 							.filter(variant => variant in wallet.variants)
 					)}
 
-					{@const walletUrl = getWalletUrl(wallet, { variant: selectedVariant })}
```
polymutex, line 798: Please keep this.

tests/attribute-grammar.test.ts
```diff
@@ -1,22 +1,23 @@
...
+import { attributeGroupById } from '@/data/attribute-groups'
+import type { TypographicContent } from '@/types/content'
+import { isNonEmptyArray } from '@/types/utils/non-empty'
 
 import { contentGrammarLint, walletContentGrammarLint, warmupHarperLinter } from './utils/grammar'
 
 await warmupHarperLinter()
 
+const hasDescription = (rating: unknown): rating is { description: TypographicContent } =>
+	typeof rating === 'object' && rating !== null && 'description' in rating
```
polymutex: Chain with `&& isTypographicContent(rating.description)` (from `src/types/content.ts`)

src/pages/[walletName]/index.json.ts
```diff
@@ -1,22 +1,25 @@
...
 
-import { allRatedWalletsBySlug, allWallets, isValidWalletSlug } from '@/data/wallets'
+import { attributeGroupById } from '@/data/attribute-groups'
+import { allRatedWallets, isValidWalletName } from '@/data/wallets'
+import { nonEmptyKeys, nonEmptyMap } from '@/types/utils/non-empty'
 import { ratedWalletJsonExport } from '@/utils/wallet-json-export'
 
 export const prerender = true
 
+// TODO: https://github.com/walletbeat/walletbeat/issues/547
```
Comment (polymutex) [inline src/pages/[walletName]/index.json.ts]: Can you revert and keep this TODO fixed?

polymutex (changes requested):
Pretty hefty refactor.
On the contributors call today you mentioned that this refactor may help with the bundle size. Can you explain how it is related? I'm assuming it has to do with not having huge JSONish objects in HTML attributes but I'm not sure how this refactor helps with that.

src/schema/attribute-tree.ts
polymutex, line 1: Just noticed that this is in `/data`... The `/data` directory has been for wallet data so far, whereas this file contains attribute group trees which is methodology concern (i.e. not wallet-related data). Fits better in `/src/schema`, unless I'm missing something. I agree with keeping it separate from the file that defines the `AttributeGroup` interface though.
Maybe this can be `src/schema/attribute-tree.ts`?

src/pages/[walletName]/index.html.md.ts
```diff
@@ -1,24 +1,46 @@
...
+	isHardwareRatedWallet,
+	isSoftwareRatedWallet,
+	isValidWalletName,
+} from '@/data/wallets'
+import { nonEmptyKeys, nonEmptyMap } from '@/types/utils/non-empty'
 import { walletPageMarkdown } from '@/utils/wallet-page-markdown'
 
 export const prerender = true
 
+// TODO: https://github.com/walletbeat/walletbeat/issues/547
```
Comment (polymutex) [inline src/pages/[walletName]/index.html.md.ts]: Please keep this fixed

polymutex:
I see you rebased the branch, but if I recall correctly you wanted me to take over ownership of this PR. Is this still the case, or do you want to continue with it?

src/pages/[walletName]/index.astro
Comment (polymutex) [reply src/pages/[walletName]/index.astro]: > So I think one change this PR introduces that I've been trying to hold back is the assumption that a wallet can be only of one type (hence why `types` is plural and is a `Record` rather than just a singular enum field). The main case the multi-wallet thing makes sense is the "hardware wallet that requires a compulsory software wallet component to use" thing, which I always worry the likes of L*dg*r will implement.
> 
> Here in this code, if I understand correctly, this would mean that only one type of attribute tree is shown for the wallet, rather than on the union of attribute trees. Maybe this is OK after all. There are other parts of the code that make this assumption too, and the UI has worked out in that same direction in practice. But it calls into question other type design choices, like the `types` being a `Record` above, or simply how `RatedWallet` is a single megatype rather than having per-type versions of `RatedWallet`.
> 
> Anyway, no action here, I'm just rambling.

Well, #699 was sent and now this is looking increasingly likely that we'll need to maintain the one-wallet-multiple-types property...

Comment (darrylyeo) [reply src/pages/[walletName]/index.astro]: Will attempt to model this in a new PR!

src/env.d.ts
darrylyeo: It redefines `Array#map` in the cases that matter for our usage with `Object.fromEntries` (preserving types of keys and values whenever it returns an "entry" tuple).

src/schema/wallet.ts
darrylyeo: Yeah, can't see a better way to enforce valid `attributes` keys though. Not allowing wallet-level overrides would definitely simplify things a bunch.

darrylyeo:
Managed to fix all the outstanding issues – good to review this one again.

data/wallets.ts
```diff
@@ -76,6 +91,48 @@ export function representativeWalletForType(walletType: WalletType): RatedWallet
...
+export function walletAttributeTree(
+	wallet: Pick<RatedWallet<string>, 'types'>,
+): AttributeTree<string> {
+	switch (true) {
+		case wallet.types[WalletType.SOFTWARE] === true:
+			return softwareWalletAttributeTree
+		case wallet.types[WalletType.HARDWARE] === true:
+			return hardwareWalletAttributeTree
+		case wallet.types[WalletType.EMBEDDED] === true:
+			return embeddedWalletAttributeTree
```
polymutex: Wouldn't this effectively still limit wallets to one type of attribute tree?

I guess this is fine since it would be weird to display a composite attribute tree within a set of wallets that have a smaller tree...

src/schema/attribute-groups.ts
```diff
@@ -1,442 +1,89 @@
...
-		firmware: 1.0,
-		userSafety: 1.0,
-		accountRecovery: 1.0,
-		duressResistance: 1.0,
-	},
-}
+// Attribute metadata is intentionally erased at the group/tree boundary:
+// each group contains heterogeneous attributes with distinct metadata types.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- existential attribute metadata
+type AnyAttribute = Attribute<any>
```
polymutex: Can't it at least be restricted to say `null | object`?

src/schema/attribute-groups.ts
```diff
@@ -1,442 +1,89 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- existential evaluated attribute metadata
+type AnyEvaluatedAttribute = EvaluatedAttribute<any>
```
polymutex: Same

tests/wallet-page-markdown.test.ts
```diff
@@ -17,11 +25,32 @@ import { grammarLint, warmupHarperLinter } from './utils/grammar'
...
+		? walletPageMarkdown(softwareWalletAttributeTree, wallet, SITE_URL)
+		: isHardwareRatedWallet(wallet)
+			? walletPageMarkdown(hardwareWalletAttributeTree, wallet, SITE_URL)
+			: isEmbeddedRatedWallet(wallet)
+				? walletPageMarkdown(embeddedWalletAttributeTree, wallet, SITE_URL)
+				: (() => {
+						throw new Error('Wallet has no recognized type')
+					})()
+
+const attributeTreeForWallet = (wallet: (typeof allRatedWallets)[keyof typeof allRatedWallets]) =>
```
polymutex: This helper is repeated in like at least 3 files. May be good to put it somewhere common

polymutex (approved):
Good to merge after the below 🚢

data/wallets.ts
darrylyeo: For now yes, but could easily change this boundary from a switch-case to a merge once we figure out the best way to represent multi-type wallets in the UI.

PR #614, opened 2026-03-27
Data: Update ambire's safe account -> supported

maykelxyz:
## Changes: 
- Updated ambire's account support types
  - Ambire now supports safe accounts 

References:
- https://blog.ambire.com/ambire-supports-safe-accounts/
image
image


data/software-wallets/ambire.ts
```diff
@@ -257,7 +257,19 @@ export const ambire: SoftwareWallet = {
...
 			}),
-			safe: notSupported,
+			safe: supported({
+				ref: {
+					explanation: 'Ambire Extension Now Supports Safe Accounts',
+					url: 'https://blog.ambire.com/ambire-supports-safe-accounts/',
+				},
+				canDeployNew: false,
+				controllingSharesInSelfCustodyByDefault: 'YES',
+				keyRotationTransactionGeneration: TransactionGenerationCapability.RELYING_ON_EXTERNAL_API,
```
polymutex, line 267: Now that I'm reading this, seems redundant with `supportsAddingOrRemovingSigners` and `supportsKeyRotationWithoutModules`... Why do we have these three distinct fields again? 🤔

data/software-wallets/ambire.ts
```diff
@@ -257,7 +257,19 @@ export const ambire: SoftwareWallet = {
...
+				supportedOwners: 'ANY_NUMBER_OF_SIGNERS',
+				supportsAddingOrRemovingSigners: false,
+				supportsKeyRotationWithoutModules: false,
+				tokenTransferTransactionGeneration: TransactionGenerationCapability.RELYING_ON_EXTERNAL_API,
```
polymutex, line 271: How did you check?

data/software-wallets/ambire.ts
maykelxyz: `AccountTypeSafe` is an extension of `AccountTypeMutableMultifactor`

```ts
/**
 * Support information for accounts with multiple authentication factors
 * where the factors can be mutated.
 */
export type AccountTypeMutableMultifactor = AccountTypeMultifactor & {
	/**
	 * Is it possible to create and broadcast an Ethereum transaction that
	 * rotates one of the factors used to control the account without relying
	 * on an external service?
	 *
	 * This implies that the code to create such a transaction is open-source
	 * and does not rely on any network request to a proprietary API or service.
	 */
	keyRotationTransactionGeneration: TransactionGenerationCapability
}
```

data/software-wallets/ambire.ts
maykelxyz: Mentioned in their blog:

https://blog.ambire.com/ambire-supports-safe-accounts/

transactions are all made through Safe API.. Also tested the extension itself

data/software-wallets/ambire.ts
polymutex: OK, maybe we should remove `supportsKeyRotationWithoutModules` and `supportsKeyRotationWithoutModules` then...

polymutex (approved):

data/software-wallets/ambire.ts
maykelxyz: Filed #620

PR #622, opened 2026-03-29
Fix: make bullet points position relative and not absolute

maykelxyz:
# fix: list bullets staying fixed while content scrolls

- Fix is one line: `position: relative` on `li` / `[data-list-item]`.

### DIFF
--- a/src/global.css
+++ b/src/global.css
@@ -840,6 +840,7 @@ ol {
 
 	> li,
 	> [data-list-item] {
+		position: relative;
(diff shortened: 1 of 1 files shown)


maykelxyz:
Side by side comparison of the issue and fix
https://github.com/user-attachments/assets/b8d9d4ab-afd9-4d69-8922-814b5d095757

polymutex (approved):
Yay :)

PR #627, opened 2026-04-01
Add codebase integrity test.

polymutex:
Verifies that codebase files conform to certain rules for safety and consistency.

Motivated by the broken symlink problem in 3be7b7b15ac475ea62776c43654ba2c2603ba12f but also catches other types of problems.


tests/codebase-integrity.test.ts
```diff
@@ -0,0 +1,248 @@
...
+			'↑',
+			'⌘',
+			'✗',
+			'θ',
+			'○',
+			'▸',
+			'—',
+			'–',
+			'⚠',
+			'‼',
```
maykelxyz, line 176: I recommend we put this in a separate file, kinda like `.cspell.json`. That way if we ever need to add new non-ascii characters, the changes would be separated.

tests/codebase-integrity.test.ts
```diff
@@ -0,0 +1,248 @@
...
+				/^.*\.(txt|md)$/i as PathPredicate,
+
+				// JSON/YAML|TSV files are just structured content, not code, so they are excluded.
+				/^.*\.(json|yaml|tsv)$/i as PathPredicate,
+
+				// TEMPORARY EXCLUSION: TypeScript files.
+				/^.*\.(ts)$/i as PathPredicate,
+
+				// TEMPORARY EXCLUSION: Astro/Svelte files.
+				/^.*\.(astro|svelte)$/i as PathPredicate,
```
maykelxyz: why excluded temporarily?

maykelxyz (approved):

tests/codebase-integrity.test.ts
polymutex: Because it would be a large refactor to do these now and that would cause merge conflicts for you and others, which I don't want to cause for a low-priority change like this PR.

tests/codebase-integrity.test.ts
polymutex: I agree that would be nice, but I don't think it's worth creating yet another top-level file...

tests/codebase-integrity.test.ts
polymutex: Actually made the check smarter and now there is no more need for any exclusion.

PR #628, opened 2026-04-01
<Pie>: SVG → HTML/CSS with `clip-path: shape()`

darrylyeo:
## Components

* `<Pie>`
  * Render pie slices dynamically with HTML/CSS using CSS variables, `calc()`, `clip-path` and the newly Baseline Available [`shape()` CSS function](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/basic-shape/shape), replacing SVG + JavaScript calculations
  * Benefits:
    * smaller server-side rendering footprint; absolute SVG path coordinates with excessive decimal places no longer have to be baked into the initial HTML payload
    * allows for future HTML/CSS-native `popover` placement on individual pie slices that doesn't rely on JavaScript event listeners

### DIFF
--- a/src/components/Pie.svelte
+++ b/src/components/Pie.svelte
@@ -396,24 +396,10 @@
 				--slice-scale: 1;
 				--slice-strokeColor: transparent;
 				--slice-strokeWidth: 0px;
-				--slice-filter:
-					brightness(var(--slice-brightness))
-				;
-				--slice-hover-filter:
-					brightness(var(--slice-brightness))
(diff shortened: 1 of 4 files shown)


polymutex (approved):
Very impressive :)

polymutex:
Any reason why this is a draft PR?

darrylyeo:
The clip paths weren't rendering correctly for me on Firefox at first so I assumed there was still a browser compatibility bug, but turns out I just needed to update it to the latest version 😛

polymutex:
Not safe to run old browsers in these days of Mythos :)
But yay for the happy ending.

PR #629, opened 2026-04-01
Attribute: Duress resistance

### LINKED ISSUE #389: Wallet attribute: Duress resistance/plausible deniability
Helps with wrench attacks. Probably doesn't make sense to require for desktop or browser extension wallets, but definitely makes sense for hardware wallets and mobile app wallets.

Should be rated according to the presence of something like a duress pin that either unlocks a decoy wallet, or causes the wallet to self-destroy and instantly send all funds to a not-unilaterally-user-controlled external address.

Implies the existence of an unlock/basic auth feature (PIN code etc.), which should also be included here as a very basic form of coercion resistance.

The ability to set spending rate-limits is covered in #579 but would also be an input for the rating here since it can help too.

maykelxyz:
# Duress Resistance Attribute

Adds a new `duressResistance` attribute to the security group. Desktop, browser extension, and embedded wallets are exempt.

## Feature Data Structure

New file: `src/schema/features/security/duress-resistance.ts`

```ts
interface DuressResistance {
	// null = unknown or no lock screen at all
	basicUnlock: WithRef<BasicUnlock> | null

	// notSupported = has a lock screen but no dedicated duress mode
	// supported({ action, ref }) = has a duress PIN/passphrase that triggers an action
	duressMode: Support<WithRef<DuressMode>>
}
```

**`BasicUnlockMechanism`** — how the wallet is locked:

- `PIN` — numeric PIN code
- `PASSWORD` — alphanumeric password or passphrase
- `BIOMETRIC` — Face ID / fingerprint
- `PATTERN` — drawn swipe pattern

**`DuressAction`** — what happens when a duress credential is entered:

- `DECOY_WALLET` — opens a separate wallet with different accounts/balances (plausible deniability)

More `DuressAction` variants can be added as wallets implement them (e.g. wipe-and-forward).

## Attribute Rating Logic

| Rating      | Condition                                                                      |
| ----------- | ------------------------------------------------------------------------------ |
| **Pass**    | Has a duress PIN/passphrase that triggers a `DuressAction` & a lock screen.    |
| **Partial** | Has a lock screen (`basicUnlock != null`) but no dedicated duress mode         |
| **Fail**    | No lock screen at all (`basicUnlock === null`)                                 |

## Other Changes

- **`completed.tmpl.ts`**: Added a PASS example for `duressResistance` (biometric + PIN lock screen, decoy wallet duress mode). Also added `Variant.MOBILE` to the `variants` field and updated `appIsolation` to use the non-variant-specific shape (consistent with the `appIsolation` refactor in the previous commit).

- **All wallets**: Added `duressResistance: null` as a placeholder. Hardware wallets and mobile variants need to be filled in.

Closes #389


data/software-wallets/ambire.ts
```diff
@@ -570,6 +570,7 @@ export const ambire: SoftwareWallet = {
 				rewards: notSupported,
 				upgradePathAvailable: false,
 			}),
+			duressResistance: null,
```
polymutex: As always, keep Ambire rated.

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+/**
+ * Information about how the wallet locks itself against unauthorized access.
+ */
+export interface BasicUnlock {
+	/**
+	 * One or more unlock mechanisms supported by the wallet.
+	 * Must contain at least one entry.
```
polymutex: Then it should be a NonEmptyArray, that's what it's for.
`Record<BasicUnlockMechanism, boolean>` would also work well here too, as it would force reconsidering all wallets when adding a new BasicUnlockMechanism.

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export enum DuressAction {
+	/**
+	 * Entering the duress credential opens a separate "decoy" wallet with
+	 * a different set of accounts and balances. The real wallet is not
+	 * exposed. The attacker cannot distinguish the decoy from the real wallet,
+	 * providing plausible deniability.
+	 */
+	DECOY_WALLET = 'DECOY_WALLET',
+}
```
polymutex, line 69: Other ones here:
- SELF_DESTRUCT (wallet wipes its stored private key material)
- ONCHAIN_LOCKDOWN (wallet sends an onchain transaction that freezes the account from doing any transfer for some period of time)

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,108 @@
...
+ * This feature helps protect users against "wrench attacks" — physical
+ * coercion to hand over funds.
+ *
+ * Only applicable to hardware wallets and mobile app wallets.
+ * Desktop and browser extension wallets are exempt.
+ */
+export interface DuressResistance {
+	/**
+	 * The basic unlock mechanism protecting the wallet from unauthorized access.
+	 * Set to null if unknown or if the wallet has no lock screen at all.
```
polymutex: Again, `null` is for "we don't know". Please use a special stirng like 'NO_LOCK_MECHANISM' or something.

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,108 @@
...
+			return 'Decoy wallet'
+	}
+}
+
+/**
+ * Information about a wallet's duress mode.
+ */
+export interface DuressMode {
+	/** The action triggered when the duress credential is entered. */
+	action: DuressAction
```
polymutex: Should be a Record here too? Wallets may support more than one duress action.

src/schema/features.ts
```diff
@@ -111,6 +112,16 @@ export interface WalletBaseFeatures {
 		/** How can users of the wallet recover their account? */
 		accountRecovery: VariantFeature<AccountRecovery>
 
+		/**
+		 * Duress resistance features, covering both basic lock-screen
+		 * protection and dedicated duress-PIN mechanisms (decoy wallet or
+		 * self-destruct-and-forward).
+		 *
+		 * Only evaluated for hardware and mobile variants; desktop, browser
+		 * extension, and embedded variants are exempt.
```
polymutex: This explanation is about methodology, so doesn't belong in feature data.

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Duress resistance features of a wallet.
+ *
+ * Covers the full spectrum from basic lock-screen protection through
+ * full duress-pin-triggered decoy wallets or self-destruct mechanisms.
+ * This feature helps protect users against "wrench attacks" — physical
+ * coercion to hand over funds.
+ *
+ * Only applicable to hardware wallets and mobile app wallets.
+ * Desktop and browser extension wallets are exempt.
```
polymutex: This is about methodology, not feature data. Please remove

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -0,0 +1,219 @@
...
+	icon: '\u{1F527}', // Wrench (references the "wrench attack" threat model)
+	displayName: 'Duress Resistance',
+	wording: {
+		midSentenceName: 'duress resistance',
+	},
+	question: sentence(
+		'Does {{WALLET_NAME}} protect users from being physically coerced into surrendering their funds?',
+	),
+	why: markdown(`
+		A "wrench attack" is when an adversary physically coerces a user into handing over their funds.
```
polymutex: Link to https://xkcd.com/538/

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -0,0 +1,219 @@
...
+		Unlike remote attacks, no amount of cryptographic security can stop an attacker who is standing
+		next to you with a weapon.
+
+		Wallets can mitigate this threat by:
+
+		1. **Lock screen (basic)**: Requiring a PIN, password, or biometric before granting access.
+		   This protects against opportunistic thieves and buys time, but a determined coercer can
+		   watch you unlock it.
+
+		2. **Duress mode (strong)**: A separate duress PIN or passphrase that, when entered,
```
polymutex: "strong" → "stronger"

(not necessarily strong, just stronger)

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -0,0 +1,219 @@
...
+
+		This attribute is only evaluated for **hardware wallets** and **mobile app wallets**, as
+		desktop and browser extension wallets do not meaningfully face this threat model.
+	`),
+	methodology: markdown(`
+		Wallets are rated based on the strongest duress protection they provide:
+
+		1. **Pass**: Implements a duress PIN or passphrase that triggers either:
+			- A decoy wallet (different accounts and balances, plausible deniability), or
+			- A wipe-and-forward mechanism (wipes device, sends funds to a pre-configured address not unilaterally user-controlled).
```
polymutex: Should be part of the feature type enum if listed here too.

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -0,0 +1,219 @@
...
+
+		2. **Partial**: Has a lock screen (PIN, password, biometric, or pattern) but no dedicated duress mode.
+
+		3. **Fail**: No lock screen at all — the wallet is accessible to anyone who picks up the device or opens the app.
+
+		4. **Exempt**: Desktop and browser extension wallets, and embedded wallets.
```
polymutex: The rating scale below already explains this, so this is redundant.

I think you can just replace this with:

```
Wallets are rated based on the strongest duress protection they provide, such as support for a lock screen and duress actions.
```

polymutex (changes requested):

data/software-wallets/ambire.ts
maykelxyz: Ambire is exempt from this since it only has a `BROWSER` variant and does not offer a `MOBILE` version, nor is it a hardware wallet. I’ll rate MetaMask as an alternative.

src/schema/features/security/duress-resistance.ts
maykelxyz: The word `lockdown` has causing me issues.. One fix that I've come up with it this in `utils/grammar.ts`

```ts
lints = lints.filter(
lint => lint.lint_kind_pretty() !== 'Word Choice' || lint.get_problem_text() !== 'lockdown',
)
```

src/schema/features/security/duress-resistance.ts
maykelxyz: Hmm, yeah now that you've mentioned it, that is a possibility.

```ts
export interface DuressMode {
	actions: NonEmptyArray<DuressAction>
}
```

data/software-wallets/ambire.ts
polymutex: I don't think that's relevant; the choice of not having the requirement on browser wallets is an attribute-side methodology choice, not a feature data question. The Ambire browser extension has a lockscreen, so it seems useful to record that here.

But even it were not relevant, this should be some other special string, not `null`, in order to preserve the invariant that `null` only ever means "we don't know". For example, check `features.integration.browser`, which for non-browser-wallets (where collecting this data would make no sense), the special string `NOT_A_BROWSER_WALLET` is used to indicate that instead of `null`.

src/schema/features/security/duress-resistance.ts
```diff
@@ -0,0 +1,130 @@
...
+			return 'wipes the wallet and forwards all funds to a pre-configured safe address'
+	}
+}
+
+/**
+ * Information about a wallet's duress mode.
+ */
+export interface DuressMode {
+	/** One or more actions triggered when duress credentials are entered. */
+	actions: NonEmptyArray<DuressAction>
```
polymutex: I still think Record is better, since it forces exhaustiveness. Same for `mechanisms` above. But OK as array if you strongly prefer.

src/schema/features/security/duress-resistance.ts
polymutex: Yeah, I agree that seems appropriate to make the grammar checker ignore this case.

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+function hasDuressMode(
+	ctx: EvaluationContext<DuressResistanceValue>,
+	duressMode: Supported<WithRef<DuressMode>>,
+): Evaluation<DuressResistanceValue> {
+	const actionNames = commaListFormat(duressMode.actions.map(duressActionName))
+	const actionDescriptions = commaListFormat(
+		duressMode.actions.map(
+			a => `entering a separate duress credential ${duressActionDescription(a)}`,
```
polymutex: Multiple of these would sound awkward:

"MyWallet implements a duress mode: entering a separate duress credential wipes the wallet, preventing the attacker from accessing funds and entering a separate duress credential wipes the wallet and forwards all funds to a pre-configured safe address"

Not very clear. I think this should use `markdownListFormat`, so that it uses an inline sentence in the common single-action case, and a Markdown list in the multi-action case. See the usage of `markdownListFormat` in `src/schema/attributes/privacy/private-transfers.ts` for inspiration.

polymutex (changes requested):

data/software-wallets/ambire.ts
maykelxyz: > I don't think that's relevant; the choice of not having the requirement on browser wallets is an attribute-side methodology choice, not a feature data question. 

I guess that's right. I guess my thinking is that "It's weird filling it up if it will still end up `EXEMPT` anyway. But yeah, your point still stands.

> The Ambire browser extension has a lockscreen, so it seems useful to record that here.

If that's the case, an argument could be made that duress resistance could apply to browser extensions too. Could be a separate discussion.

I think there's two options here moving forward for browser wallets:

1) Create a special string in combining `NOT_A_MOBILE_WALLET` + `NOT_A_HARDWARE_WALLET`,  I've come up with `NOT_A_STANDALONE_WALLET`. Which can then be used for Ambire & browser wallets.

2) Keep as is. Software wallets should still fill up the feature-side of `duress-resistance` even if it means resulting in `EXEMPT`. This way, *if* we ever had the discussion of also considering browser wallets for duress resistance, changes would be minimal.

I'd go with 2.

Thoughts?

polymutex: The idea behind the feature data vs attribute logic is to make point 2 happen. By decoupling them, we can ask wallets everything we _think_ we would ever need from them for rating purposes, and we grant ourselves the ability to change the methodology without necessarily having to go back and repopulate the data. That's why for example it's important in #603 to build a pipeline that grab the wallet's whole manifest, rather than just pulling in the subset of fields we care about today, because it means we can expand our methodology tomorrow to look at other manifest fields, and still have all the data we need without further wallet cooperation.

For a feature field like the one in this PR, there's also a simplicity argument: we just collect what the wallet does and it's easier to explain how to go about collecting the data if the same information is collected regardless of the variant. That's not always possible (e.g. the aforementioned `features.integration.browser`, and hardware wallets are also an example where the possible set of feature data we collect is inherently quite different), but when it is possible (which is the case for duress resistance) then we might as well do it.

As for the methodology question (should duress resistance apply to browser wallets?), IMO I think browser wallets should have basic lockscreen functionality, to protect against the "I left my laptop unlocked in my living room and my roommate that I just had a fight with is looking for revenge"-type of situation. It's not the whole criminal-robs-you-on-the-street situation but you can see how an unlock code is useful for these situations too.

src/schema/attributes/security/duress-resistance.ts
maykelxyz: Fixed. Here's what it looks like now:

## With Duress mode
image

## Without duress mode

image

src/schema/features/security/duress-resistance.ts
maykelxyz: After tinkering with the `completed` template, I resorted now to `Record<DuressAction, boolean>`. It's also more coherent with other features.

data/software-wallets/ambire.ts
maykelxyz: >  "I left my laptop unlocked in my living room and my roommate that I just had a fight with is looking for revenge"-type of situation. It's not the whole criminal-robs-you-on-the-street situation but you can see how an unlock code is useful for these situations too.

I think this can fall under _"What if someone turned evil while you're using your laptop?"_. Using also the roommate reference, the angry roommate can turn it up a notch and turn it into a whole wrench attack, forcing the victim to unlock the wallet because he is in danger. Then the duress resistance attribute could be useful even for browser wallets.

But again, I think this is a bit subjective, that's why I mentioned this could be a separate discussion from this PR. We can prioritize merging and revisit the discussion..

Updated ambire anyway:

```ts
duressResistance: {
basicUnlock: {
	ref: refTodo,
	mechanisms: {
		[BasicUnlockMechanism.PIN]: false,
		[BasicUnlockMechanism.PASSWORD]: true,
		[BasicUnlockMechanism.BIOMETRIC]: false,
		[BasicUnlockMechanism.PATTERN]: false,
	},
},
duressMode: notSupported,
},
```

polymutex (approved):
🚢

polymutex:
Please also update the wallet questionnaires. This is an easy attribute to ask in a questionnaire format, for both software and hardware wallets.

maykelxyz:
Updated the software wallet questionnaire as well as its respective fileverse doc 

Will wait for the CI before merging. Thanks for the review @polymutex :)

PR #633, opened 2026-04-04
Fix: Use a kebab-case wallet id slug for url

### LINKED ISSUE #547: Wallet IDs in URLs should be slugs
Wallet IDs are currently `lowerCamelCase` because that's what's convenient for TypeScript names.
We currently use them in URLs, so for example, the Keycard Shell wallet is `https://beta.walletbeat.eth/keycardShell/`.
This looks off, as URLs are (by convention) case-insensitive. The right URL should be more like `https://beta.walletbeat.eth/keycard-shell`.
This isn't a big problem because most wallet IDs are single-word, so there is no distinction between `lowerCamelCase` and `slug-style`.

Tagged pre-launch because our URLs have to be stable at launch time.

maykelxyz:
# Fix: Use kebab-case wallet IDs for URLs

[Wallet IDs were inconsistently cased (e.g. `keycardShell`, `baseApp`), which caused messy URLs. 
](https://github.com/walletbeat/walletbeat/issues/547)
## Changes

- **Renamed wallet IDs and image files** to kebab-case (`keycardShell` → `keycard-shell`, `uniswapWallet` → `uniswap-wallet`, `baseApp` → `base-app`)
- **Added `allRatedWalletsBySlug` and `isValidWalletSlug`** to `data/wallets.ts` for slug-based wallet lookups
- **Updated navigation hrefs** to use the wallet's `metadata.id` directly
- **Added a wallet integrity test** to enforce that all `metadata.id` values match the kebab-case pattern

## Other Changes
- Renaming the ids were causing cspell test to fail so **renamed `uniswap` and `keycard` to lowercase** to fix.
- **Added `.DS_Store` to `commonExclusions`** so macOS metadata files are ignored in codebase checks

Closes #547


tests/wallet-integrity.test.ts
```diff
@@ -87,6 +88,10 @@ describe('wallets', () => {
 
 	for (const wallet of Object.values(allWallets)) {
 		describe(`wallet ${wallet.metadata.displayName}`, () => {
+			it('has a slug-style metadata.id (lowercase, hyphens only)', () => {
+				expect(wallet.metadata.id).toMatch(/^[a-z][a-z0-9-]*$/)
```
polymutex: Better regex: `^[a-z\d]+(?:-[a-z\d]+)*$`

tests/utils/codebase.ts
```diff
@@ -19,8 +19,8 @@ export const commonExclusions: PathPredicate[] = [
 	// Exclude entries from .gitignore.
 	await GitIgnoredFiles(),
 
-	// Exclude known binary files.
-	/\.(png|pdf|jpg|jpeg|gif|ico)$/i,
+	// Exclude known binary files and macOS metadata files.
+	/\.(png|pdf|jpg|jpeg|gif|ico)$|\.DS_Store$/i,
```
polymutex: `.DS_Store` files should already be ignored by `GitIgnoredFiles` above, since it is already present in the repo's `.gitignore`. If it isn't, please figure out why instead of adding it here.

data/wallets.ts
```diff
@@ -45,6 +45,16 @@ export const allRatedWallets = Object.fromEntries(
 	Object.entries(allWallets).map(([name, wallet]) => [name, rateWallet(wallet)]),
 )
 
+/** All rated wallets keyed by their slug (metadata.id). */
+export const allRatedWalletsBySlug: Record<string, RatedWallet> = Object.fromEntries(
+	Object.entries(allWallets).map(([, wallet]) => [wallet.metadata.id, rateWallet(wallet)]),
+)
+
+/** Check if a string is a valid wallet slug (metadata.id). */
+export function isValidWalletSlug(slug: string): boolean {
```
polymutex: Replace `boolean` with `slug is keyof allRatedWalletsBySlug`

polymutex (changes requested):
I think there are way more links to wallet pages out there. For example, `walletUrl` in `src/views/WalletTable.svelte`.

I think it is a good time to write a `getWalletUrl(wallet: WalletName | RatedWallet, options?: { variant: Variant | null, attributeAnchor: string | null })` function that formats the URL accordingly so that there is only one place to dig for such URLs in the code.

polymutex (approved):
🚢

PR #634, opened 2026-04-05
update Uniswap Wallet listing with licensing, monetization, bug bounty, and privacy policy

ren2140eth:


ren2140eth:
CI notes: 3 remaining failures are pre-existing and not introduced by this PR:

- CONTRIBUTING.md symlink is broken on beta
- features.md needs regeneration after adding CANTINA to BugBountyPlatform enum, could not run pnpm fix locally due to dependency issues; requesting maintainer assistance
- Harper.js grammar false positive on "a Free & Open-Source (FOSS) license" from open-source.ts template, affects all FOSS-licensed wallets

maykelxyz:
Will check the CI issues. Thanks @ren2140eth

maykelxyz:
Further investigation:
> * features.md needs regeneration after adding CANTINA to BugBountyPlatform enum, could not run pnpm fix locally due to dependency issues; requesting maintainer assistance
This should be fixed through a PR I'll create. This is caused by windows path issues 

> * Harper.js grammar false positive on "a Free & Open-Source (FOSS) license" from open-source.ts template, affects all FOSS-licensed wallets

The issue is actually from 

> 'The Uniswap Wallet mobile app is explicitly licensed under GPL 3.0 or later. The browser extension directory in the same monorepo does not contain **a LICENSE file.** Development happens in a private repository and only production-ready code is published.'

~~A temporary fix I can think of is rephrasing it to:~~

~~'The Uniswap Wallet mobile app is explicitly licensed under GPL 3.0 or later. The browser extension directory in the same monorepo has no LICENSE file. Development happens in a private repository and only production-ready code is published.'~~

Edit: 

For more context, Harper flags "a LICENSE file" because it treats LICENSE as an acronym starting with a vowel sound, expecting "an" instead of "a". Wrapping it in backticks (`` `LICENSE` ``) signals that it's a filename, not a word subject to grammar rules.

.claude/worktrees/awesome-jones
polymutex, line 1: I don't know what this `.claude/worktrees/awesome-jones` file is, but it does not belong in the repo. Seems like a Claude Code artifact. Please delete/revert it. I merged #641 which should help with this in the future.

package.json
```diff
@@ -12,7 +12,7 @@
 		"check:lint": "eslint --cache src data tests",
 		"check:lint:quick": "cross-env WALLETBEAT_PRECOMMIT_FAST=true pnpm run check:lint --report-unused-disable-directives-severity=off",
 		"check:spelling": "cspell lint .",
-		"check:spelling:reorder": "tests/cspell/reorder-cspell.sh",
+		"check:spelling:reorder": "tsx tests/cspell/reorder-cspell.ts",
```
polymutex, line 15: Please keep this as-is (the ts script by itself doesn't format properly)

tests/utils/codebase.ts
```diff
@@ -273,10 +303,43 @@ export async function crawlCodebase(options: CodebaseCrawOptions): Promise<void>
...
+						const contents = raw.toString('utf8').replaceAll('\r\n', '\n')
+
+						entryData = { type: CodebaseEntryType.FILE, path: rootRelativePath, raw, contents }
+					}
+				} else {
+					const raw = await concurrencyLimit(() => fs.readFile(fullPath))
+
+					// Normalize CRLF → LF in contents so line-ending checks reflect
+					// what is stored in git, not what git's autocrlf added on checkout.
+					const contents = raw.toString('utf8').replaceAll('\r\n', '\n')
```
polymutex, line 339: This kind of defeats the point of the `codebase-integrity` test, as it would paper over files that accidentally do contain `\r\n` in git.
Not sure how to fix this though...

polymutex (changes requested):

data/software-wallets/uniswap-wallet.ts
```diff
@@ -80,13 +103,34 @@ export const uniswapWallet: SoftwareWallet = {
...
+				availability: BugBountyProgramAvailability.ACTIVE,
+				coverageBreadth: 'FULL_SCOPE' as const,
+				dateStarted: '2024-11-26' as CalendarDate,
+				disclosure: notSupported,
+				legalProtections: notSupported,
+				platform: BugBountyPlatform.CANTINA,
+				rewards: supported({
+					currency: 'USD',
+					maximum: 50000,
+					minimum: 0,
```
maykelxyz, line 130: Their max reward is $15,500,000 as per the cantina link

.claude/worktrees/awesome-jones
maykelxyz: Yep. It's a Claude Code artifact

maykelxyz (changes requested):

ren2140eth:
yep so I think that is specifically related to uni v4 etc. not the wallets. It does look like the max reward is different for the wallets based on this. [image.png]

tests/utils/codebase.ts
maykelxyz:
```ts
  const { stdout } = await exec('git ls-files --eol', { cwd: root })                                                                        
  const crlfInIndex = stdout.split('\n')                                                                                                    
    .filter(line => /^i\/crlf/.test(line))
    .map(line => line.split('\t')[1]?.trim())                                                                                               
    .filter(Boolean)     
```

I was thinking of this solution, lmk what you think

polymutex: Something like that could work but it would be better as a whitelist than a blacklist, i.e. make sure that all lines match an expected format rather than looking for lines that specifically say `i/crlf`.

maykelxyz:
Merged through #674. Thanks for working on this, @ren2140eth! :)

PR #639, opened 2026-04-09
Fix/windows ci

maykelxyz:
# Fix: Windows CI & address uniswap PR issues 

### Changes: 
- Created `function getGitSymlinkPaths(root: string)` that should be use for Windows. - fixes codebase-integrity test issues on Windows
- Normalize path for windows 
  - fixes generating `features.md` to use `/` instead of `\` (Windows) 
- Use tsx version of `reorder-cspell` 
- Add backtick for `.LICENSE` file in Uniswap reference. 
  - Harper flags "a LICENSE file" because it treats LICENSE as an acronym starting with a vowel sound, expecting "an" instead of "a". Wrapping it in backticks (`LICENSE`) signals that it's a filename, not a word subject to grammar rules. 
- Removed Contributing.md from /resources/docs/contribute/ and consolidated everything into the root CONTRIBUTING.md.
  - getGitSymlinkPaths(root: string) resolves the false positives when running tests. However, running pnpm fix (which executes prettier --write .) reformats CONTRIBUTING.md and reintroduces the false positives. It seems that Prettier behaves inconsistently across environments (notably on Windows). Since this originates from a dependency rather than our code, the proposed approach is to handle or normalize this behavior accordingly.

cc: @polymutex @ren2140eth


data/software-wallets/uniswap-wallet.ts
```diff
@@ -68,7 +68,7 @@ export const uniswapWallet: SoftwareWallet = {
 			walletAppLicense: {
 				ref: {
 					explanation:
-						'The Uniswap Wallet mobile app is explicitly licensed under GPL 3.0 or later. The browser extension directory in the same monorepo does not contain a LICENSE file. Development happens in a private repository and only production-ready code is published.',
+						'The Uniswap Wallet mobile app is explicitly licensed under GPL 3.0 or later. The browser extension directory in the same monorepo does not contain a `.LICENSE` file. Development happens in a private repository and only production-ready code is published.',
```
polymutex: Should be `LICENSE` (no dot), as the filename doesn't have a dot in repos that have this file.

Does that make Harper not happy? If so we should add another harper exception.

resources/docs/contribute/wallet-data.md
```diff
@@ -4,7 +4,7 @@ _This guide aims to help you contribute wallet data to Walletbeat. If you'd like
 
 ## Development setup & workflow
 
-Before you can contribute wallet data to Walletbeat, you need to make a copy of the site and its data and be able to run it on your computer so you can make changes. Please read the [`/resources/docs/contribute/CONTRIBUTING.md`](/resources/docs/contribute/CONTRIBUTING.md) guide if you need help with this. The rest of this guide **assumes you have already set up a Walletbeat development environment**.
+Before you can contribute wallet data to Walletbeat, you need to make a copy of the site and its data and be able to run it on your computer so you can make changes. Please read the [`//CONTRIBUTING.md`](//CONTRIBUTING.md) guide if you need help with this. The rest of this guide **assumes you have already set up a Walletbeat development environment**.
```
polymutex: Single slash before `CONTRIBUTING.md`.

polymutex (approved):

ren2140eth:
Any changes i need to make here ?

data/software-wallets/uniswap-wallet.ts
maykelxyz: Ah, no. Removed the dot.

maykelxyz:
> Any changes i need to make here ?

Nope. @ren2140eth :) Thanks for reporting the issues. `pnpm fix` and `pnpm check:all` should now work even on Windows. Feel free to do both and lmk if it works for you.

PR #665, opened 2026-04-25
Add test enforcing commit hashes in ref links to GitHub

### LINKED ISSUE #519: Add test enforcing commit hashes in `ref` links to GitHub

pheobeayo:
fixes #519 
This PR adds a test enforcing that GitHub blob URLs in ref links use a 40-character commit hash rather than a branch name, as requested in #519. The test immediately finds 60 existing violations across 17 wallet data files, which are intentional failures that show why enforcement is needed. A follow-up PR will fix all existing violations.

### DIFF
--- a/tests/github-ref-commit-hash.test.ts
+++ b/tests/github-ref-commit-hash.test.ts
@@ -1,0 +1,128 @@
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+
+const REPO_ROOT = path.resolve(fileURLToPath(import.meta.url), '../..');
+
+
+const DATA_DIR = path.join(REPO_ROOT, 'data');
+
+
+const GITHUB_BLOB_URL_RE =
+  /https:\/\/github\.com\/[^/"'\s]+\/[^/"'\s]+\/blob\/([^/"'\s]+)\/[^"'\s]*/;
+
+const COMMIT_HASH_RE = /^[0-9a-f]{40}$/;
+
+
+// Helpers
+function collectTsFiles(dir: string): string[] {
+  const results: string[] = [];
+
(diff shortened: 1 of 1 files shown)


polymutex (changes requested):
Thanks for this. The test looks good, but as you have noticed, there are already offending links here, so the test is failing. If this PR were merged as-is, this would block all future code pushes until all of the links are fixed.
Please either:
- (a) fix all the links in this same PR (this way the test passes immediately)
- (b) create the follow-up PR now (and we can merge both at the same time, so pushes are not blocked)
- (c) add specific exceptions to the test in this PR to specifically allow the current set of failing URLs, but reject any new non-conformant URLs

pheobeayo:
> Thanks for this. The test looks good, but as you have noticed, there are already offending links here, so the test is failing. If this PR were merged as-is, this would block all future code pushes until all of the links are fixed. Please either:
> 
> * (a) fix all the links in this same PR (this way the test passes immediately)
> * (b) create the follow-up PR now (and we can merge both at the same time, so pushes are not blocked)
> * (c) add specific exceptions to the test in this PR to specifically allow the current set of failing URLs, but reject any new non-conformant URLs


@polymutex I implemented option B and made a follow-up PR at #671 
Kindly review

polymutex:
PR #671 contained this test and the fixes together, so I merged that one, which means this PR is now superseded. Closing.

PR #670, opened 2026-04-27
Add SVG to font generator tooling

polymutex:
Keeps a font full of Walletbeat icons and a CSS file in sync with a directory of SVGs.
Ensures they are monochrome too.


src/styles/wbicons.css
```diff
@@ -0,0 +1,32 @@
...
+.wbicons-discuss::before { content: "\ea03"; }
+.wbicons-faq::before { content: "\ea04"; }
+.wbicons-newsletter::before { content: "\ea05"; }
+.wbicons-repository::before { content: "\ea06"; }
+.wbicons-wallet_browser::before { content: "\ea07"; }
+.wbicons-wallet_desktop::before { content: "\ea08"; }
+.wbicons-wallet_embedded::before { content: "\ea09"; }
+.wbicons-wallet_hardware::before { content: "\ea0a"; }
+.wbicons-wallet_mobile::before { content: "\ea0b"; }
+.wbicons-wallet_software::before { content: "\ea0c"; }
```
darrylyeo: Is the CSS output customizable? Would be cool to incorporate into the existing CSS attributes system and reuse the tooling to take into account this new generated file.

```css
/**
...
*/
[data-icon] {
  font-family: 'wbicons';
  font-style: normal;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;

  &::before {
    content: var(--icon-content);
  }

  &[data-icon~="about"] {
    --icon-content: "\ea01";
  }
  &[data-icon~="account-type"] {
    --icon-content: "\ea02";
  }
  ...
}
```

darrylyeo (approved):

src/tools/icon-font-generator/icon-font-generator-lib.ts
```diff
@@ -0,0 +1,417 @@
...
+const PRESENTATION_ATTR_REGEX = new RegExp(
+	`(?:${SVG_COLOR_ATTRIBUTES.join('|')})\\s*=\\s*["']([^"']*)["']`,
+	'gi',
+)
+
+// Regex to match CSS-in-style declarations like fill: ...; or stroke: ...;
+const STYLE_COLOR_ATTR_REGEX = new RegExp(
+	`(?:${SVG_COLOR_ATTRIBUTES.join('|')})\\s*:\\s*([^;]+);`,
+	'gi',
+)
```
darrylyeo, line 24: Worth looking into whether `svgo` comes with options to reliably convert all the CSS to pure SVG attributes so we have less edge cases like these to check for.

src/styles/wbicons.css
polymutex: Unfortunately not with the svgtofont library, as far as I can tell: https://github.com/jaywcjlove/svgtofont#css
But I added code to manually generate the CSS instead :)

src/tools/icon-font-generator/icon-font-generator-lib.ts
polymutex: Eh, this code is all Qwen and it works :shrug: Doesn't need to be performant, it only runs at font-regeneration time.

PR #671, opened 2026-04-27
fix: replace branch-name GitHub refs with commit hashes

### LINKED ISSUE #519: Add test enforcing commit hashes in `ref` links to GitHub

pheobeayo:
Fixes all 60 existing violations found by the new enforcement test. All GitHub blob URLs in ref links now use 40-char commit hashes instead of branch names like main/master/develop/publish.

Also updates knownValidUrls cache in url-check.test.ts to reflect the new commit-hash URLs.

Closes #519  and completes #665

### DIFF
--- a/tests/github-ref-commit-hash.test.ts
+++ b/tests/github-ref-commit-hash.test.ts
@@ -1,34 +1,14 @@
-import * as fs from 'node:fs'
-import * as path from 'node:path'
-import { fileURLToPath } from 'node:url'
-
 import { describe, expect, it } from 'vitest'
 
-const REPO_ROOT = path.resolve(fileURLToPath(import.meta.url), '../..')
+import { CodebaseEntryType, crawlCodebase, getRepositoryRoot } from './utils/codebase'
 
-const DATA_DIR = path.join(REPO_ROOT, 'data')
+const REPO_ROOT = getRepositoryRoot()
+const DATA_DIR = `${REPO_ROOT}data/`
 
 const GITHUB_BLOB_URL_RE =
 	/https:\/\/github\.com\/[^/"'\s]+\/[^/"'\s]+\/blob\/([^/"'\s]+)\/[^"'\s]*/
 
 const COMMIT_HASH_RE = /^[0-9a-f]{40}$/
-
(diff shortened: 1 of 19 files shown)


tests/github-ref-commit-hash.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+		if (entry.isDirectory()) {
+			results.push(...collectTsFiles(fullPath))
+		} else if (entry.isFile() && entry.name.endsWith('.ts')) {
+			results.push(fullPath)
+		}
+	}
+
+	return results
+}
```
polymutex: Use `crawlCodebase` from `tests/utils/codebase.ts` instead of this function

polymutex: Use `crawlCodebase` from `tests/utils/codebase.ts` instead of this function

tests/github-ref-commit-hash.test.ts
```diff
@@ -0,0 +1,122 @@
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+import { describe, expect, it } from 'vitest'
+
+const REPO_ROOT = path.resolve(fileURLToPath(import.meta.url), '../..')
```
polymutex: Use `getRepositoryRoot()` from `tests/utils/codebase.ts`.

polymutex (approved):
Impressive :)

polymutex (approved):

polymutex:
Thanks again, nice work to fix this in one shot.

PR #674, opened 2026-04-27
Fix: Windows setup & tests

maykelxyz:
## Changes
- Fixed `tests/css-attributes-codegen.test.ts` issue on Windows
  - Same issue as the `codebase.ts` one. 
```diff
- const css = await fs.readFile(cssPath, 'utf8')
- const onDisk = await fs.readFile(dtsPath, 'utf8')
+ const css = (await fs.readFile(cssPath, 'utf8')).replaceAll('\r\n', '\n')
+ const onDisk = (await fs.readFile(dtsPath, 'utf8')).replaceAll('\r\n', '\n')
```
- Refactored crawling from git:
```ts
for (const line of output.split('\n')) {
	if (line === '') {
		continue
	}

	// Format: "i/<eol>  w/<eol>  attr/<attrs>\t<filepath>"
	const match = /^i\/(crlf|lf|mixed|none)\s+w\/(crlf|lf|mixed|none)\s+attr\/\S*\t(.+)$/.exec(
		line,
	)

	if (match === null) {
		throw new Error(`Unexpected git ls-files --eol output: ${line}`)
	}

	if (match[1] === 'crlf') {
		crlfFiles.add(normalizePath(match[3]))
	}
}
```
- Reverted `check:spelling:reorder` to use the shell script:
  `"check:spelling:reorder": "tests/cspell/reorder-cspell.sh"`
  - However, this introduced a Windows issue: npm runs scripts through `cmd.exe` by default, so `.sh` scripts fail unless run through Git Bash.
  - Adding `script-shell=bash` ensures `.sh` scripts run through Bash instead of `cmd.exe`.

## Other Changes
- Merged beta into this branch


polymutex:
I'm not sure the source/target branch is correct here... the diff is huge.

.claude/worktrees/awesome-jones
polymutex, line 1: Delete.

polymutex (changes requested):
Looks fine with `beta` as target branch but now note that this is mixing ren's changes with the windows fixups.
Also still need to delete the Claude submodule.

maykelxyz:
Yep, this was originally pointed to Ren's branch. But I realized late that Ren's branch hasn't merged beta in a long time

.claude/worktrees/awesome-jones
maykelxyz: Deleted

polymutex (approved):

PR #676, opened 2026-04-28
Add release transparency attribute

### LINKED ISSUE #580: Wallet attribute: Release process transparency
A transparency wallet attribute about the release process of the wallet. Included from [1TS security benchmark](https://github.com/walletbeat/walletbeat/discussions/407), but also a good Transparency attribute.

This includes things like:

- Signed builds and/or provenance information of the build
- Use of lockfiles or similar for software supply chain security
- Hermetic build environment, ideally reproducible build

Worth looking at [SLSA](https://slsa.dev/) for standards in this area.

johngrantuk:
Closes #580 

See original issue for feature discussion. Although not all added features are used in current attribute eval logic I thought it was worth adding them for possible future use.

## Release process attribute: evaluation logic

- Defines a **verifiable pass/partial/fail** transparency attribute for software wallet release safety, with **hardware wallets exempted** (will be tracked separately in future).
- Evaluates **4 binary signals** from `features.transparency.releaseTransparency`:
  - public changelog
  - reproducible or hermetic builds
  - artifact signing
  - dependency locking
- Applies a **source-visibility gate** to build-related checks: reproducible/hermetic builds only count if source is publicly visible (`isSourcePubliclyVisible(...)`), to ensure independent verification is possible.
- Returns **`unrated` early** when required signal data is missing (`null`) for any of the four checks (including both build fields missing).
- Counts supported signals and maps outcomes as:
  - `4/4` → **PASS** (`Strong release process`)
  - `1-3/4` → **PARTIAL** with score `signalCount / 4`
  - `0/4` → **FAIL** (`No release process signals`)
- Uses `pickWorstRating` as the aggregation strategy across evaluations.


src/schema/features/transparency/release-transparency.ts
```diff
@@ -0,0 +1,81 @@
...
+	| {
+			artifactsSigned: NotSupported
+			artifactSigner: null
+			signaturePublication: null
+	  }
+	| {
+			artifactsSigned: Supported<WithRef<{}>>
+			artifactSigner: ArtifactSignerType | null
+			signaturePublication: SignaturePublicationType | null
+	  }
```
polymutex: How about:

```typescript
export type ArtifactSigning = Support<WithRef<{
	signer: ArtifactSignerType | null
	publication: SignaturePublicationType | null
}>>
```

src/schema/features/transparency/release-transparency.ts
```diff
@@ -0,0 +1,81 @@
...
+ * workflows for the wallet.
+ */
+export type DependencyVulnerabilityScanning = Support<WithRef<{}>>
+
+/**
+ * Observable repository-level change controls for the wallet's source
+ * repository. Each sub-field is independently researchable and may be `null`
+ * until determined.
+ */
+export type RepositoryChangeControls = Nullable<{
```
polymutex: Wrap with WithRef

src/schema/features.ts
```diff
@@ -557,6 +586,36 @@ export function resolveFeatures(
...
+					features => features.transparency.releaseTransparency.hasPublicChangelog,
+				),
+				hermeticBuilds: baseFeat(
+					'transparency.releaseTransparency.hermeticBuilds',
+					features => features.transparency.releaseTransparency.hermeticBuilds,
+				),
+				repositoryChangeControls: baseFeat(
+					'transparency.releaseTransparency.repositoryChangeControls',
+					features => features.transparency.releaseTransparency.repositoryChangeControls,
+				),
```
polymutex: Need to call `nullable(...)` here

.cspell.json
```diff
@@ -360,6 +361,7 @@
 		"sidepanel",
 		"signup",
 		"sigri",
+		"SIGSTORE",
```
polymutex: Please put these two new words as lowercase, as I'm sure we'll want to write them as non-enum-ish normal words sometime down the line.

src/schema/features/transparency/license.ts
```diff
@@ -343,6 +343,31 @@ export function licenseSourceIsVisible(license: License): license is SourceAvail
 	return sourceAvailableLicense.is(license)
 }
 
+/**
+ * Returns true if the wallet's source code is publicly visible,
+ * based on its resolved licensing. Useful for gating signals that
+ * require source access for external verification (e.g. reproducible builds).
+ */
+export function isSourcePubliclyVisible(licensing: ResolvedWalletLicensing): boolean {
```
polymutex: Please make this return `boolean | null`, in order be able to distinguish the case of the source not being public vs the case of the license information not being present. Whether to treat these two cases similarly seems like a decision that should be up to the caller of the function.

data/software-wallets/ambire.ts
```diff
@@ -745,6 +750,86 @@ export const ambire: SoftwareWallet = {
...
+				}),
+				dependencyVulnerabilityScanning: notSupported,
+				hasPublicChangelog: supported({
+					ref: 'https://github.com/AmbireTech/extension/releases',
+				}),
+				hermeticBuilds: notSupportedWithRef({
+					ref: [
+						{
+							explanation:
+								'The release build workflow updates git submodules during build execution, requiring network access rather than using a fully pre-fetched offline input set.',
```
polymutex, line 781: Nice catch

data/software-wallets/ambire.ts
```diff
@@ -745,6 +750,86 @@ export const ambire: SoftwareWallet = {
...
+							url: 'https://github.com/AmbireTech/extension/blob/main/.github/workflows/build-extension-gecko.yml',
+						},
+					],
+				}),
+				repositoryChangeControls: null,
+				reproducibleBuilds: supported({
+					ref: [
+						{
+							explanation:
+								'The repository README documents local installation and build commands (`yarn setup`, `yarn build:extensions`, platform build commands), indicating independent builders can perform builds themselves.',
```
polymutex: Build reproducibility means a byte-for-byte equal binary output, so that a user can ascertain that the build they are downloading actually contains the exact same source files as what the source code repo contains. The presence of instructions on how to run the build isn't a guarantee of this. Best way to test for this would be to run these instructions in two separately-configured VMs or similar, and see if the two builds are exactly the same. In the absence of that, leave `null`, or maybe assume `notSupported` (very few build processes are actually reproducible unfortunately...)

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+import type { WalletMetadata } from '@/schema/wallet'
+import { WalletType } from '@/schema/wallet-types'
+import { mdParagraph, paragraph, sentence } from '@/types/content'
+import { commaListFormat } from '@/types/utils/text'
+
+import { exempt, pickWorstRating, unrated } from '../common'
+
+function describeSupportedSignals(signals: string[]): string {
+	return commaListFormat(signals)
+}
```
polymutex: Doesn't seem like a very useful helper function if it's just calling `commaListFormat`...

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+
+function pass(ctx: EvaluationContext, supportedSignals: string[]): Evaluation {
+	return ctx.build({
+		outcome: {
+			id: 'pass',
+			rating: Rating.PASS,
+			displayName: 'Strong release process',
```
polymutex: "Transparent" is a more descriptive word than "Strong" here

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+
+			- **Public changelog**: publish release notes or a changelog for each release.
+			- **Reproducible or hermetic builds**: ensure independent parties can rebuild
+			  the same source to obtain a byte-for-byte identical artifact, or that the
+			  build can run fully offline from a pre-fetched, integrity-verified input set.
+			- **Artifact signing**: sign release artifacts so users can verify they have not
+			  been tampered with.
+			- **Dependency locking**: use a lockfile (or equivalent) to pin all dependencies
+			  to known versions.
+		`),
```
polymutex, line 257: Should filter this list to the ones that the wallet needs to actually do

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
```
polymutex, line 309: This and the function above could maybe use a helper function that builds the list, given a list of supported (or unsupported) signals.

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+		Without these signals, a compromised or tampered release may go undetected.
+	`),
+	methodology: mdParagraph(`
+		Four independently-verifiable signals are assessed, each binary (present or absent):
+
+		1. **Public changelog**: the wallet publishes release notes or a changelog.
+		2. **Reproducible or hermetic builds**: independent parties can verify the build
+		   output matches the source, or the build can run fully offline.
+		   This signal only counts when the wallet's source code is publicly visible,
+		   since external verification requires source access.
```
polymutex: This sounds like if the wallet is closed-source, it would get a pass on this. It won't, right? So I think you can replace "This signal only counts when the wallet's source code is publicly visible" with "This requires the wallet's source code to be publicly visible".

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+		3. **Artifact signing**: release artifacts are cryptographically signed.
```
polymutex: and these signatures are published

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+		if (rt.reproducibleBuilds === null && rt.hermeticBuilds === null) {
+			return unrated(ctx)
+		}
+
+		const reproducibleSupported =
+			rt.reproducibleBuilds !== null && isSupported(rt.reproducibleBuilds)
+		const hermeticSupported = rt.hermeticBuilds !== null && isSupported(rt.hermeticBuilds)
+
+		// Source-visibility cap: reproducibility cannot be externally verified
+		// without public source access, so the signal does not count.
```
polymutex: Same comment regarding wording here.

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,277 @@
...
+
+		1. **Public changelog**: the wallet publishes release notes or a changelog.
+		2. **Reproducible or hermetic builds**: independent parties can verify the build
+		   output matches the source, or the build can run fully offline.
+		   This signal only counts when the wallet's source code is publicly visible,
+		   since external verification requires source access.
+		3. **Artifact signing**: release artifacts are cryptographically signed.
+		4. **Dependency locking**: a lockfile or equivalent pins all dependency versions.
+
+		All four signals present earns a **pass**. Any partial coverage earns a **partial**.
```
polymutex: I like the simplicity, but IMO these signals don't all have the same value. A changelog is the absolute bare minimum in my view, and something that having a mobile version kind of requires anyway, so I worry that all wallets would clear this bar just from having a changelog and not being motivated to work on any of the other things because they would need to clear *all* of the rest at once before going to "pass" territory.

In my opinion, a stack ranking would be (from least to most meaningful): `changelog < dep locking < artifact signing < reproducible builds`. Seems to me like we should draw the "fail" vs "partial" line somewhere in between this. For example, we could say "changelog = fail, changelog + dep locking = partial, anything other than changelog and dep locking = partial"
or just "anything other than changelog or dep locking = partial"

It's not super principled, so if you have a better idea here we could use that... for example we could try to map this to SLSA and pick some levels on the ladder to map the boundaries onto.

polymutex (changes requested):
Thanks again for working on this, looking pretty good.

src/schema/attributes/transparency/release-process.ts
johngrantuk: Updated so the logic is grouped into two buckets:
- **Process transparency**: changelog + dependency locking  
  - `processPass` only when both are supported
- **Supply-chain integrity**: artifact signing + reproducible/hermetic builds (with public source)  
  - `supplyLevel = fail | partial | pass` based on 0 / 1 / 2 supported signals
### Classification matrix
Ratings are selected by `(processPass, supplyLevel)` (not raw signal count):
- `false + fail` → `FAIL`
- `true + fail` → `PARTIAL` (`partial_process_pass_supply_fail`, 0.40)
- `false + partial` → `PARTIAL` (`partial_process_fail_supply_partial`, 0.60)
- `false + pass` → `PARTIAL` (`partial_process_fail_supply_pass`, 0.65)
- `true + partial` → `PARTIAL` (`partial_process_pass_supply_partial`, 0.75)
- `true + pass` → `PASS`


Also it uses strict unknown handling and returns `UNRATED` when required inputs are missing, including unknown source visibility.  
Rationale: avoid inferring weaker classifications from incomplete supply-chain verifiability context.

src/schema/features/transparency/license.ts
```diff
@@ -343,6 +343,32 @@ export function licenseSourceIsVisible(license: License): license is SourceAvail
...
+export function isSourcePubliclyVisible(licensing: ResolvedWalletLicensing): boolean | null {
+	if (licensing === null || licensing.walletAppLicense === null) {
+		return null
+	}
+
+	switch (licensing.type) {
+		case LicensingType.SINGLE_WALLET_REPO_AND_LICENSE:
+			return licenseSourceIsVisible(licensing.walletAppLicense.license)
+		case LicensingType.SEPARATE_CORE_CODE_LICENSE_VS_WALLET_CODE_LICENSE:
+			if (licensing.coreLicense === null) {
```
polymutex: Should check either license for null-ness.

src/schema/features/transparency/release-transparency.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Observable repository-level change controls for the wallet's source
+ * repository. Sub-fields may be `null` in source data while research is in
+ * progress.
+ *
+ * During resolution, this feature is normalized all-or-nothing: if any
+ * sub-field remains unknown, the resolved value becomes `null`.
+ */
+export type RepositoryChangeControls = Nullable<
```
polymutex: You should put the `Nullable` part of this out of this type definition, and instead into the fields in `src/schema/features.ts`. Specifically, it is only `Nullable` in `WalletBaseFeature`, but not in `ResolvedFeatures`. Look at `transparency.operationFees` and the `BasicOperationFees` type for an example.
The advantage of doing it this way over the current PR is that this will let the attribute code not have to worry about the partially-`null` case.

src/schema/features/transparency/release-transparency.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Artifact signing information for a wallet variant.
+ *
+ * Uses the common Support shape and carries signer/publication details
+ * when signing is supported. Metadata may still be `null` if unknown.
+ */
+export type ArtifactSigningDetails = WithRef<{
+	signer: ArtifactSignerType | null
+	publication: SignaturePublicationType | null
+}>
```
polymutex: Seems like this could be wrapped in `Nullable` too (although as per other comment, do that in `src/schema/features.ts` in `WalletBaseFeatures` only, not in this type definition.

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,494 @@
...
+	},
+	exempted: (ctx: EvaluationContext, _metadata: WalletMetadata) => {
+		if (ctx.features.type === WalletType.HARDWARE) {
+			return exempt(ctx, sentence('Release process is tracked separately for hardware wallets.'))
+		}
+
+		return null
+	},
+	evaluate: (ctx: EvaluationContext): Evaluation => {
+		ctx.setVerifiability(Verifiability.VERIFIABLE)
```
polymutex: Not quite; I think this one is actually quite difficult to verify and you may need to implement some custom logic here.
For example, a few lines earlier this states:

```typescript
	// Build-integrity claims only count when source is publicly visible,
	// because external reproducibility checks require source access.
```

This should be part of the condition that controls the `outcome.verifiability` field of the evaluation result. The tricky part is that it comes into play _only in the case_ of build integrity claims. This is unlike other attributes where verifiability is more of an all-or-nothing matter.

Specifically, here, you need source visibility to verify build reproducibility, build hermeticity, dependency locking, and `RepositoryChangeControls`. But you don't need it to verify build artifact signing or the presence of a public changelog. So the verifiability bit should be set accordingly.

I think the easiest way to do that is to replace this `ctx.setVerifiability` line with a variable like `let verifiabilityNeedsSourceCodeVisibility = false` here instead, which later blocks in the function can modify as soon as verifiability requires source code visibility. Then at the end of the evaluation function, you can put `ctx.setVerifiability(verifiabilityNeedsSourceCodeVisibility ? verifiabilityRequiresSourceCodeAccess({ coreOnlyIsSufficient: false }) : Verifiability.VERIFIABLE)`, and then change the

```typescript
		if (sourceVisible === null) {
			return unrated(ctx)
		}
```

block to be closer to the end of the function into

```typescript
		if (verifiabilityNeedsSourceCodeVisibility && sourceVisible === null) {
			return unrated(ctx)
		}
```

And then finally the `&& sourceVisible` part of the checks above, as the lack of verifiability will now be expressed by the `outcome.verifiability` field rather than being built into the `outcome.rating` field.

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,494 @@
...
+		Four binary signals are assessed, grouped into two categories:
+
+		**Process transparency** (baseline hygiene):
+		1. **Public changelog**: the wallet publishes release notes or a changelog.
+		2. **Dependency locking**: a lockfile or equivalent pins all dependency versions.
+
+		**Supply-chain integrity** (high-trust):
+		3. **Artifact signing**: release artifacts are cryptographically signed and these signatures are published.
+		4. **Reproducible or hermetic builds**: independent parties can verify that build output matches
+		   source, or the build can run fully offline. This requires public source code.
```
polymutex: These groups make sense, but their current labels don't, given that dependency locking is definitely more of a supply chain integrity measure than a process one.
Maybe just label them group 1 and group 2 :) Or "basic" and "advanced"? Up to you.

src/schema/attributes/transparency/release-process.ts
```diff
@@ -0,0 +1,494 @@
...
+					'The wallet has a changelog, dependency locking, and artifact signing, but no reproducible or hermetic builds.',
+				),
+				partialProcessPassSupplyPartial(
+					EvaluationContext.forTest(() => releaseProcess),
+					['public changelog', 'dependency locking', 'artifact signing'],
+					{ signing: true, builds: false },
+				),
+			),
+		],
+		fail: exampleRating(
```
polymutex: Please add a case for a wallet having a changelog only, as that'd be perhaps a more common case than the case of a wallet having nothing.

polymutex (changes requested):
I'm thinking that maybe we should also delete the `src/schema/attributes/transparency/source-visibility.ts` attribute. It now seems entirely obsoleted by this one. Can be a separate PR.

src/schema/features/transparency/release-transparency.ts
johngrantuk: Thanks for the explanation. I changed and also added some of the detail to the Cursor rule files and AGENTS.md as it might help in the future.

src/schema/attributes/transparency/release-process.ts
johngrantuk: I think I've managed to cover the requested changes but will be worth a double check.

src/schema/features/transparency/release-transparency.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Observable repository-level change controls for the wallet's source
+ * repository.
+ *
+ * In wallet feature data, use `Nullable<RepositoryChangeControls>` (see
+ * `WalletBaseFeatures`) so sub-fields may be `null` while research is in
+ * progress.
+ *
+ * During resolution, this feature is normalized all-or-nothing: if any
+ * sub-field remains unknown, the resolved value becomes `null`.
```
polymutex: That's a fairly remote implementation detail, I'm not sure it's worth explaining here specifically.

polymutex (approved):
Looks good :) Awesome

polymutex:
Yet another epic PR. Do you want to write a social media thread about it? e.g. see #731

Issue #619, opened 2026-03-28
Add support for repo-file-based references


polymutex:
Related but not completely identical to #551 (image-based references), we should also support references based on files that Walletbeat has as part of its own repository.

This would be useful for references to questionnaire answers from the wallet teams via the intake form.

References would be of non-URL type and would contain a repo-relative filename under `public/`, with a unit test verifying its existence in the repo. When rendered in the UI, it would link to the file using a document-root-relative URL.

Khemmie-Ray:
Can I work on this issue?

polymutex:
Yes.

PR #694, opened 2026-05-04
Add support for repo-file-based references

Khemmie-Ray:
This closes issue #619

### DIFF
--- a/src/schema/reference.ts
+++ b/src/schema/reference.ts
@@ -34,7 +34,10 @@ import {
  */
 export interface LooseReference {
 	/** The URL(s) the reference is about. */
-	url: NonEmptyArray<Url> | Url
+	url?: NonEmptyArray<Url> | Url
+
+	/** The repo-relative filename under public/ (New Field) */
+  file?: string
 
 	/** The text of the link that goes to `url`; defaults to the domain name of `url`. */
 	label?: string
@@ -48,18 +51,23 @@ export interface LooseReference {
 
 /** Type predicate for LooseReference. */
 export function isLooseReference(x: unknown): x is LooseReference {
-	return (
-		x !== undefined &&
-		x !== null &&
-		typeof x === 'object' &&
-		Object.hasOwn(x, 'url') &&
-		((url: unknown) =>
-			isUrl(url) ||
-			(Array.isArray(url) && isNonEmptyArray(url) && url.every(u => isUrl(u) || isLabeledUrl(u))))(
-			// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we just checked the field exists.
-			(x as unknown as { url: unknown }).url,
-		)
-	)
+  if (x === undefined || x === null || typeof x !== 'object') {
+    return false
+  }
+
+  if (Object.hasOwn(x, 'file') && typeof (x as { file: unknown }).file === 'string') {
+    return true
+  }
+
+  return (
+    Object.hasOwn(x, 'url') &&
+    ((url: unknown) =>
+      isUrl(url) ||
+      (Array.isArray(url) && isNonEmptyArray(url) && url.every(u => isUrl(u) || isLabeledUrl(u))))(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      (x as unknown as { url: unknown }).url,
+    )
+  )
 }
 
 /**
@@ -138,11 +146,9 @@ export function toFullyQualified(
 
 	if (Array.isArray(reference)) {
 		const qualified: FullyQualifiedReference[] = []
-
 		for (const ref of reference) {
 			qualified.push(...toFullyQualified(ref))
 		}
-
 		return mergeRefs(...qualified)
 	}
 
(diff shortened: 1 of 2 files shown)


src/schema/reference.ts
```diff
@@ -34,7 +34,10 @@ import {
  */
 export interface LooseReference {
 	/** The URL(s) the reference is about. */
-	url: NonEmptyArray<Url> | Url
+	url?: NonEmptyArray<Url> | Url
+
+	/** The repo-relative filename under public/ (New Field) */
+  file?: string
```
polymutex: Please run `pnpm fix`.

src/schema/reference.ts
```diff
@@ -168,12 +175,46 @@ export function toFullyQualified(
...
+		Object.hasOwn(reference, 'file') &&
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we verify the "file" field exists.
+		typeof (reference as { file: unknown }).file === 'string'
+	) {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we just verified the "file" field exists and is a string.
+		const filePath = (reference as { file: string }).file
+
+		if (!filePath.startsWith('public/')) {
+			throw new Error(
+				`RepoFileReference path must be a repo-relative path under public/, got: "${filePath}"`,
```
polymutex: `RepoFileReference` isn't really a technical term that exists outside of this. I think this message can just say "File path-based references must be a repository-relative path under ..."

src/schema/reference.ts
```diff
@@ -48,18 +51,23 @@ export interface LooseReference {
...
-			(Array.isArray(url) && isNonEmptyArray(url) && url.every(u => isUrl(u) || isLabeledUrl(u))))(
-			// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we just checked the field exists.
-			(x as unknown as { url: unknown }).url,
-		)
-	)
+  if (x === undefined || x === null || typeof x !== 'object') {
+    return false
+  }
+
+  if (Object.hasOwn(x, 'file') && typeof (x as { file: unknown }).file === 'string') {
```
polymutex: This should also check that the `file` strings starts with `public/`

src/schema/reference.ts
```diff
@@ -194,6 +235,10 @@ export function toFullyQualified(
 		]
 	}
 
+	if (reference.url === undefined) {
+		return []
```
polymutex: Under what circumstances would this block execute?

tests/file-references.test.ts
```diff
@@ -0,0 +1,113 @@
...
+					for (const match of matches) {
+						fileReferences.push({
+							filePath: match[1],
+							sourceFile: fullPath,
+						})
+					}
+				}
+			}
+
+			scanDir(dataDir)
```
polymutex: Let's not parse TypeScript with regular expressions :)
Please check the `findRefs` function here: https://github.com/walletbeat/walletbeat/blob/83757271a72ea633d435f86628b89c8486475341/tests/url-check.test.ts#L1920-L1945
(The test this was part of was deleted in 08e680960cb78bbdf561efe98aaab8db0d3b0119 but the logic for finding refs is good still)

tests/file-references.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+			for (const { filePath, sourceFile } of fileReferences) {
```
polymutex: Please add a check that `filePath` starts with the string `public/` and does not contain any path-traversal components such as `.` or `..`, and produce an error earlier in this loop if that's the case.

polymutex (changes requested):
Thanks for the support and test :)

Khemmie-Ray:
Alright, I will work on that

src/schema/reference.ts
```diff
@@ -34,7 +34,10 @@ import {
  */
 export interface LooseReference {
 	/** The URL(s) the reference is about. */
-	url: NonEmptyArray<Url> | Url
+	url?: NonEmptyArray<Url> | Url
+
+	/** The repo-relative filename under public/ (New Field) */
+	file?: string
```
polymutex: To indicate that either `url` or `file` may be specified, you can use:

```typescript
export type LooseReference = (
{
	/** The URL(s) the reference is about. */
	url: NonEmptyArray<Url> | Url
} | {
	/** The repo-relative filename under public/ (New Field) */
	file: string
}) & {
	/** The text of the link that goes to `url`; defaults to the domain name of `url`. */
	label?: string

	/** A human-readable string that explains what the reference is about. */
	explanation?: string

	/** The date the reference was last retrieved. */
	lastRetrieved?: CalendarDate
}
```

src/schema/reference.ts
```diff
@@ -48,15 +51,25 @@ export interface LooseReference {
...
+
 	return (
-		x !== undefined &&
-		x !== null &&
-		typeof x === 'object' &&
 		Object.hasOwn(x, 'url') &&
 		((url: unknown) =>
 			isUrl(url) ||
 			(Array.isArray(url) && isNonEmptyArray(url) && url.every(u => isUrl(u) || isLabeledUrl(u))))(
-			// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- Safe because we just checked the field exists.
```
polymutex, line 59: Please preserve the explanation here

src/schema/reference.ts
polymutex: bump. I think this block would become no longer necessary if you use the union type as suggested in my other comment.

polymutex (changes requested):
All looks good, just some typing change

Khemmie-Ray:
fixed

polymutex (approved):

polymutex:
Looks good but need to fix the type errors:

```
/home/runner/work/walletbeat/walletbeat/src/schema/reference.ts
Error:    63:11  error  Unsafe type assertion: type '{ file: unknown; }' is more narrow than the original type   @typescript-eslint/no-unsafe-type-assertion
Error:   214:12  error  Unsafe type assertion: type '{ label: unknown; }' is more narrow than the original type  @typescript-eslint/no-unsafe-type-assertion
Error:   215:8   error  Unsafe type assertion: type '{ label: string; }' is more narrow than the original type   @typescript-eslint/no-unsafe-type-assertion
```

Khemmie-Ray:
Done.

polymutex (approved):

polymutex:
Thanks!

PR #699, opened 2026-05-06
Add Cryptograph hardware wallet listing

perpetua-engineering:
Adds Cryptograph, a self-custodial wallet for Apple Watch by Perpetua Labs LLC, to the hardware wallets directory. Cryptograph uses the Apple Watch Secure Enclave as its trust boundary rather than shipping new hardware. The watch generates the mnemonic, the encryption key that secures it is held in the SE and never leaves hardware, and only the watch can sign. The iPhone composes transactions but holds no keys and cannot sign. 

Most hardware wallets use the seed phrase itself as the recovery material. Cryptograph's primary recovery is an encrypted printed QR sheet or a steganographic JPEG, decryptable only with the user's password; keys never leave the watch in cleartext. Seed-phrase visibility on the watch is preserved for BIP-39 portability.

All key-handling flows begin in the watchOS UI, so even in the event of a future WCSession/Bluetooth exploit, no watchOS code exists that exports key material in cleartext via remote RPC.

## What's in this PR

**Files:**

- `data/entities/perpetua-labs.ts` — wallet developer entity
- `data/entities/{reown,solana-foundation,tron-foundation,xrpl-labs}.ts` — upstream entities referenced in dataCollection
- `data/contributors/perpetua.ts` — pseudonymous contributor file with affiliation disclosure
- `data/hardware-wallets/cryptograph.ts` — wallet listing, all four attribute groups populated
- `data/hardware-wallets.ts` — index registration
- `public/images/wallets/cryptograph.png` — 1024×1024 app icon
- `.cspell.json` — three new dictionary words

**One structural schema change:** `scamAlerts` moved from `WalletSoftwareFeatures` to `WalletBaseFeatures` in `src/schema/features.ts`. Cryptograph implements scam-prevention entirely on-device. Both the verified-contracts allowlist (with lockdown mode for unknown contracts) and the `eth-phishing-detect` bundled domain list are compiled into the IPA at build time and cryptographically signed via Apple's distribution chain. The watch makes scam-prevention decisions without contacting any servers, the iPhone, or any third party. The existing schema did not let hardware wallets populate this. Adds `scamAlerts: null` to pre-existing hardware wallets.

## Working through Walletbeat's rubric drove eight feature ships

The act of mapping Cryptograph to the schema surfaced specific gaps that became real defense improvements. 

| Walletbeat criterion that surfaced the gap | Feature shipped (internal CR ticket) |
|---|---|
| `sendTransactionWarning.newRecipientWarning` | First-time-recipient acknowledgement gate (CR-1227) |
| `transactionLegibility.calldataDecoded` Safe-nested benchmarks | Recursive Gnosis Safe `execTransaction` and `multiSend` decoding (CR-1228) |
| `messageSigningLegibility` domain / message / Safe hash views | EIP-712 hash exposure in signing Details disclosure (CR-1229) |
| `scamAlerts.scamUrlWarning` | Bundled `eth-phishing-detect` checked at WalletConnect connection time (CR-1230) |
| `bugBountyProgram` with rewards + Safe Harbor | Dedicated bug bounty page with disclose.io Safe Harbor, $100 to $5,000 USD bootstrap-tier rewards, $20k annual cap, KYC + OFAC screening (CR-1231) |
| `hardwareVerification` beyond StoreKit AppTransaction | DCAppAttestService runtime device attestation (CR-1232) |
| `detailsDisplayed.nonce` and `dataExtraction.HASHES` | Transaction nonce and signing digest in EVM approval Details disclosure (CR-1236) |
| `dataExtraction.QRCODE` | QR code export of pending transaction for independent verification (CR-1237) |

## Per-attribute-group summary

**Security.** Scam prevention populated across all three legs (contract layer, URL layer, send warnings). Transaction legibility full ABI decoder coverage including the two Safe-nested benchmarks. Bug bounty self-hosted at [`cryptograph.watch/bug-bounty`](https://cryptograph.watch/bug-bounty) with disclose.io-derived Safe Harbor and bootstrap-tier rewards. Duress resistance populated. Firmware FAIL across all four sub-fields (the schema is built around traditional hardware-wallet firmware images, which App Store distribution structurally does not fit; the `details` text lists the iOS-distribution-model mitigations the schema does not reward, including per-update App Store review, FairPlay launch-time signature verification, and retroactive kill-switch on installed copies). User safety populated with on-device curated decoder, simulation, and risk analysis: 9 PASS, 3 PARTIAL, 4 FAIL across 16 sub-fields.

**Privacy.** Analytics (`usage` and `crashReports`) `notSupported`. Cryptograph does not integrate any third-party analytics or crash-reporting SDK. Hardware privacy PASS across all three sub-fields. Transaction privacy populated with default `PUBLIC` and the four EVM privacy techs `notSupported` (Cryptograph supports shielded Zcash, but the `PrivateTransferTechnology` enum is EVM-side only). `dataCollection` populated. Cryptograph's apiproxy backend abstracts most external chain-data calls; upstream providers (Alchemy, CoinGecko, Jupiter, mempool.space, CryptoCompare) see only Cloudflare Worker IPs and never the client IP. Three direct chain RPCs (Solana, Tron, XRPL) do see client IPs and are listed as separate entities.

**Self-Sovereignty.** Interoperability PASS on both sub-fields. Cryptograph generates a standard BIP-39 mnemonic, BIP-32 derived per BIP-44, viewable on the watch via Settings, and portable to any compatible wallet. Standard ECDSA / EdDSA signatures with no Cryptograph-specific markers.

**Transparency.** Maintenance PARTIAL (Apple Watch durability and AppleCare+ as the manufacturer-side baseline; MTBF documentation FAIL because Apple does not publish formal MTBF figures). Operation fees populated. Reputation PARTIAL (originalProduct, availability, warrantySupportRisk, bugBounty all PASS; disclosureHistory PARTIAL because the bug bounty program is recently launched and has no public CVEs yet).

## Three deliberate `null` cells

Each is documented in the data file with a code comment.

1. **`accountRecovery: null`.** Schema models guardian-based ("social") recovery. Cryptograph uses encrypted-backup-with-password (Recovery Sheet QR or Photo Backup steganography). Same posture as Ledger, Trezor, and Keystone, all of which are also `null`. GridPlus alone sets `{ guardianRecovery: notSupported }`, which produces a misleading FAIL message; that path is not the right reading for our recovery model.

2. **`secureElement: null`.** Apple's Secure Enclave is not publicly Common Criteria-certified at any of the EAL levels in the existing enum, and is not PCI-certified. Setting any current value would misrepresent Apple's published posture. Bitbox, Firefly, and ImKey are also `null` here.

3. **`publicSecurityAudits: null`.** Cryptograph has internal security audit work but no public audit report we can reference yet.

## Four populates worth your specific input

These are calls where reasonable readings might differ. Our reading is below; we'd value your read in particular.

1. **`supplyChainFactory: PASS` with Apple as the manufacturer.** The schema's sub-fields (factoryOpsecDocs, factoryOpsecAudit, tamperResistance, hardwareVerification, etc.) ask substantive questions about manufacturing security. For Cryptograph, the device users wear is an Apple Watch, and Apple's published posture answers each sub-field affirmatively. The `details` text frames each sub-field in hardware-wallet-equivalent language (Trezor's PIN-wipe analog for watchOS lockout escalation; Ledger's manufacturer-cert-chain analog for DCAppAttestService) and references Apple Platform Security guide, supplier responsibility audits, Apple Security Bounty up to $1M with public Secure Enclave payouts, and the Apple Security Research Device program. This is the cell where "wallet vendor" and "device manufacturer" diverge most. cc @PatrickAlphaC @nconsigny.

2. **`hardwareWalletManufactureType: FACTORY_MADE`.** Closest existing enum value. Apple manufactures the watch; Perpetua Labs does not. Happy to contribute a new enum value (`THIRD_PARTY_FACTORY_MADE` or similar) as a follow-up PR if you'd prefer that mapping.

3. **`dataCollection` populated with apiproxy-level abstraction.** Cryptograph's apiproxy mediates most external chain-data calls and does not forward client IPs to upstream providers (Alchemy, CoinGecko, Jupiter, mempool.space, CryptoCompare). The populate lists Perpetua Labs (apiproxy operator) as the data-receiving entity for those flows rather than the upstream providers. Three chains call directly (Solana, Tron, XRPL) and are listed as separate entities. Reviewer judgment: is the apiproxy-as-shielding mapping the right read, or should the upstream providers be listed anyway?

4. **`firmware: FAIL` across all sub-fields.** Honest reading. The four sub-fields are built around traditional hardware-wallet firmware images that the wallet vendor controls. App Store-distributed iOS apps do not structurally fit. We populated FAIL rather than `null` so the headline rating reflects what the rubric measures, with the `details` text listing the iOS-distribution-model mitigations the schema does not reward.

## Listing category

Listed under `data/hardware-wallets/` because the security model is hardware isolation: the encryption key that secures the mnemonic is held in the Apple Watch Secure Enclave and never leaves hardware. See [`cryptograph.watch/security`](https://cryptograph.watch/security).

## Maintainer disclosure

Submitted by a maintainer of Cryptograph contributing data about our own wallet. The contributor file (`data/contributors/perpetua.ts`) is pseudonymous per project policy and discloses `EMPLOYEE` affiliation with `hasEquity: true`. Perpetua Labs LLC is the wallet developer.

## Why we built this

The maintainer submitting this listing was a long-time Ledger user until receiving a physical scam letter at their home address, a likely downstream of the 2020 customer-database breach and similar public-list exposures of crypto holders. The realization that *adversaries know I have crypto and where I live* became the design constraint. Cryptograph's premise is that hardware isolation does not have to come with the targeting risks of a dedicated, shipped, registered-to-an-address device. Apple Watch as the trust boundary addresses three threat-model legs that a traditional hardware wallet cannot:

- **No shipping address required for purchase.** App Store distribution means Apple knows the install; Cryptograph the company never receives a name or shipping address. The Ledger 2020 breach disclosed approximately 270,000 customers' physical addresses; that surface is excluded by not collecting it.
- **No identifiable physical form factor.** An Apple Watch in airport security looks like a watch. A Ledger or Trezor in luggage is obviously a hardware wallet. For users whose threat model includes the targeting phase rather than just the on-device phase, this is material.
- **Same cryptographic hardware-isolation property** as a dedicated device. The Apple Secure Enclave is the trust boundary; the iPhone composes transactions but holds no keys and cannot sign. watchOS (and Wear OS in the future) are mass-market embedded device platforms which lack many of the exploit vectors common to mobile phone-based hot wallets (e.g., no browser, no remote access, no USB port). Their wearable form factor also is more resistant to petty theft, although still vulnerable to coercion.

A different answer to the same hardware-isolation problem, optimized for users whose threat model includes the targeting phase.

## Verification

```bash
pnpm install
pnpm run check:astro     # 0 errors
pnpm run check:vitest --run
pnpm run check:syntax    # clean
pnpm run check:misc      # clean
pnpm run check:spelling:quiet  # clean
pnpm run dev             # Cryptograph renders correctly
```

`pnpm run check:lint` reports pre-existing lint warnings on upstream `beta` (`src/constants/variants.ts`, `src/types/content/news.ts`, `src/views/NavigationItems.svelte`); none of our files trip new ESLint issues. Verified by stashing changes and re-running.


polymutex:
Hey there, thanks for contributing and really happy to hear that some of Walletbeat's attributes encouraged some changes :)

This is going to be a bit of a difficult wallet to rate given Walletbeat's current schema. There's clearly a hardware-wallet-ish aspect to this, but at the same time my understanding here is that the user is required to use the Cryptograph app specifically in order to use the wallet overall, which means this looks more like a software wallet with a secure enclave component to it. Will have to think about this more... The wallet feature schema was meant to allow for simultaneously-software-and-hardware wallets for this reason, but we recently lifted this constraint and now looks like it's coming back to bite us, sorry :(

On the specific questions:

> `supplyChainFactory: PASS` with Apple as the manufacturer
> `hardwareWalletManufactureType: FACTORY_MADE`

Seems fine for now; note that in general we'll revamp the hardware wallet feature set at a later date and right now the schema is pretty bare-bones anyway as we focus on software wallets.

> `dataCollection` populated with apiproxy-level abstraction

This is the right thing to do.

perpetua-engineering:
Thanks for the thoughtful read. The core argument for hardware-wallet categorization is that signing happens on a separate physical device. The Apple Watch is a distinct computer from the iPhone, with its own SoC, memory, and Secure Enclave. The iPhone composes transactions but cannot sign; the channel between them is WCSession over BLE, which is network-shaped rather than same-process. Even a kernel-level rootkit on the iPhone cannot sign without user action on the watch face. That's the hardware-isolation property the schema's hardware-wallet attributes are designed to measure, and it holds regardless of whether the user has signing-software choice.

The app-bound signing observation is fair. We're currently the only signing software that talks to the watch-side signer. If a third-party-signing-app interface is something walletbeat would value seeing in hardware-wallet listings, it's something we can design and add. A WalletConnect-style protocol that lets other iPhone apps reach the watch-side signer is a real product feature we'd be happy to build, and your input on what shape would be most useful is welcome.

polymutex:
> Thanks for the thoughtful read. The core argument for hardware-wallet categorization is that signing happens on a separate physical device. The Apple Watch is a distinct computer from the iPhone, with its own SoC, memory, and Secure Enclave. The iPhone composes transactions but cannot sign; the channel between them is WCSession over BLE, which is network-shaped rather than same-process. Even a kernel-level rootkit on the iPhone cannot sign without user action on the watch face. That's the hardware-isolation property the schema's hardware-wallet attributes are designed to measure, and it holds regardless of whether the user has signing-software choice.

Understood.

I think ultimately this model makes sense under what we've been wanting to implement as "multi-wallet-type" wallets, the canonical example being Ledger wallets which are typically set up using with Ledger Live and which can be used with Ledger Live as the main software wallet interface, but which also support integration into other software wallets. We don't support representing such wallets yet unfortunately, so ultimately the choice currently has to be constrained by either representing the wallet as a software wallet, or as a hardware wallet.

> The app-bound signing observation is fair. We're currently the only signing software that talks to the watch-side signer. If a third-party-signing-app interface is something walletbeat would value seeing in hardware-wallet listings, it's something we can design and add. A WalletConnect-style protocol that lets other iPhone apps reach the watch-side signer is a real product feature we'd be happy to build, and your input on what shape would be most useful is welcome.

Definitely useful. You'll notice one of the attributes Walletbeat rates for is hardware wallets' permissionless integration ability by software wallets, without intermediaries (whether the hardware wallet manufacturer, or another network-intermediary-requiring protocol like WalletConnect). Once that is achieved, it starts making sense to separately rate the hardware component (here, the watch-side signer) and its security properties on its own, and the software-side component (the iOS app) also on its own, since they would be separate and modular components.

Until this is the case, I still think this system in its current form looks more like what Walletbeat calls a "software wallet", with the added security benefit of having a difficult-to-compromise hardware enclave where the key material resides. What Walletbeat calls "hardware wallets" are more of the signer-only component. Here, since the system is a package deal, "software wallet" is the closer abstraction, because that designation encompasses the parts that users interface with beyond the hardware signer, which they are required to use here. For example, this includes capabilities like scam prevention measures in the UI, Account Abstraction, L1 provider independence, ERC-20 permissions management, fee visibility in the TX approval flow, permissionless transaction inclusion, cross-chain bridging support, etc., all of which users would expect out of an integrated wallet system, and require software-side components. And those components are currently only rate-able as part of "software wallets".

So, course of action here, I suggest either:

A. Update this PR to follow the software wallet schema, though include heavy comments or a partially-filled `HardwareWallet`-typed object next to it in the same data file so that we can fold that data in once we have true multi-wallet-type support.

B. Put this on pause PR for the moment, work on modularizing the hardware signer bits away to enable integration by other software wallets. Then, once done:

- If Walletbeat has implemented multi-wallet-type support by then, then use that to represent this wallet.
- Otherwise (no multi-wallet-type support), then repurpose this PR to add separate data files for the hardware signer part and the software wallet iOS app as 2 separate wallet entries in Walletbeat, and it'll be on us to merge them back once we implement multi-wallet-type support.

perpetua-engineering:
Thanks for the framework. The signer-as-modular-component vs full-system-as-software-wallet distinction is clarifying, and we'd like to pursue option B's modularization path. Before scoping the engineering work, we want to confirm the design would satisfy your permissionless-integration criterion.

The iOS architecture constrains what's possible. watchOS apps can only communicate with their paired iPhone app via WCSession; a third-party iOS app cannot reach a different bundle's watchOS app directly. So any third-party software wallet (MetaMask, Rabby, Frame) reaching Cryptograph's watch-side signer has to flow through the Cryptograph iPhone app as a local relay:

```
Third-party iOS app → Cryptograph iPhone app (local relay) → Cryptograph Watch app (signer)
```

The cleanest iOS-native protocol for the third-party → Cryptograph leg is a documented URL scheme + callback URL pattern. Roughly:

```
cryptograph://sign?chain=1&payload=0x...&callback=metamask://signed
```

Properties of this design:

- Publicly documented spec; any wallet can implement client-side without our involvement
- No Perpetua-side gatekeeping or per-app approval; the spec is the API
- No central server, no WalletConnect-style cloud relay
- The user sees the calling app's identity in the approval surface on both iPhone and watch (anti-phishing affordance)
- The Cryptograph iPhone app is in the local flow but is not adding intermediary control beyond routing and approval display

The question: does this satisfy the permissionless-integration criterion, or does "without intermediaries" include the local-relay pattern that iOS forces? If yes, we'd treat the URL scheme as the spec target and update this PR after the modularization work ships. If no, we'd want to understand what shape would satisfy, given the iOS constraint that prevents direct third-party-app-to-watch communication.

PR #719, opened 2026-05-10
fix(schema): exempt hardware wallets from software-only attributes (#535)

### LINKED ISSUE #535: Exclude hardware wallets from attributes that don't support them
Looks like the "scam prevention" attribute doesn't correctly exempt hardware wallets. We should fix that. There are probably others like it.

premsreelathasugeendran:
## Summary

Two attributes that conceptually only apply to software wallets were missing the HARDWARE exemption check in their `evaluate` function, causing hardware wallets to be rated (and typically fail) on criteria that don't apply to them:

- **`scamPrevention`** — scam alerts (URL warnings, recipient warnings, contract warnings) target software wallets that interact with dapps and addresses directly. Hardware wallets sign transactions presented by a host application and don't have an independent scam-alert layer.
- **`passkeyImplementation`** — the docstring at line 235 already states *"Wallets that don't support smart contract accounts or are hardware wallets are exempt from this rating,"* but the `evaluate` function never enforced it.

## What changed

For each of the two attributes, added a single early-return check right after `setVerifiability(...)`:

```ts
if (ctx.features.type === WalletType.HARDWARE) {
    return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'))
}
```

This follows the exact pattern already used by `chain-verification`, `security-audits`, and other attributes that correctly exempt hardware wallets.

## Verification

- `git diff --cached --stat` → 2 files, 13 insertions, 2 deletions
- TypeScript checks: my changes type-check clean (existing tsconfig/test errors are pre-existing and unrelated to this diff)
- No new dependencies, no behavior change for software wallets
- Hardware wallets now correctly drop these two columns from the `/hww/summary` aggregate view

## Closes

#535

## Notes for the reviewer

The issue body says *"There are probably others like it."* — I checked the rest of the security attributes and couldn't find further software-only ones still missing the HW exemption (the obvious candidates — `chain-verification`, `security-audits`, `security-best-practices`, `supply-chain-diy` — already have it; the `*hardware*` files are HW-only). Happy to look further if you'd point me at suspected ones.

### DIFF
--- a/src/schema/attributes/security/passkey-implementation.ts
+++ b/src/schema/attributes/security/passkey-implementation.ts
@@ -5,9 +5,10 @@ import {
 } from '@/schema/features/security/passkey-verification'
 import { isSupported } from '@/schema/features/support'
 import { verifiabilityRequiresSourceCodeAccess } from '@/schema/verifiability'
+import { WalletType } from '@/schema/wallet-types'
 import { markdown, mdParagraph, mdSentence, paragraph, sentence } from '@/types/content'
 
-import { pickWorstRating, unrated } from '../common'
+import { exempt, pickWorstRating, unrated } from '../common'
 
 export type PasskeyImplementationMetadata = {
 	library: PasskeyVerificationLibrary | null
@@ -325,6 +326,11 @@ export const passkeyImplementation: Attribute<PasskeyImplementationMetadata> = {
 	aggregate: pickWorstRating,
 	evaluate: ctx => {
 		ctx.setVerifiability(verifiabilityRequiresSourceCodeAccess({ coreOnlyIsSufficient: true }))
+
+		if (ctx.features.type === WalletType.HARDWARE) {
+			return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'))
+		}
+
 		const passkeyVerification = ctx.features.security.passkeyVerification
 
 		if (passkeyVerification === null) {
--- a/src/schema/attributes/security/passkey-implementation.ts
+++ b/src/schema/attributes/security/passkey-implementation.ts
@@ -328,7 +328,13 @@ export const passkeyImplementation: Attribute<PasskeyImplementationMetadata> = {
 		ctx.setVerifiability(verifiabilityRequiresSourceCodeAccess({ coreOnlyIsSufficient: true }))
 
 		if (ctx.features.type === WalletType.HARDWARE) {
-			return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'))
+			return exempt(
+				ctx,
+				sentence(
+					'This attribute is not applicable to hardware wallets because hardware wallets rely on dedicated secure hardware.',
+				),
+				{ library: null },
+			)
 		}
 
 		const passkeyVerification = ctx.features.security.passkeyVerification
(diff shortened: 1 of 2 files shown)


src/schema/attributes/security/scam-prevention.ts
```diff
@@ -584,6 +585,10 @@ export const scamPrevention: Attribute<ScamPreventionMetadata> = {
 			}),
 		)
 
+		if (ctx.features.type === WalletType.HARDWARE) {
+			return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'))
+		}
+
```
maykelxyz, line 591: This will return an error. Please return something like:

```ts
if (ctx.features.type === WalletType.HARDWARE) {
	return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'), {
		scamAlerts: null,
	})
}
```

src/schema/attributes/security/passkey-implementation.ts
```diff
@@ -325,6 +326,11 @@ export const passkeyImplementation: Attribute<PasskeyImplementationMetadata> = {
 	aggregate: pickWorstRating,
 	evaluate: ctx => {
 		ctx.setVerifiability(verifiabilityRequiresSourceCodeAccess({ coreOnlyIsSufficient: true }))
+
+		if (ctx.features.type === WalletType.HARDWARE) {
+			return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'))
+		}
+
```
maykelxyz, line 339: Same here:

```ts
if (ctx.features.type === WalletType.HARDWARE) {
	return exempt(ctx, sentence('This attribute is not applicable for hardware wallets.'), {
		library: null,
	})
}
```

maykelxyz (changes requested):
Looks good to me overall. Just need to fix some things stated in the comments. 

Small suggestion: Add reason to why hardware wallets are exempted: 'This attribute is not applicable to hardware wallets because hardware wallets rely on dedicated secure hardware.'

Please make sure to run `pnpm check:all` to check if all CI checks are OK

premsreelathasugeendran:
Thanks for the careful review @maykelxyz — much appreciated.

Pushed `776ba56c` addressing all three points:

1. **scam-prevention** (line 591): now returns `exempt(ctx, ..., { scamAlerts: null })` per your suggestion ✅
2. **passkey-implementation** (line 333): now returns `exempt(ctx, ..., { library: null })` ✅
3. **Reworded the exemption message** to explain *why*: *"This attribute is not applicable to hardware wallets because hardware wallets rely on dedicated secure hardware."*

Drive-by cleanup from `pnpm check:lint --fix`: once the new metadata-typed `exempt()` call landed, TS could infer the outcome shape correctly, so I removed six now-unnecessary `as NonNullScamPreventionMetadataScaffold` assertions in `evaluateScamAlerts`, plus the now-unused local type alias and `EvaluationScaffold` import. Lint + prettier both clean locally.

Let me know if you'd like that cleanup split into a separate commit.

maykelxyz (approved):
Looks good. I just realized Passkey Implementation is not showing anywhere in the attribute groups even on software wallets. I forgot I removed it last time when I was doing the `Security Best Practices` attribute [here](https://github.com/walletbeat/walletbeat/pull/582/changes#diff-5472cb8d1e494acd2e512426782d1e6b089fea2801beb591ff4becc2914dbaf6).

Iirc, it was also discussed in another PR that the `Passkey Implementation` should still exist, and we should bring it back..

cc: @polymutex

polymutex:
> Looks good. I just realized Passkey Implementation is not showing anywhere in the attribute groups even on software wallets. I forgot I removed it last time when I was doing the `Security Best Practices` attribute [here](https://github.com/walletbeat/walletbeat/pull/582/changes#diff-5472cb8d1e494acd2e512426782d1e6b089fea2801beb591ff4becc2914dbaf6).
> 
> Iirc, it was also discussed in another PR that the `Passkey Implementation` should still exist, and we should bring it back..
> 
> cc: @polymutex

We should just remove it and fold it in Security Best Practices. But that's for another PR.

polymutex:
Thanks :)

PR #720, opened 2026-05-11
Data: Add rainbow wallet

maykelxyz:
## Updates
- Added bug bounty program: not supported
- Duress resistance supported
  - Basic unlock supported
  - Duress mode not supported
- Added Ethereum L1 light client: Not supported
- No published public security audits
- Permissions management not supported
- Transaction submission
  - L1
    - Self broadcast via direct gossip: not supported
    - Self broadcast via self hosted node: supported 
  - L2
    - OP: Supported but no force inclusion
    - Arbitrum: Supported but no force inclusion


polymutex (commented):
👍

PR #740, opened 2026-05-16
Data: add account recovery for Base App via recovery key

ren2140eth:
## Summary

Sets `accountRecovery.guardianRecovery` for Base App to `supported(...)` modelling Coinbase Smart Wallet's "recovery key" feature as a single SELF_CUSTODY guardian with 1-of-1 client-side reconstitution. Follows [Rainbow's loose use](data/software-wallets/rainbow.ts) of `SECRET_SPLIT_ACROSS_GUARDIANS` for similar single-secret backup patterns.

## How Coinbase's recovery key works

Per [Coinbase Help — Recover your smart wallet](https://help.coinbase.com/en/wallet/getting-started/smart-wallet-recovery) and the [`MultiOwnable.sol`](https://github.com/coinbase/smart-wallet/blob/0fe87f18488fa89b792896d79de3200242778a68/src/MultiOwnable.sol) contract:

- A recovery key is a standalone Ethereum private key the user generates
- It is registered on-chain as an additional owner via `addOwnerAddress`
- A holder of the recovery key has full owner powers — can drain the wallet, evict the original passkey, add new owners
- No encryption layer; the recovery key itself is the secret
- No threshold (1-of-N: any owner can act unilaterally)
- No timelock, no notification mechanism
- Multiple recovery keys may be registered, each acting independently

## Schema mapping

| Field | Value | Why |
|---|---|---|
| `type` | `SECRET_SPLIT_ACROSS_GUARDIANS` | Closest fit. CSW's pattern is neither pure secret-split nor classic K-of-N-with-timelock; the schema's `K_OF_N_WITH_TIMELOCK` requires `timelockWarningSentByAllOf: NonEmptyArray<Entity>` which CSW has no honest value for. |
| `optionalGuardians` | `[{ type: SELF_CUSTODY }]` | The recovery key is a self-custodied private key generated by the wallet and stored by the user. |
| `optionalGuardiansMinimumConfigurable` | `1` | User configures one recovery key. |
| `optionalGuardiansMinimumNeededForRecovery` | `1` | One key is sufficient to recover. |
| `requiredGuardians` | `[]` | Nothing else is mandatory — no wallet password, no Coinbase server, no mandated cloud account. The recovery key alone is sufficient. |
| `secretReconstitution` | `'CLIENT_SIDE'` | Recovery happens via on-chain transaction signed with the recovery key; no Coinbase-side reconstitution. |

The `descriptionMarkdown` documents the security model candidly (no encryption, full unilateral powers, no timelock/notification) so the evaluator's UI surfaces these properties.

## Caveat

This is a slight stretch of `SECRET_SPLIT_ACROSS_GUARDIANS` since CSW's recovery key isn't really *split*. The schema's docstring acknowledges this gap: *"if a wallet uses a recovery scheme that doesn't fit any of these types, a new enum value should be added."* A future schema extension (e.g. `PRE_AUTHORIZED_FULL_OWNER`) could model this more precisely. Tracked informally.

## Test plan

- [ ] CI lint, prettier, type-check, and build pass
- [ ] Account Recovery section renders on the Base App detail page with the descriptionMarkdown's security caveats visible

### DIFF
--- a/data/software-wallets/base-app.ts
+++ b/data/software-wallets/base-app.ts
@@ -2,7 +2,6 @@ import { ren2140 } from '@/data/contributors/ren2140'
 import { AccountType } from '@/schema/features/account-support'
 import type { AddressResolutionData } from '@/schema/features/privacy/address-resolution'
 import { WalletProfile } from '@/schema/features/profile'
-import { GuardianPolicyType, GuardianType } from '@/schema/features/security/account-recovery'
 import {
 	BugBountyPlatform,
 	BugBountyProgramAvailability,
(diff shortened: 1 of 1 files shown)


polymutex:
Sorry, but please see comment on https://github.com/walletbeat/walletbeat/pull/745#issuecomment-4472574966. Apparently this feature requires accessing `keys.coinbase.com`, aka relying on an out-of-wallet service in order to perform recovery. As such, the wallet itself doesn't implement account recovery.

ren2140eth:
Hmmm this is a bit nuanced though so for example if you’re using an external service for recovery (e.g. your guardian is linked to an email address you control so in order to recover your wallet you will need to click a link emailed to you that will open in your browser and activate the recovery flow). This is just as example based on my memory of how the old argent Ethereum wallet worked that had implemented social recovery features. I think this would constitute account recovery functionality even though the flow takes place outside the wallet UI? Similarly then the flow for base doesn’t seem very different than this. OR since you can only rely on coinbase to recover we will assume this is not sufficient to pass which would be different logic than the integrated in UI argument?

polymutex (changes requested):
Reading the [Coinbase help page on this](https://help.coinbase.com/en/wallet/getting-started/smart-wallet-recovery), it seems like the recovery phrase isn't stored on Coinbase's servers at all, so that makes it "self-custody" (as the PR correctly states). However:
- The *tool* to execute the recovery is a Coinbase-hosted webpage, and there doesn't appear to be an open-source self-hosted version of it, i.e. you are reliant on Coinbase to maintain this tool and that it works by the time you need it.
- There is no other recovery factor; the seed phrase unilaterally controls the smart wallet account. So the use of "SECRET_SPLIT_ACROSS_GUARDIANS" doesn't make much sense, since there is no split happening.

Given this, I think the best thing to do here is to keep this as `null` for now, as we need a new type of `GuardianPolicy` to represent this situation better than the current two options (`GuardianPolicyType.SECRET_SPLIT_ACROSS_GUARDIANS` and `K_OF_N_WITH_TIMELOCK`) can.

As for the Argent case, I think there's been many flavors of Argent wallets so I'm not sure which one you refer to. Nonetheless, the distinction I think is: *must* you rely on Argent itself, or is Argent just one of multiple configurable guardian options?

- If Argent is *only* one of many possible guardians, and the way you prove to Argent-the-guardian that you "own" your account is by email verification, then this is OK. That would be an "optional guardian" per the `SECRET_SPLIT_ACROSS_GUARDIANS` scheme.
- If Argent is the *sole* guardian option, then obviously it is not an "optional guardian", it is a required one, and you now have a single point of failure on Argent.
- If Argent is only of many possible guardians, but that the account recovery process still requires some type of interaction with Argent's live servers even if the user configured their wallets to be recoverable using non-Argent guardians, then Argent is also a "required guardian", and you have a single point of failure on Argent.

Ways in which Argent can provide account recovery without becoming a "required guardian" involves self-hosted or self-run tools. For example, they have an [account recovery CLI](https://github.com/argentlabs/argent-starknet-recover) which users can run on their own computer, without Argent in the loop.

Another way seems to be the [encrypted recovery kit](https://support.argent.xyz/hc/en-us/articles/4410361683601-Saving-Encrypted-Recovery-Kit-for-offchain-recovery), which according to the help article is detected during startup of the Argent wallet. That implies the user doesn't need any Argent server interaction, they just need to download the Argent app from the app store and the wallet will implement the recovery flow as part of its onboarding process, no Argent servers involved.

Hope this makes sense :)

ren2140eth:
Thanks this is helpful. Yes in the Argent example it's one of multiple guardians and you didn't need to use argent as a guardian at all. But for Coinbase it does appear that you can only recover your wallet by interacting with Coinbase so this would be a single point of failure. So can revert to null for now but would be valuable to surface this as an option that does exist even though it will fail once we can update the schema

ren2140eth:
Closing per the discussion above — keeping `accountRecovery: null` for now (which beta already reflects).

Conclusion: Base App's smart-wallet recovery is self-custodial in that the recovery phrase isn't stored on Coinbase's servers, **but** the only tool to execute recovery is a Coinbase-hosted webpage with no self-hostable alternative, and the seed phrase unilaterally controls the account (no secret split). Neither current `GuardianPolicyType` (`SECRET_SPLIT_ACROSS_GUARDIANS`, `K_OF_N_WITH_TIMELOCK`) can represent a 'sole provider-hosted recovery tool = single point of failure on the provider' situation.

So this stays `null` until the schema gains a type that models it.

PR #745, opened 2026-05-17
Data: add permissionsManagement for Base App via Spend Permissions

ren2140eth:
## Summary

Sets `selfSovereignty.permissionsManagement` for Base App to `supported(...)` modelling Coinbase Smart Wallet's "Spend Permissions" mechanism. Was `null` (not yet checked).

## How Coinbase's Spend Permissions work

A Coinbase-specific mechanism (the `SpendPermissionManager` contract added as an owner of the user's smart wallet) where dApps request **recurring spending authority** on the user's native ETH or ERC-20 tokens with **amount + frequency + start/end-date** semantics. Distinct from traditional `token.approve()` allowances.

Sources:
- [github.com/coinbase/spend-permissions](https://github.com/coinbase/spend-permissions) — README confirms ETH + ERC-20 scope, NFT exclusion
- [Base docs — Build with Spend Permissions](https://docs.base.org/identity/smart-wallet/guides/spend-permissions/)
- [Coinbase Help — Base account spend permissions](https://help.coinbase.com/en/wallet/getting-started/smart-wallet-permissions)

## Schema mapping

| Field | Value | Reason |
|---|---|---|
| `erc20Approvals` | `CAN_INSPECT_AND_REVOKE` | Spend Permissions cover ERC-20 spend authority and the Base Account web dashboard exposes both inspection and revocation. Scope caveat: traditional `token.approve()` allowances granted via legacy paths are NOT shown (users must use external tools like revoke.cash). |
| `erc721Approvals` | `CANNOT_INSPECT` | NFTs are architecturally excluded from Spend Permissions per the Coinbase repo README ("tighter and fully-known scope of account control"). |
| `erc1155Approvals` | `CANNOT_INSPECT` | Same as above. |

## Caveat: web-only revocation

The Base mobile app v29.94.123 does NOT have an in-app UI for inspecting or revoking Spend Permissions. Users are linked out to the Base Account web dashboard at [keys.coinbase.com](https://keys.coinbase.com) (Settings → Settings for your Base Account → Manage permissions). This is consistent with how we modelled `accountRecovery` in PR [#740](https://github.com/walletbeat/walletbeat/pull/740) — walletbeat treats the wallet as the holistic Base Account product, not just the mobile binary.

## Test plan

- [ ] CI lint, prettier, type-check, and build pass
- [ ] Base App detail page renders the permissions management section reflecting CAN_INSPECT_AND_REVOKE for ERC-20 and CANNOT_INSPECT for NFTs

### DIFF
--- a/data/software-wallets/base-app.ts
+++ b/data/software-wallets/base-app.ts
@@ -7,7 +7,6 @@ import {
 	BugBountyProgramAvailability,
 } from '@/schema/features/security/bug-bounty-program'
 import type { ContractTransactionWarning } from '@/schema/features/security/scam-alerts'
-import { SpendingApprovalsControl } from '@/schema/features/self-sovereignty/permissions-management'
 import { TransactionSubmissionL2Type } from '@/schema/features/self-sovereignty/transaction-submission'
 import { featureSupported, notSupported, supported } from '@/schema/features/support'
 import { comprehensiveFeesShownByDefault } from '@/schema/features/transparency/fee-display'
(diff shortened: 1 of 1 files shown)


polymutex:
> This is consistent with how we modelled accountRecovery in PR https://github.com/walletbeat/walletbeat/pull/740 — walletbeat treats the wallet as the holistic Base Account product, not just the mobile binary.

Sorry, but I strongly disagree with this approach. Relying on an out-of-wallet-software website for a feature like spend permission management is as meaningless as claiming that a wallet integrates built-in swaps just because it has a button that opens app.uniswap.org in a new tab when clicked.

Please adjust both PRs and the feature comments/documentation accordingly to ensure Claude doesn't end up logic'ing its way down this path in future PRs.

ren2140eth:
Noted but this is a feature specifically built by Coinbase to serve this functionality so it’s a bit different than the uniswap analogy since the equivalent would be using revoke.cash not going to keys.coinbase logging in with the passkey that controls your wallet then revoking permissions that are natively built into the wallet. So then I’d have a question I see on rabby we also have set to fail although you definitely can revoke erc 20 and nfts (both 721 & 1155) they just enabled the feature to let you do this within the browser wallet itself but before it directed you to a website built by rabby where you could go thru and revoke approvals (in this case would we say rabby doesn’t provide this functionality just because it redirects you to a website? If it redirected you to revoke cash then I think it would NOT provide the functionality but that’s not the case).

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

polymutex:
Yes, if Rabby just goes to `revoke.cash` then it also doesn't support it either. (It's nice that they link to it vs the alternative of not doing anything at all to help the user, but still not qualifying.)

The same entity being behind both the wallet and the service is a relevant distinction here, sure, but I don't think that's enough of a justification either.

If we accepted that justification, then we would say that "Google Chrome has integrated search functionality" just because typing a non-URL in the address bar goes to the Google Search website, which is operated by the same entity. Yet I don't think someone would claim "Google Chrome has integrated search functionality". It just has a UI shortcut to a service that Google also runs.

Similarly, consider Uniswap Wallet. Would you say that it has "built-in swaps" if its swap button just opened a new tab to app.uniswap.org? Probably not, yet both services are operated by the same foundation.  So the bar for having "integrated [something]" seems higher than just "has a button to go on a website for [something] operated by the wallet entity".

ren2140eth:
Well rabby currently has it integrated within the browser wallet pop up UI (so we should update this). This is a recent update. Previously it opened a tab in your browser to their own revoke UI NOT revoke.cash based on our strict definition of integrated we’d rate this as not passing? The fact that rabby provides this is genuinely useful and some solution is better than none even if it’s not fully integrated but I think it’s fair to have this strict standard for integrated features. So then we’re essentially saying if the feature is not available in the browser wallet pop up UI or the mobile app UI then we wouldn’t consider it. I think there’s even more nuance with base because they have their own custom implementation but wouldn’t change the fact that you would need to go to an external site to manage the permissions. I’ll revert this back to null in this case

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

ren2140eth:
Reverted in [3605c4b8](https://github.com/walletbeat/walletbeat/pull/745/commits/3605c4b8). `permissionsManagement` is back to `null`, with a code comment documenting (a) what the external Spend Permissions mechanism is, (b) where it lives (keys.coinbase.com), and (c) why we're not counting it as supported per the in-wallet-UI-only standard discussed in this thread. The principle is now noted for future PRs so the same misclassification doesn't recur.

polymutex:
I recognize there can some nuance here (a wallet with a button to do [good thing] is better than a wallet with no button at all), but perhaps if we're going to add nuance it should be based on how critical and fungible the feature is.

For token spend permissions, revoke.cash works with any wallet, because these permissions are part of the ERC-20 token standard and so all wallets are essentially compatible with each other.

Contrast that to a feature like account key rotation, where there is no such standard and where if you use Coinbase's flavor of smart accounts, you need Coinbase's specific web-based tool to rotate keys. That is a single point of failure and you are essentially at the mercy of Coinbase to keep this tool.

For simplicity, I'd say let's keep this as "feature either exists directly implemented in the wallet software, not in any external site" just to be consistent across all such features.

polymutex (approved):

ren2140eth:
Ok makes sense

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

PR #761, opened 2026-05-20
Refactor: Security best practices

### LINKED ISSUE #746: Security best practices: Allow evaluation to proceed even when manifests are not available
The Security Best Practices attribute currently relies on extension manifests to derive things like extension permissions. Some wallets are closed-source, so we can't get their manifests. However, it is bad from a data entry perspective that we simply cannot rate a wallet on security just because it is closed-source, because it means some wallets would forever have gray on them, which looks like it's Walletbeat being lazy/incomplete, rather than the wallet's closed-ness that is the problematic part.

We should have a sentinel value for "manifest not available / not published" (similar to the `NOT_AN_ANDROID_APP`-like sentinel values), and return the whole evaluation as `FAIL` and unverifiable in such a case.

maykelxyz:
# Refactor: Security best practices

Closed-source wallets had no way to populate `securityBestPractices`, leaving them permanently gray (`UNRATED`). This looks like Walletbeat "not being able to gather data" rather than the wallet being closed-source.

Fixed by adding two mechanisms:

- **`SOURCE_NOT_AVAILABLE`** on variant fields (`browser`, `mobile`, `desktop`): returns as `FAIL` when **no data** is obtainable at all
- **`KeyStorageMechanism.NOT_VERIFIABLE` / `SecureRngSource.NOT_VERIFIABLE`**: allows browser wallets fill in real manifest data (which is always public in extension stores) while marking the enum fields as `FAIL`
  - Allows users / researches to fill in individual data (manifest data) regardless if the source code is not publicly available.

A sample wallet data would look like:

```ts
securityBestPractices: {
  browser: {
	  ref: refTodo,
	  browserExtensionHardening: parseBrowserExtensionManifest(metamaskRawExtManifest),
	  keyStorageMechanism: KeyStorageMechanism.NOT_VERIFIABLE,
	  secureRng: SecureRngSource.NOT_VERIFIABLE,
  },
  desktop: 'NOT_A_DESKTOP_APP',
  mobile: 'SOURCE_NOT_AVAILABLE',
},
```

Fixes #746

### DIFF
--- a/src/schema/attributes/security/security-best-practices.ts
+++ b/src/schema/attributes/security/security-best-practices.ts
@@ -985,7 +985,7 @@ export const securityBestPractices: Attribute<SecurityBestPracticesValue> = {
 			),
 			exampleRating(
 				mdParagraph(
-					"The wallet is closed-source. For browser wallets with a public extension manifest, set `keyStorageMechanism` and `secureRng` to `NOT_VERIFIABLE` while filling in actual manifest data. For mobile/desktop or when no data is obtainable at all, set the variant to `'SOURCE_NOT_AVAILABLE'`.",
+					'The wallet is closed-source. Security properties such as key storage, random number generation, and manifests cannot be independently verified.',
 				),
 				sourceNotAvailable(EvaluationContext.forTest(() => securityBestPractices)),
 			),
--- a/src/schema/attributes/security/security-best-practices.ts
+++ b/src/schema/attributes/security/security-best-practices.ts
@@ -44,6 +44,27 @@ import { aggregateVariantEvaluations, exempt, pickWorstRating, unrated } from '.
 
 export type SecurityBestPracticesValue = null
 
+function sourceNotAvailable(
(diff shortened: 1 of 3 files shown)


src/schema/attributes/security/security-best-practices.ts
```diff
@@ -922,6 +983,12 @@ export const securityBestPractices: Attribute<SecurityBestPracticesValue> = {
 					},
 				),
 			),
+			exampleRating(
+				mdParagraph(
+					"The wallet is closed-source. For browser wallets with a public extension manifest, set `keyStorageMechanism` and `secureRng` to `NOT_VERIFIABLE` while filling in actual manifest data. For mobile/desktop or when no data is obtainable at all, set the variant to `'SOURCE_NOT_AVAILABLE'`.",
```
polymutex: This is definitely too much technical detail and codebase jargon to show on the website.

polymutex (changes requested):

polymutex (approved):

PR #772, opened 2026-05-26
Feat: Add ERC-7730 support

### LINKED ISSUE #725: Add ERC-7730 support
References:
- https://clearsigning.org
- https://ethereum-magicians.org/t/eip-7730-proposal-for-a-clear-signing-standard-format-for-wallets/20403/2

maykelxyz:
Closes #725

# ERC-7730 Support

### Schema / Features
- Replaced `calldataDecoded: CalldataDecodingTypes` with `erc7730: Support<HardwareWalletErc7730>` on `HardwareTransactionLegibilitySupport`
- Added `erc7730: Support<SoftwareWalletErc7730>` and `transactionSimulations: Support<SoftwareTransactionSimulations>` to `SoftwareTransactionLegibilitySupport`
- Simplified `SoftwareTransactionDetailsDisplay` from a per-benchmark-transaction record to an individual struct (`gas`, `nonce`, `from`, `to`, `chain`, `value`)
- Added new types: `Erc7730TransactionDetails`, `SoftwareWalletErc7730`, `HardwareWalletErc7730`, `TransactionOutcomeEntry`, `SimulatedFailedTransaction`, `SimulatedNondeterministicTransaction`, `SoftwareTransactionSimulations`
- Removed: `HardwareBenchmarkTransactions`, `CalldataDecodingTypes`, `noCalldataDecoding`, `supportsAnyCalldataDecoding`, `isSupportedOnDevice`
- Moved `isShown` helper from attributes to features (now exported)

### Benchmark Transactions
- Added `AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND` to `ComplexBenchmarkTransactions` (EOA-side nested multisend, non-Safe variant)
- Fixed label for `SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND`: now `'complex Safe nested multisend transactions'`

### EIPs
- Added `data/eips/erc-7730.ts`.

### Wallet Data
- Migrated all wallets from `calldataDecoded` to `erc7730`

## Note
- Attribute evaluation changes to follow


maykelxyz:
For the transaction legibility attribute changes, here's currently what I'm thinking:
- To be able to "PASS" rating transaction legibility attribute, a wallet should FULLY SUPPORT both ERC-8213 and ERC-7730
  - ERC-8213 -> Required calldata hex, AND calldata digest. Requied to show eip-712 digest.
  - ERC-7730 -> Required to also decode inner calldata of batched transactions (ex. USDC APPROVE + AAVE SUPPLY) and safe transactions (both single transactions and batched transactions (USDC APPROVE + AAVE SUPPLY BATCHED))
- A wallet can get a "PARTIAL" rating if a wallet "partially" supports erc-8213 or erc-7730
  - ERC-8213 -> Shows  EIP-712 digest but no calldata digest.
  - ERC-7730 -> Has human readable calldata decoded format but doesn't decode inner calldata of batched transactions & safe transactions
- A wallet will automatically fail if a wallet DOES NOT support either.
  - A wallet fully supports ERC-8213 but does not have any ERC-7730 support. And vice versa.

polymutex (commented):
Types looking good so far.

src/schema/features/security/transaction-legibility.ts
```diff
@@ -406,35 +449,56 @@ export interface DisplayedNondeterministicTransactionDetails extends Omit<
...
+export function isTransactionOutcomeExplained(entry: TransactionOutcomeEntry | null): boolean {
+	return (
+		entry !== null &&
+		entry !== undefined &&
+		entry.transactionOutcome === TransactionOutcome.EXPLAINED
+	)
+}
+
+/** Returns true if the entry exists and its calldata was decoded and shown. */
+export function isTransactionDecoded(entry: { decoded: DataDisplayOptions } | null): boolean {
```
polymutex: Should return `boolean | null` in order not to lose the fact that the data wasn't present.
Either that or assert that the entry isn't `null`.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -605,206 +579,166 @@ function analyzeSoftwareFeatures({
...
+				erc7730Data !== null ? erc7730Data[ComplexBenchmarkTransactions.AAVE_SUPPLY] : null
+			const aaveSimEntry = simData[ComplexBenchmarkTransactions.AAVE_SUPPLY]
 
-			if (missing.length > 0) {
-				details.transactions.failing.push(`Aave supply (missing: ${commaListFormat(missing)})`)
-			} else if (!isShown(tx.calldataDecoded)) {
+			if (!isTransactionDecoded(aaveErc7730Entry)) {
 				details.transactions.failing.push('Aave supply (calldata not decoded)')
-			} else if (tx.transactionOutcome !== TransactionOutcome.EXPLAINED) {
+			} else if (!isTransactionOutcomeExplained(aaveSimEntry)) {
```
polymutex: Careful when updating this code too if you change the return signature to `boolean | null`, as this condition would still treat `false` and `null` equivalently, though they should not.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -605,206 +579,166 @@ function analyzeSoftwareFeatures({
...
 
-			if (missing.length > 0) {
-				details.transactions.failing.push(
-					`Safe nested Aave supply (missing: ${commaListFormat(missing)})`,
-				)
-			} else if (!isShown(tx.calldataDecoded)) {
+			if (!isTransactionDecoded(safeNestedErc7730Entry)) {
 				details.transactions.partial.push('Safe nested Aave supply (calldata not decoded)')
-			} else if (tx.transactionOutcome !== TransactionOutcome.EXPLAINED) {
+			} else if (!isTransactionOutcomeExplained(safeNestedSimEntry)) {
```
polymutex: Same

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -605,206 +579,166 @@ function analyzeSoftwareFeatures({
...
 
-			if (missing.length > 0) {
-				details.transactions.failing.push(
-					`Safe nested multisend (missing: ${commaListFormat(missing)})`,
-				)
-			} else if (!isShown(tx.calldataDecoded)) {
+			if (!isTransactionDecoded(safeMultisendErc7730Entry)) {
 				details.transactions.partial.push('Safe nested multisend (calldata not decoded)')
-			} else if (tx.transactionOutcome !== TransactionOutcome.EXPLAINED) {
+			} else if (!isTransactionOutcomeExplained(safeMultisendSimEntry)) {
```
polymutex: Same

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -605,206 +579,166 @@ function analyzeSoftwareFeatures({
...
+					? erc7730Data[
+							ComplexBenchmarkTransactions.AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND
+						]
+					: null
+			const batchApproveSupplySimEntry =
+				simData[ComplexBenchmarkTransactions.AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND]
+
+			if (!isTransactionDecoded(batchApproveSupplyEntry)) {
+				details.transactions.partial.push('EOA nested multisend (calldata not decoded)')
+			} else if (!isTransactionOutcomeExplained(batchApproveSupplySimEntry)) {
```
polymutex: Same, etc.

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1356,24 +1186,37 @@ export const transactionLegibility: Attribute = {
 		address, amount, fees, etc.) are clearly displayed on the wallet screen alongside decoded calldata,
 		allowing users to make informed decisions before authorizing transactions.
 
-		**ERC-8213: Standardizing What Users Sign**
-		[ERC-8213](https://erc8213.eth.limo) introduces two cryptographic digests that give users a
+		**${eipMarkdownLink(erc7730)}: Standardizing Calldata Decoding**
+		${eipMarkdownShortLink(erc7730)} defines a standard JSON descriptor format for structured calldata,
+		enabling wallets to display human-readable descriptions of smart contract calls.
+		Instead of showing raw hex calldata, wallets that implement ${eipMarkdownShortLink(erc7730)} can show
+		the function name and decoded parameters in plain language — for example,
```
polymutex: Emdash

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1356,24 +1186,37 @@ export const transactionLegibility: Attribute = {
...
+		"Approve 100 USDC for Aave" instead of the raw \`0x095ea7b3...\` bytes.
+
+		For hardware wallets, on-device ${eipMarkdownShortLink(erc7730)} decoding is especially valuable:
+		even if the companion software is compromised, the device itself can still display the correct
+		decoded description of the transaction being signed. Wallets are evaluated against a set of
+		benchmark transactions — from simple token approvals to complex Safe nested multisend batches — to
+		measure how broad their ${eipMarkdownShortLink(erc7730)} coverage is.
```
polymutex, line 1235: Same

polymutex (changes requested):

src/schema/features/security/transaction-legibility.ts
maykelxyz: Another solution I can think of is to have null guards in the function `evaluateSoftwareWalletTransactionLegibility` to return an `UNRATED` transaction legibility, and let `isTransactionDecoded` & `isTransactionOutcomeExplained` have a nullable entry (guarded in the `evaluateSoftwareWalletTransactionLegibility` function anyway) and still return a boolean.

```ts
if (isSupported(transactionSimulations)) {
  if (
	  transactionSimulations[BasicBenchmarkTransactions.ERC_20_TRANSFER] === null ||
	  transactionSimulations[BasicBenchmarkTransactions.ERC_721_TRANSFER] === null ||
	  transactionSimulations[BasicBenchmarkTransactions.ERC_1155_TRANSFER] === null ||
	  transactionSimulations[ComplexBenchmarkTransactions.USDC_APPROVAL] === null ||
	  transactionSimulations[ComplexBenchmarkTransactions.AAVE_SUPPLY] === null ||
	  transactionSimulations[ComplexBenchmarkTransactions.SAFEWALLET_AAVE_SUPPLY_NESTED] === null ||
	  transactionSimulations[
		  ComplexBenchmarkTransactions.SAFEWALLET_AAVE_USDC_APPROVE_SUPPLY_BATCH_NESTED_MULTISEND
	  ] === null
  ) {
	  return unrated(ctx)
  }
}
```

src/schema/features/security/transaction-legibility.ts
```diff
@@ -470,18 +470,13 @@ export type SoftwareTransactionSimulations = Record<
 
 /** Returns true if the entry exists and its outcome is fully explained. */
 export function isTransactionOutcomeExplained(entry: TransactionOutcomeEntry | null): boolean {
```
polymutex: Same comment here

src/schema/features/security/transaction-legibility.ts
polymutex: Sure, that doesn't solve the problem of this function's signature though. The point is that no function should be lossy:
- Either a function accepts a nullable input and returns a nullable output; the typesystem ensures the caller must handle the `null` case by virtue of the function potentially returning `null`.
- Either a function accepts a non-nullable input and returns a non-nullable output; the typesystem ensures the caller must handle the `null` case by virtue of forcing it to structure the code in a way that guarantees non-null-ness.
- (Less optimal but still acceptable) Either a function accepts a nullable input and returns a non-nullable output, but verifies that its input was non-null at runtime and `throw`s an error otherwise; we don't get typesystem-time checks, but at least we check at runtime, which is better than never or silent failure.

Even if in practice this function would only see non-null inputs, the fact that it accepts them in the input creates the possibility that it'll be misused in the future or after a refactor where we forget this (especially likely for an `export`ed function in a different package from where it's used), which is a footgun we can just avoid creating.

polymutex (changes requested):

polymutex (approved):

polymutex:
:ship:

PR #774, opened 2026-05-29
Ambire/ update privacy ratings

JIOjosBG:
Updates ambire's address and multiaddress privacy ratings

Also adds monad and citrea entities

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -68,6 +68,7 @@
 		"checksummed",
 		"chromeos",
 		"chromewebstore",
+		"citrea",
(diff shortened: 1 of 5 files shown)


polymutex (approved):
Looks good, thanks :) but note that this will have to be replaced with an actual packet capture file (see `pnpm wallet-data-collection` for the tool to help with collecting this) in the near future

JIOjosBG:
Great, as discussed outside of gh, i will wait for the tool to be finished before i do this

PR #778, opened 2026-05-30
Talks: Create outline for dappcon

maykelxyz:


resources/talks/2026-06-eth-berlin/dappcon/outline.md
```diff
@@ -0,0 +1,101 @@
...
+- Brief introduction to Walletbeat’s evaluation attributes:
+  - Security
+  - Privacy
+  - Transparency
+  - Self-sovereignty
+  - Ecosystem alignment
+
+---
+
+### The Transaction Legibility Problem
```
polymutex, line 31: Kind of a big jump to go from generic Walletbeat description to specifically the Transaction Legibility Problem. Need some sort of transition or reason to focus on this problem in particular.

One way to do this may be to lead with a "Why do wallets matter?" or "What do wallets really get wrong?" and then explain the Bybit hack, and only *then* afterwards talk about the blind signing problem and the ERC solutions.

It also gives the audience a reason to care upfront, then generalizes the problem, describes the solution. You could even move the general Walletbeat section to the end too as a generalization of this fix pattern. Then jump to demo

resources/talks/2026-06-eth-berlin/dappcon/outline.md
```diff
@@ -0,0 +1,101 @@
...
+#### Key Message
+
+Wallet UX and wallet security are increasingly becoming standards problems.
+
+Better standards improve:
+
+- interoperability
+- transaction clarity
+- user trust
+- wallet safety
```
polymutex, line 74: Add:
- Open competition

resources/talks/2026-06-eth-berlin/dappcon/outline.md
```diff
@@ -0,0 +1,101 @@
...
+- Comparing readable vs unreadable signing experiences
+- A real transaction flow using:
+
+  - Uniswap
+  - CoW Swap
+
+
+### Final Takeaway
+
+- Standards like ERC-7730 and ERC-8213 aim to improve user safety and interoperability
```
polymutex, line 111: And state Walletbeat's role in making those happen

polymutex (changes requested):

polymutex (approved):

PR #799, opened 2026-06-09
Data: Add Rainbow privacy.analytics.crashReports (Sentry, ALWAYS)

ren2140eth:
## Summary

Sets `privacy.analytics.crashReports` for Rainbow to `supported` (entity: Sentry, policy: `ALWAYS`).

Both Rainbow clients integrate **Sentry** for crash/error reporting, and it cannot be disabled by the user:

- **Mobile** (`src/logger/sentry.ts`): `initSentry` runs in all production builds (disabled only for dev/test). Its `beforeSend` only drops noisy fetch errors and does not honor any opt-out.
- **Browser extension** (`src/core/sentry/index.ts`): `initializeSentry` runs in all non-dev builds and is not gated by the analytics opt-out.
- The in-app **Analytics** toggle (Settings → Privacy) only disables RudderStack *usage* analytics via `analytics.disable()`. Its own description states that when disabled, *"only essential crash diagnostics are collected"* — confirming crash reporting stays on.

### Evidence

Refs include the relevant source files (pinned to commit hashes) **and** an in-app screenshot of the Privacy settings, committed as a file-based reference under `public/references/wallets/rainbow/screenshots/` (using the convention from #694 / #792).

## Validation

- `pnpm check:astro` (tsc) — 0 errors
- `eslint` / `cspell` / `prettier` — clean


polymutex (commented):
Thanks for taking the time to take the screenshot :)

PR #806, opened 2026-06-10
feat : added wallet eips page

### LINKED ISSUE #759: Wallet EIPs page
I think there's some benefit to having a dedicated "Wallet EIPs" page. A page that consists of EIPs that are wallet-related (712, 7730, 8312, etc.). It'll be very useful for when users are insterested in checking out different Wallet EIPs, why they matter for wallets, etc. in one page.

We can label this as low priority or post-launch :)

pheobeayo:
## Wallet EIPs page

Closes #759 

### What this adds

A dedicated `/wallet-eips` page listing all EIPs tracked by Walletbeat, with accordion-style detail panels and EIP/ERC filter pills.

### Files changed

- `src/components/WalletEipList.svelte` — list component using the existing `TooltipOrAccordion` pattern; renders filter pills (All / EIPs / ERCs) and one accordion row per EIP showing the short label, friendly name, and status badge. Expanding a row reveals the summary, wallet relevance explanation, optional notes, and a link to the canonical spec.
- `src/pages/wallet-eips/index.astro` — page at `/wallet-eips` using the standard `Layout` + `Navigation` pattern; passes `Object.values(eips)` directly into `WalletEipList`.

### Design decisions

- Imports directly from `data/eips.ts` — no duplicate data file
- Uses `TooltipOrAccordion` to match the existing wallet page accordion design
- Category filtering via pills rather than section grouping, per review feedback
- All EIP content (`summaryMarkdown`, `whyItMattersMarkdown`, `noteMarkdown`) comes from the existing `Eip` type — adding a new EIP to the page is automatic once it's added to `data/eips/`


data/wallet-eips.ts
maykelxyz, line 1: This feels like a duplicate of all EIP files in `data/eips`. I think most (if not all) EIPs in that directory are wallet-related. So I think there's no need for this file.

src/pages/wallet-eips/index.astro
maykelxyz, line 1: I like the idea of "grouping" EIPs based on their category.. But I think instead of doing multiple tabs based on EIP category, "Filtering" would be a bit better. 

And instead of doing "cards" as the UI, we can follow the current wallet page design, a dropdown component, that way, design is consistent across different pages. 

Sample:

Image

maykelxyz (changes requested):
Hi @pheobeayo, thanks for working on this :)

pheobeayo:
@maykelxyz @polymutex , Can you please review this?

src/components/WalletEipList.svelte
```diff
@@ -0,0 +1,276 @@
...
+		activeFilter === 'ALL' ? eips : eips.filter((e) => e.prefix === activeFilter),
+	)
+
+	// Status badge helpers
+	const statusLabel: Record<EipStatus, string> = {
+		[EipStatus.FINAL]: 'Final',
+		[EipStatus.DRAFT]: 'Draft',
+		[EipStatus.LIVING]: 'Living',
+		[EipStatus.LAST_CALL]: 'Last Call',
+	}
```
maykelxyz: Could be moved to `src/schema/eips.ts`

maykelxyz (changes requested):
Hi there, apologies for the late review, bit caught up with conference stuff. Currently reviewing this.

- The hover effect is useful but feels buggy ([See video](https://github.com/user-attachments/assets/80c5934f-d211-444c-8aa8-63ed350c4a8c)). For some reason it shows the popup component on the lower right part of the screen. Some other suggestions for the hover effect:
  - Only show summary. I think it's okay to not include "Why it matters for wallets"
- I suggest to also remove the "bullet" design per wallet EIP. It feels redundant design.
-  Maybe try to copy `/stages` design. https://beta.walletbeat.eth.limo/stages/ to highlight the card design. Instead of having a "no background"

Image


Directionally looks good :) one additional thing we can do is to add a "per-EIP page" that would track which wallets implement the certain EIP.. `/wallet-eips/7730` then would show which wallets implements erc-7730. We already have an ERC-7702 tracker in `/wallet/7702`, we can use that as a guide. Though, this isn't in the original scope (#759), so maybe in a separate PR.

maykelxyz:
Changes look good to me

image

Probably the last one:
-> A navigation item to redirect to the Wallet EIPs page :)

Also tagging @polymutex for a final review

polymutex (changes requested):
Looks good, but need a way to get to that page from the navbar or something, otherwise it's just a floating page without a way for users to get to it. Can you add it to `src/constants/navigation.ts`? You may need to find an icon for it.

pheobeayo:
@polymutex @maykelxyz can you help give a final review on this?

maykelxyz (approved):
Looks good to me. Im ok with having `transaction_legibility` as the icon I guess we have a couple of EIPs under the attribute

polymutex (approved):
Thank you again for your work :)

PR #807, opened 2026-06-10
Implement orderflow transparency attribute

johngrantuk:
See [previous WIP PR](https://github.com/walletbeat/walletbeat/pull/268) for history. I just started a fresh branch as it made it a bit easier to get my head around the whole thing.

Working from the spec given in [this comment](https://github.com/walletbeat/walletbeat/pull/268#issuecomment-3494095323)

Note - normally try to keep Ambire wallet up to date but research for this attribute will be quite heavy and might be best done as a follow up.

### Attribute logic

To achieve **PASS**, a wallet may do **any one** of:

- **A. Fully local** — Never send pre-inclusion transaction data to external parties by default; document via `dataCollection`
- **B. Non-extractive services** — May send pre-inclusion data by default but does not auction by default, and all recipients are verifiably non-extractive. Otherwise **PARTIAL**.
- **C. Transparency package** — Auctions by default **and** meets disclosure + practices page requirements. Onchain verification at `METHOD_DOCUMENTED_AND_EFFECTIVE` is required for **PASS** on this path; weaker documentation yields **PARTIAL**.

1. **EXEMPT** — `WalletType.HARDWARE` or `WalletType.EMBEDDED`.

2. **UNRATED** — Insufficient feature data. Stop here when any of the following apply (strict, like `feeTransparency` / `release-process`):
   - `privacy.dataCollection` is `null`, or any of the four transaction-related flows is still `null` (not `FLOW_NOT_SUPPORTED`).
   - `hasMempoolCollectionWithoutEndpoint(dataCollection)` is `true`.
   - Wallet auctions orderflow by default (`auctionsOrderflowByDefault(dataCollection)` is non-empty), but `transparency.orderflowPractices` is `null` after `nullable()` resolution — including when any required nested field (e.g. `pageLastUpdated`) is still `null`.
   - Wallet auctions by default and step 7 needs a prominence comparison, but `transparency.operationFees` is `null` or yields no worst-case `FeeDisplay` — **UNRATED**, not FAIL.

3. **PASS (fully local)** — No external service receives `MEMPOOL_TRANSACTIONS` with `BY_DEFAULT` or `ALWAYS` in any of the four transaction-related flows. No further checks.

4. _(From here, wallet sends pre-inclusion data to at least one external service by default or always.)_

5. **PASS (non-auctioning, non-extractive)** — Wallet does **not** auction orderflow by default, **and** every endpoint for those pre-inclusion recipients is verifiably non-extractive.

6. **PARTIAL (non-auctioning, extractive)** — Does not auction by default, but at least one pre-inclusion recipient endpoint is not verifiably non-extractive.

_Locked:_ Steps 7–12 apply **only** when the wallet **auctions orderflow by default**. Requires `transparency.orderflowPractices` and a worst-case `FeeDisplay` from `transparency.operationFees` (otherwise **UNRATED** at step 2). Wallets that only send pre-inclusion data for broadcast/simulation without default auctioning stop at step 5 or 6.

7. **FAIL (disclosure — UI)** — Orderflow disclosure in the transaction UI is `OrderflowDisclosureLevel.NONE`, or less prominent than fees (`compareOrderflowDisclosureToFeeDisplay(disclosure, worstFeeDisplay)` returns `< 0`).

8. **FAIL (disclosure — page link)** — No orderflow practices page when auctioning by default (`practicesPage` is `notSupported`, or no working Learn more link to the documented `url`).

9. **FAIL (page — entities)** — `practicesPage` supported but `contents.listsEntitiesAndWhatTheyDo` is false.

10. **FAIL (page — defaults)** — `explainsDefaultOrderflowAuctioning` is false.

11. **FAIL (page — user control)** — `documentsHowToChangeDefaults` is false when required (e.g. when `userCanRemoveAuctioning` is not supported, page must state defaults cannot be changed).

12. **PARTIAL (page — onchain verification)** — `onchainVerification` is `NO_METHOD_DOCUMENTED` or `METHOD_DOCUMENTED` but not `METHOD_DOCUMENTED_AND_EFFECTIVE`.

13. **PASS** — All applicable checks above satisfied.


polymutex:
Haven't reviewed the code at all, but the summary above is helpful.
`documentsHowToChangeDefaults` is interesting, not sure this was in the previous PRs but it's a good addition.

This attribute is going to be very difficult to explain in terms of its `ratingScale`. I'm wondering if we should support having a flowchart as an alternative for that section...

Will review in more details later, thanks for working on this so far :)

resources/docs/wallet-testing/data-collection/data-collection.md
```diff
@@ -110,3 +110,14 @@ export const someWallet: SoftwareWallet = {
...
 Once this is done, run the unit tests (`pnpm vitest`) to verify the integrity of the data. If everything passes, you are done! 🫡
+
+## Orderflow auctioning (`ORDERFLOW_AUCTION`)
+
+Use the `ORDERFLOW_AUCTION` data-collection purpose when an entity receives **pre-inclusion** transaction data (`MEMPOOL_TRANSACTIONS` with `BY_DEFAULT` or `ALWAYS`) **to auction orderflow** (MEV), as distinct from:
+
+- `TRANSACTION_BROADCAST` — plain broadcast or private-mempool routing without auctioning
+- `TRANSACTION_SIMULATION` — simulation / preview only
+
+**Validation:** A row must not list `ORDERFLOW_AUCTION` unless qualified `MEMPOOL_TRANSACTIONS` on that same row is `BY_DEFAULT` or `ALWAYS`. `OPT_IN` / `PROMPTED` mempool collection may coexist with `ORDERFLOW_AUCTION` on the row (optional auctioning).
```
polymutex, line 126: Can you add a unit test in `wallet-integrity.test.ts` for this?

src/schema/attributes/transparency/fee-transparency.ts
```diff
@@ -293,6 +221,21 @@ function computeWorstFees(feeTransparency: FeeTransparency): WorstFeeDisplay | n
 	return worstFeeTypes
 }
 
+/** Worst-case fee display among basic operation fees only. */
+export function worstOperationFeeDisplay(operationFees: BasicOperationFees): FeeDisplay | null {
```
polymutex, line 227: Hmm, seems maybe we should have `BasicOperationFees`'s fields be (a subset of) the keys from the `FeeType` enum, in order to avoid this type of redundancy... Maybe as a PR earlier than this one?

src/schema/features/transparency/fee-display.ts
```diff
@@ -125,3 +131,132 @@ export interface BasicOperationFees {
...
+			case 0:
+				worstOperationFees = {
+					feeDisplay: worstOperationFees.feeDisplay,
+					references: mergeRefs(worstOperationFees.references, refs(feeDisplay)),
+				}
+				break
+			case 1:
+				worstOperationFees = {
+					feeDisplay,
+					references: refs(feeDisplay),
```
polymutex, line 358: Interesting way to get the right `refs` (from *all* the worst-level fee displays of the same level) and no other. I wonder if we're doing something less sophisticated in other places...

src/schema/features/transparency/orderflow-derived.ts
```diff
@@ -0,0 +1,128 @@
...
+export const orderflowTransactionFlows = [
+	UserFlow.SEND_ETHER,
+	UserFlow.SEND_USDC,
+	UserFlow.NATIVE_SWAP,
+	UserFlow.MAKE_TRANSACTION,
+] as const
+
+type OrderflowTransactionFlow = (typeof orderflowTransactionFlows)[number]
+
+function flowDataCollection(
```
polymutex: Why does this function need to exist, rather than just `dataCollection[flow]`?

src/schema/features/transparency/orderflow-derived.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+function mempoolCollectedByDefaultOrAlways(
+	collectionByEntity: WithRef<DataCollectionByEntity>,
+): boolean {
+	const policy = qualifiedDataCollection(collectionByEntity.dataCollection)[
+		WalletInfo.MEMPOOL_TRANSACTIONS
+	]
+
+	return policy === CollectionPolicy.BY_DEFAULT || policy === CollectionPolicy.ALWAYS
```
polymutex: Use `collectedByDefault` from data-collection.ts

src/schema/features/transparency/orderflow-derived.ts
polymutex, line 1: I think these could just be in `orderflow.ts`, no separate file. Otherwise they're likely just going to get lost and forgotten when someone or something next looks at `orderflow.ts`. You've also already added other helper functions for orderflow stuff in `orderflow.ts`. Thoughts?

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,172 @@
...
+	 * Single line or toggle (e.g. "MEV protection", "Private tx") without full
+	 * breakdown.
+	 */
+	AGGREGATED = 'AGGREGATED',
+
+	/**
+	 * Clear block: auctioning disclosed, stats or kickback called out, Learn more
+	 * to practices page.
+	 */
+	COMPREHENSIVE = 'COMPREHENSIVE',
```
polymutex, line 35: I think there's a middle ground between these two (maybe `MENTIONED_WITH_EXTERNAL_EXPLANATION`? maybe there's a shorter way to say it), which is when the wallet has a nice-sounding "MEV protection" line (`MENTIONED`), but also has a "Learn more" link that explains the implications of this. `COMPREHENSIVE` implies something stronger.

How do we treat this middle ground? Not sure, probably the same as `MENTIONED`, but I think it's worth making the distinction at wallet-data-collection time nonetheless so that we have the flexibility to change how to treat that distinction later.

EDIT: Or maybe this could be represented by leaving this enum as 3-state, but distinguishing this at the `OrderflowDisclosure` level by setting `byDefault` to `MENTIONED` and `afterSingleAction` to `COMPREHENSIVE`? If so, that sounds fine to me as well, in which case no changes needed here.

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,172 @@
...
+ */
+export enum OrderflowDisclosureLevel {
+	/** No mention of auction / orderflow / MEV. */
+	NONE = 'NONE',
+
+	/**
+	 * Single line or toggle (e.g. "MEV protection", "Private tx") without full
+	 * breakdown.
+	 */
+	AGGREGATED = 'AGGREGATED',
```
polymutex: "Aggregated" doesn't sound right, since there are may be no numbers being aggregated here.  Note that it's not possible to "estimate" MEV ahead of time, so we can't expect this to be exactly displayed the way fees are (i.e. can't expect there to be a number).
Maybe this could be `MENTIONED`?

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/** Orderflow transparency practices beyond data-collection entity rows. */
+export type OrderflowPractices = WithRef<{
+	disclosure: WithRef<OrderflowDisclosure>
+	userCanRemoveAuctioning: Support
+	practicesPage: Support<
+		MustRef<{
+			url: string
+			contents: WithRef<OrderflowPracticesPageContents>
```
polymutex: `practicesPage` is already a `MustRef`, so there's no need to have `WithRef` here.

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,172 @@
...
+	pageLastUpdated: CalendarDate
+}
+
+/** Orderflow transparency practices beyond data-collection entity rows. */
+export type OrderflowPractices = WithRef<{
+	disclosure: WithRef<OrderflowDisclosure>
+	userCanRemoveAuctioning: Support
+	practicesPage: Support<
+		MustRef<{
+			url: string
```
polymutex: `url` is already present in `MustRef`, no need for this `url` field here.

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,172 @@
...
+	documentsHowToChangeDefaults: boolean
+	onchainVerification: OnchainVerificationDocumentation
+	/** Calendar date shown on the wallet's practices page (YYYY-MM-DD). */
+	pageLastUpdated: CalendarDate
+}
+
+/** Orderflow transparency practices beyond data-collection entity rows. */
+export type OrderflowPractices = WithRef<{
+	disclosure: WithRef<OrderflowDisclosure>
+	userCanRemoveAuctioning: Support
```
polymutex: I suggest making this a `Support<WithRef<{}>>`, then making `OrderflowPractices` no longer a `WithRef`. Then the references for each part of the orderflow practices are associated with the specific thing (disclosure / setting / practices page), no OrderflowPractices-top-level refs.

src/schema/features.ts
```diff
@@ -174,6 +175,9 @@ export interface WalletBaseFeatures {
 		/** Information on how fees are displayed for basic operations. */
 		operationFees: VariantFeature<Nullable<BasicOperationFees>>
 
+		/** Orderflow auctioning disclosure and practices page. */
+		orderflowPractices: VariantFeature<Nullable<OrderflowPractices>>
```
polymutex: Maybe this should only be in SoftwareWallet and EmbeddedWallet? Not sure what a hardware wallet could do about orderflow...

tests/orderflow-feature.test.ts
```diff
@@ -0,0 +1,380 @@
...
+			byEntity: exampleOrderflowAuctioneer,
+			dataCollection: {
+				[WalletInfo.MEMPOOL_TRANSACTIONS]: CollectionPolicy.BY_DEFAULT,
+				endpoint: RegularEndpoint,
+			},
+			purposes: [DataCollectionPurpose.TRANSACTION_BROADCAST],
+		}
+
+		// Incomplete research: endpoint removed before row is fully documented.
+		Reflect.deleteProperty(row.dataCollection, 'endpoint')
```
polymutex: Why is this worth defending against? Wouldn't the typesystem complain about such an object that doesn't have an `endpoint` field?

tests/orderflow-feature.test.ts
```diff
@@ -0,0 +1,380 @@
...
+	fullySponsored: false,
+}
+
+const fullySponsoredFeeDisplay: FeeDisplay = {
+	byDefault: FeeDisplayLevel.NONE,
+	afterSingleAction: FeeDisplayLevel.NONE,
+	fullySponsored: true,
+}
+
+describe('compareOrderflowDisclosureToFeeDisplay', () => {
```
polymutex, line 217: This seems overly complicated for what is a rather trivial function.

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+	displayName: 'Orderflow transparency',
+	wording: {
+		midSentenceName: null,
+		howIsEvaluated: "How is a wallet's orderflow transparency evaluated?",
+		whatCanWalletDoAboutIts: sentence(
+			'What can {{WALLET_NAME}} do to improve orderflow transparency?',
+		),
+	},
+	question: sentence(
+		'Does the wallet transparently disclose orderflow auctioning, or avoid sharing pre-inclusion transaction data with extractive external parties by default?',
```
polymutex: Very jargon-heavy. This `question` field is displayed at the top level of the page and needs to speak to a wallet user who doesn't know what MEV or orderflow is.

Suggest: "Does the wallet transparently disclose how it monetizes your transaction data?"

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+	),
+	why: markdown(`
+		Before a transaction is included on-chain, wallet software may send transaction
```
polymutex: onchain (please add something in `grammar.ts` similar to the `dApp` → `app` detector for this), maybe a separate PR

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+		data to external services for broadcast, simulation, or orderflow auctioning (MEV).
+		Users cannot see this path as clearly as on-chain execution unless the wallet
+		discloses it.
```
polymutex: \+ ", or Ethereum implements an encrypted mempool scheme" (hint for a later version of this attribute)

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+		They should also publish a practices page explaining who receives data and how
+		defaults can be changed.
+
+		Wallets that share pre-inclusion data without auctioning should route that data
+		through endpoints that are documented and independently verifiable as
+		non-extractive, so external parties cannot retain or exploit transaction data.
+	`),
+	methodology: markdown(`
+		Wallets are **exempt** from this attribute when they are hardware or embedded
+		wallets.
```
polymutex: Embedded wallets should not be exempt.

I guess it's OK to keep it exempt for embedded wallets for now since they're very underdeveloped, but the wording here should not say so because this is a temporary state of affairs, not the intention.

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+
+		Software wallets are **unrated** when Walletbeat has not yet researched where
+		the wallet sends transaction data before on-chain inclusion, which external
+		services receive it by default, or how those services handle it. This includes
+		common transaction flows such as sending Ether, built-in swaps, and general 
+		contract interactions, as well as cases where a default recipient is known 
+		but how data is handled at that endpoint is not.
```
polymutex: No need to mention this. Other attribute methodology doesn't really ever explain how "UNRATED" is arrived at, seems superfluous.

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,746 @@
...
+
+		Wallets that auction orderflow by default are also **unrated** when Walletbeat
+		lacks documented orderflow practices or enough fee-display information to compare
+		UI prominence across basic operations.
```
polymutex: Same

polymutex (changes requested):
Attribute methodology was surprisingly easy to understand despite how complex of a flowchart it actually is. Well done.

resources/docs/wallet-testing/data-collection/data-collection.md
polymutex: I see this may already be covered by the validation functions. If so, no need for a unit test.

johngrantuk: Yes, should already be covered.

src/schema/attributes/transparency/fee-transparency.ts
johngrantuk: I could look into this as a follow up PR if that works?

src/schema/features/transparency/orderflow-derived.ts
johngrantuk: It doesn't. I've updated.

src/schema/features/transparency/orderflow-derived.ts
johngrantuk: Agreed, I've merged in to orderflow.ts.

src/schema/features/transparency/orderflow.ts
johngrantuk: I find the byDefault=MENTIONED and afterSingleAction=COMPREHENSIVE naturally easy to understand. Do we need to add comments or anything to highlight that to a wallet reviewer?

src/schema/features.ts
johngrantuk: Yeah, that makes sense. I've left embedded because it will be developed in the future.

tests/orderflow-feature.test.ts
johngrantuk: You seem to be right that the typesystem would complain. I've removed the test and also some dead code.

tests/orderflow-feature.test.ts
johngrantuk: Simplified.

src/schema/attributes/transparency/orderflow-transparency.ts
johngrantuk: See #870

src/schema/features/transparency/orderflow.ts
polymutex: You're right, it's fine this way too. No changes necessary.

src/schema/features/transparency/orderflow.ts
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Validates that `OrderflowDisclosure` levels are consistent (parallel to fee display
+ * rules).
+ *
+ * - Rule 1: `afterSingleAction` cannot be `NONE` when `byDefault` is `MENTIONED` or
+ *   `COMPREHENSIVE`.
+ * - Rule 2: `byDefault` `COMPREHENSIVE` requires `afterSingleAction` `COMPREHENSIVE`.
+ *
+ * @throws When either rule is violated.
+ */
```
polymutex: This could be enforced at a typesystem level rather than a runtime check.

```typescript
export type OrderflowDisclosure = { /* ... same as before... */ } &  (
  // `afterSingleAction` must be at least as good as `byDefault`:
  (
   | { byDefault: 'NONE' }
   | { byDefault: 'MENTIONED', afterSingleAction: 'MENTIONED' | 'COMPREHENSIVE' }
   | { byDefault: 'COMPREHENSIVE', afterSingleAction: 'COMPREHENSIVE' }
  )
)
```

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,717 @@
...
+			),
+		},
+		details: mdSentence(
+			`{{WALLET_NAME}} does not auction orderflow by default, but under default settings sends pre-inclusion transaction data to ${links} through endpoints that are not documented and verifiable as non-extractive.`,
+		),
+		howToImprove: mdSentence(
+			'{{WALLET_NAME}} should route default pre-inclusion sharing through documented, independently verifiable non-extractive endpoints, or keep pre-inclusion transaction data fully local by default.',
+		),
+		impact: mdSentence(
+			'Users may not realize that external parties can retain or exploit their transaction data before it is included onchain.',
```
polymutex: "exploit" is a bit too strong. "monetize"?

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,717 @@
...
+	const names = entityNames(failingRecipients)
+	const links = entityLinks(failingRecipients)
+
+	ctx.addRef(...extractiveRows)
+
+	return ctx.build({
+		outcome: {
+			id: 'non_auctioning_extractive',
+			rating: Rating.PARTIAL,
+			displayName: 'Extractive pre-inclusion sharing',
```
polymutex: *Potentially*-extractive pre-inclusion sharing

src/schema/attributes/transparency/orderflow-transparency.ts
```diff
@@ -0,0 +1,717 @@
...
+		whatCanWalletDoAboutIts: sentence(
+			'What can {{WALLET_NAME}} do to improve orderflow transparency?',
+		),
+	},
+	question: sentence(
+		'Does the wallet transparently disclose how it monetizes your transaction data?',
+	),
+	why: markdown(`
+		Before a transaction is included onchain, or Ethereum implements an encrypted
+		mempool scheme, wallet software may send transaction data to external services
```
polymutex: > Before a transaction is included onchain, or Ethereum implements an encrypted mempool scheme

Woah :) While this is technically correct, those are very different-timeline ways to solve this problem. I meant that Encrypted mempools solve the orderflow monetization problem in the long term, and that a future version of the attribute would accept wallets using this mechanism as a way to pass this orderflow transparency attribute, but that's a bit too strong to mention this upfront here.

Suggest splitting as:

> Before a transaction is included onchain, wallet software may send transaction data to external services [...]

And then as a final paragraph:

> If the Ethereum protocol later implements a native mempool encryption mechanism, wallets may also protect their users from orderflow extraction by using this mechanism to shield transaction data before it is executed onchain.

Then maybe also as an extra paragraph in `methodology`:

> If the Ethereum protocol later implements a native mempool encryption mechanism, wallets that use this mechanism to shield transaction data before it is included onchain would also pass this attribute.

polymutex (changes requested):
Looking great, just wording changes below. The type definition change may also obviate the need for some of the tests.

src/schema/attributes/transparency/fee-transparency.ts
polymutex: Sure, also check out #884

polymutex (approved):

polymutex (commented):
Want to write a social media thread on this?

johngrantuk:
> Want to write a social media thread on this?

Yeah, I'll do that next week 👍

PR #808, opened 2026-06-10
Data: Update ambire's ERC 7730

maykelxyz:
References:
image
image
https://x.com/ambire/status/2064674667126984993


polymutex (approved):
You've already taken the time to take screenshots, so feels like you should include them as `ref`s in the repo now that we support that.

polymutex (approved):

PR #822, opened 2026-06-16
Bundle size improvements [WIP]

darrylyeo:


polymutex:
Love to see this

polymutex:
Should this be dropped? Looks like we're down to 2 MiB after the recent improvements done outside this PR.

PR #824, opened 2026-06-16
<Pie>: slice gradients

darrylyeo:
## Components

* `<Pie>`:
  * support parent slice gradients derived from child slice rating colors
  * use `radial-gradient(in oklch ...)` for perceptual color interpolation
  * place gradient stops using the attribute scoring weights
  * use a baked inverse-area lookup for the default rounded category petal shape, generated from the actual clipped geometry so outer-tip curvature is represented without doing the full area integration at render time
  * render unrated portions as transparent over the existing unrated gray

## Views

* `<WalletTable>`:
  * order parent-petal stops as unrated → fail → partial → pass
  * exclude exempt attributes from parent-petal gradients

---

Screenshot 2026-06-16 at 06 12 03

Screenshot 2026-06-16 at 06 25 42

### DIFF
--- a/src/components/Pie.svelte
+++ b/src/components/Pie.svelte
@@ -8,6 +8,11 @@
 		arcIconId?: string
 		titleText: string
 		href?: string
+		gradient?: {
+			areaRadiusStops?: number[]
+			colors: string[]
+			transparentStopColor?: string
+		}
 		children?: Slice[]
 	}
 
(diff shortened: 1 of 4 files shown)


polymutex (approved):
Man they look quite cool.
I'm really not a fan of how the grey looks though. But that's a reflection of us sucking at filling in the data, not a dig on this PR. Maybe Tere can play around to find another color that works better for the unrated part of the gradient.

PR #825, opened 2026-06-16
Attribute groups: fix Astro boundary serialization issues

darrylyeo:
Revert of the [revert](https://github.com/walletbeat/walletbeat/pull/823) to [walletbeat/walletbeat#608](https://github.com/walletbeat/walletbeat/pull/608), + fixes for issues it introduced related to JavaScript functions getting dropped during serialization while crossing the Astro component boundary.

* Restore [type param generalization](https://github.com/walletbeat/walletbeat/pull/608) for attribute groups and wallet data structures
* Instantiate interactive state entirely within Svelte, without passing a serialized representation from the Astro component boundary

## Pages

* `[walletName]`:
  * omit ladder definitions from hydrated wallet props so stage criteria are not serialized through Astro/Svelte island props

## Views

* New – `<SoftwareWalletTable>` / `<HardwareWalletTable>` / `<EmbeddedWalletTable>`:
  * introduce data context wrappers that instantiate `<WalletTable>` from within Svelte, avoiding Astro serialization

* `<WalletPage>`:
  * preserve attribute weights used by wallet-page pies after filtering exempt attributes
  * default to canonical software ladders client-side
  * accept wallet props without serialized ladder evaluations
  * keep variant/model selectors writable and preserve hash navigation when updating query params

* `<WalletAttributeSummary>`:
  * use the ladder type returned by `getWalletStageAndLadder()` instead of relying on evaluation object identity

## Data

* `getWalletStageAndLadder()`:
  * canonicalize serialized ladder evaluations against `allWalletLadders`
  * recompute ladder evaluations when wallet props omit serialized ladders
  * return the selected ladder type for stage UI lookups


eslint.config.mjs
```diff
@@ -79,6 +75,14 @@ export default [
...
 		},
 	},
+	{
+		files: ['data/**/*.ts', 'src/schema/attributes/security/bug-bounty-program.ts'],
+		rules: {
+			// CalendarDate is a template-literal union; object-property literals widen to string without `as CalendarDate`.
+			// typescript-eslint flags that cast as unnecessary even though tsc and astro check require it.
+			'@typescript-eslint/no-unnecessary-type-assertion': 'off',
+		},
+	},
```
polymutex: Seems too wide. Surely this can be worked around in other ways, e.g. using `as const` or just have a TypeScript predicate function taking in a string and returning a `CalendarDate`, and used that for the data files.

src/schema/attribute-groups.ts
```diff
@@ -1,445 +1,84 @@
...
-/** Evaluated self-sovereignty attributes for a single wallet. */
-export interface SelfSovereigntyEvaluations extends EvaluatedGroup<SelfSovereigntyValues> {
-	l1ProviderIndependence: EvaluatedAttribute
-	accountPortability: EvaluatedAttribute
-	permissionsManagement: EvaluatedAttribute
-	transactionInclusion: EvaluatedAttribute
-	chainVerification: EvaluatedAttribute
-	accountUnruggability: EvaluatedAttribute<AccountUnruggabilityMetadata>
+/** Attribute with hardcoded weight for use in attribute groups. */
+export type _AttributeWithWeight<
```
polymutex: Why the leading underscore? Seems that convention was to use this for abstract types, which this isn't. It's also not a private type, it's exported.

tests/attribute-grammar.test.ts
```diff
@@ -39,46 +57,37 @@ describe('attribute', () => {
...
-												),
+												ratings.map(async rating => await contentGrammarLint(rating.description)),
 											)
 										})
 									})
 								}
 
 								describe('rating scale', () => {
-									checkRatings('pass', passRatings)
+									if (isNonEmptyArray(passRatings)) {
```
polymutex: Should assert non-emptiness for the pass and fail ratings, like the previous version implicitly did

polymutex (approved):
Talk about a mega-PR. Please be very careful when rebasing.

polymutex (approved):

polymutex:
:guitar:

PR #826, opened 2026-06-16
Actions: parallelize checks and deploy steps

darrylyeo:
Speed up check scripts and action workflows by parallelizing work wherever applicable.

Screenshot 2026-06-16 at 08 46 00

## Scripts

* Run local checks and misc shell checks in parallel with [pnpm script regexes](https://pnpm.io/cli/run#running-multiple-scripts)
* Run CSS attribute generators in parallel

## Actions

* Bump workflow action dependencies to latest versions
* Update labels for steps and tool names across all action workflows
* Check workflow:
  * Split checks into parallel matrix jobs
* Deploy workflow:
  * Build `dist` once and share it through an artifact
  * Split Orbiter deploy and 4EVERLAND IPFS pinning into independent jobs


package.json
```diff
@@ -6,24 +6,28 @@
 	"scripts": {
 		"astro": "cross-env ASTRO_TELEMETRY_DISABLED=1 astro",
 		"build": "deploy/build.sh",
-		"check:all": "pnpm run check:astro && pnpm run check:vitest --run && pnpm run check:lint && pnpm run check:syntax && pnpm run check:spelling:quiet && pnpm run check:misc && pnpm run check:build:post:quiet",
-		"check:quick": "pnpm run check:lint:quick && pnpm run check:syntax && pnpm run check:spelling:quiet && pnpm run check:misc",
+		"check:all": "pnpm run \"/^check:(astro|vitest|lint|syntax|spelling:quiet|misc|build:post:quiet)$/\"",
+		"check:quick": "pnpm run \"/^check:(lint:quick|syntax|spelling:quiet|misc)$/\"",
```
polymutex, line 10: wat, does this work? Pretty wild

.github/workflows/deploy.yaml
polymutex, line 1: Can we keep the `deploy` workflow to still use `check:all`?
It's useful to have a dumb "just check all the things" step, both locally for steering agents, and for CI to check everything before deploy. Having CI have its own notion of what "checking everything" means increases the risk that we add something to `check:all` but forget to add it to the `deploy` workflow. Meanwhile, IMO it's OK for the `deploy` workflow specifically to be slow, since it's not blocking any PR or anything humans really need to wait on urgently.

package.json
```diff
@@ -32,11 +36,12 @@
 		"dev:background:restart": "deploy/dev/background-dev.sh --restart",
 		"lint": "pnpm run check:syntax --write && pnpm run check:lint --fix && pnpm run check:spelling:reorder",
 		"preview": "pnpm run astro preview",
-		"fix": "pnpm run lint && pnpm run check:treasury:markdown && pnpm run wallet-data-collection lint-fix && pnpm run check:features:markdown && pnpm run generate:css-attributes && pnpm run generate:wbicons",
+		"fix": "pnpm run \"/^(lint|check:(treasury:markdown|features:markdown)|generate:(css-attributes|wbicons)|wallet-data-collection:lint-fix)$/\"",
```
polymutex: Isn't there a risk for this one that the tools will step over each other?
e.g. `features:markdown` will copy contents that may get reformatted by `lint`.
Probably better to keep this one serial since it's not a read-only command.

polymutex (changes requested):

polymutex (approved):

PR #829, opened 2026-06-16
Layout improvements: logo, icons, header, footer, navigation, table of contents, scroll areas, sticky positioning

darrylyeo:
Styling / layout / animation pass covering logo, header, footer, navigation menus (mobile and desktop breakpoints, item states), scroll areas and sticky positioning behavior.

       

## Assets

* Add dark mode/light mode adaptive Walletbeat logo SVG using `@media (prefers-color-scheme)`
  * Add `[data-safari~='invert-lightness']` utility as a workaround for Safari which doesn't yet support this

## CSS attributes

* `[data-scroll-container]`:
  * hide scrollbars while keeping layout stable
  * constrain inline scroll containers so tables attach / detach against the base scroll area correctly
  * deepen sticky inset nesting through `[data-sticky-container]` levels

* `[data-sticky]`:
  * split backdrop placement (`backdrop-self` / `before` / `after`) from stuck / always conditions
  * inherit card background when sticky sits on `[data-card]`
  * fall back sticky backdrops when scroll-state containers are unsupported

* `[data-icon]`:
  * render icons as `inline-flex` globally
  * share circular icon sizing across nav, header controls, and footer
  * keep literal `data-icon` / list-marker content ASCII-safe

* `[data-badge]`:
  * restore content gap after row-primitive cleanup

* `label:has(> input)`:
  * support search inputs where the wrapping label owns the pressable boundary
  * avoid nested pressable effects from the child input

## Views

* `src/layouts/Layout.astro`:
  * restore original CSS grid layout with an explicit `Content` grid area
  * mount a detached footer in the root grid without making it sticky
  * when mobile nav shell is open, translate content backward in the z-direction
  * add accessible first-tab "Skip to content" link for keyboard/switch-control users

* `<Navigation>`:
  * rebuild the sticky header + flexible menu shell
  * use the new adaptive SVG logo
  * keep the sticky logo bar layered above scrolled menu items
  * keep header controls as shared circular icons
  * add mobile nav menu popover transition with `backdrop-filter`, `clip-path` and `translate` in the z-direction
  * rename menu items:
    * "Wallet EIPs" → "Standards"
    * "Wallet Playground" → "Playground"
  * restore disclaimer as a sticky bottom card with a subtler theme-derived background
  * rename `currentPathname` to `currentHref`

* `<NavigationItems>`:
  * keep primary and secondary nav groups under the same search
  * align sticky offsets through the sticky container hierarchy
  * render mobile search at the bottom of the nav flow
  * re-token item colors, radii, and sticky summary backgrounds
  * highlight group links via `scroll-target-group` / `:target-current` bubbling
  * render rating icons through the emoji variant where appropriate
  * place search bar at bottom on mobile breakpoints

* `<Footer>`:
  * add social links
  * use the new adaptive SVG logo

* Home:
  * restore the detached hero layout

* `<WalletPage>`:
  * rebuild the TOC as a sticky header + sibling content shell
  * reserve TOC width in page sticky insets without inheriting that clearance into the TOC itself
  * keep section headings sticky with backdrop tokens

* `<WalletTable>`:
  * remove detached filter behavior that was fighting horizontal table scrolling
  * stop marking detached headers sticky

## Components

* `<Table>`:
  * apply sticky backdrop tokens on table headers / sticky columns

* `<Pie>`:
  * clip local paint overflow without creating competing scroll containers
  * render pie icons as emoji


src/middleware.ts
```diff
@@ -106,6 +106,7 @@ const shouldShieldByExtension = {
 	gif: false,
 	jpg: false,
 	jpeg: false,
+	ico: false,
```
polymutex: Fixed this one in #881.

polymutex:
Planning to start merging this?

src/views/Footer.astro
```diff
@@ -0,0 +1,88 @@
...
+] as const
+---
+
+<footer data-row='center wrap align-center gap-6' data-scroll-item='inline-detached'>
+	<a href='/' data-link='camouflaged' data-row-item='wrap-center' aria-label='Walletbeat'>
+		<img src='/logo.svg' alt='Walletbeat' data-safari='invert-lightness' />
+	</a>
+
+	<p data-row-item='wrap-center basis-6'>
+		made with ❤️ + 🌸 by{' '}
```
polymutex, line 31: ❤️ + 🌸

src/views/WalletTable.svelte
```diff
@@ -1976,7 +1974,7 @@
 		display: none;
 	}
 
-	@media (max-width: 768px) {
+	@media (max-width: 1024px) {
```
polymutex, line 1977: Feels like this width should be a var

polymutex (approved):

src/views/WalletTable.svelte
darrylyeo: Not possible in CSS yet unfortunately, hopefully eventually if they implement custom media queries:

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@custom-media

PR #833, opened 2026-06-17
Data: Fill Rainbow securityBestPractices (browser + mobile)

ren2140eth:
Fills the previously-`null` `security.securityBestPractices` for Rainbow, for both shipping variants (no desktop app).

## Data filled

| Variant | keyStorage | secureRng | Hardening source |
|---|---|---|---|
| **browser** | `ENCRYPTED_WITH_USER_SECRET_STANDARDIZED_KDF` — keychain vault encrypted with the user password via `@metamask/browser-passworder` (PBKDF2, 600k iterations) | `OS_CSPRNG` | published Chrome Web Store manifest |
| **mobile** | `HARDWARE_SECURITY_MODULE` — seed/private key in `react-native-keychain`, biometric/passcode-gated and wrapped by an RSA key in the OS keystore / Secure Enclave | `OS_CSPRNG` | `AndroidManifest.xml` + `Info.plist` from `rainbow-me/rainbow` |
| **desktop** | — | — | `NOT_A_DESKTOP_APP` |

Manifest URLs were added to Rainbow's metadata and the files collected via `pnpm collect:manifests`.

## New Android permission ratings (apply globally)

Rainbow's manifest declared four permissions not previously catalogued, so they're added to the `AndroidPermission` enum and rated in the evaluator:

- `POST_NOTIFICATIONS` → **Pass** (just the ability to notify).
- `READ_EXTERNAL_STORAGE` → **Fail** (broad read of user files; mirrors the existing iOS photo-library-read = Fail).
- `WRITE_EXTERNAL_STORAGE` → **Fail**, pessimistically: on Android ≤ 18 (API) the storage permission group also granted implicit read, and the manifest doesn't pin the user's OS version, so we assume the read-capable case.
- `USE_FULL_SCREEN_INTENT` → **Fail**: lock-screen full-screen UI is unnecessary for a wallet and is a phishing vector (auto-granted on Android < 14).

## ⚠️ Notable side-effect for reviewers: Android manifest parser fix (2nd commit)

While collecting Rainbow's manifest I found it declares `USE_FULL_SCREEN_INTENT` with `tools:node="remove"` — a manifest-merger directive that **strips** the permission from the shipped app (a dependency adds it; Rainbow removes it). The parser was ignoring merger directives and counting such permissions as if they were requested.

The 2nd commit makes `parseAndroidManifest` skip `<uses-permission>` entries carrying `tools:node="remove"`. Permissions a wallet actually grants are declared *without* the directive and are still captured, so this only ever drops genuinely-removed permissions. Regenerating drops `USE_FULL_SCREEN_INTENT` from Rainbow's parsed data.

**Blast radius is Rainbow only.** MetaMask is the only other wallet with an Android manifest; its two `tools:node="remove"` directives are on a `<uses-permission-sdk-23>` element (not parsed) and a `<receiver>` (not a permission), so its data is unchanged and `collect:manifests --check` stays green.

## Note on the rating outcome

Because the attribute is scored worst-of, Rainbow lands at **Fail on deployment hardening for both variants regardless** — browser (`UNRESTRICTED` host permissions + `web_accessible_resources`) and mobile (`SYSTEM_ALERT_WINDOW` + `READ_EXTERNAL_STORAGE`). The key-storage/RNG classifications are accurate data points but don't change the final rating.

## Checks
prettier · eslint (0 errors) · `astro check` (0 errors) · cspell · `collect:manifests --check` · `manifest-integrity.test.ts` — all green.


src/schema/attributes/security/security-best-practices.ts
```diff
@@ -599,6 +606,12 @@ const androidPermissionRatings: Record<AndroidPermission, Rating.PASS | Rating.F
 	[AndroidPermission.ACCESS_FINE_LOCATION]: Rating.FAIL,
 	// Modifying global audio settings can suppress security alerts or notifications from other apps.
 	[AndroidPermission.MODIFY_AUDIO_SETTINGS]: Rating.FAIL,
+	// Broad read of shared storage exposes private files.
+	[AndroidPermission.READ_EXTERNAL_STORAGE]: Rating.FAIL,
+	// Writes to shared storage. On Android <= 18 (API) this permission group also granted
```
polymutex: Android API 18 was released in 2013. I think we can assume users are using newer versions now. It is OK for a wallet to write to external storage. The line below should be changed to `Rating.PASS` and the comment updated.

polymutex (changes requested):

src/schema/attributes/security/security-best-practices.ts
ren2140eth: Got it, updated.

polymutex (approved):

PR #607, opened 2026-03-23
Add Uniswap Calibur and Rainbow Calibur delegation contracts

DanielSinclair:
## Summary

Adds EIP-7702 delegation contracts for Uniswap Wallet and Rainbow Wallet, plus a comprehensive Rainbow wallet entry.

### New wallet contracts

- **`data/wallet-contracts/uniswap-calibur.ts`** — [Calibur](https://github.com/uniswap/calibur) (`0x000000009b1d0af20d8c6d0a44e162d11f9b8f00`), Uniswap's EIP-7702 delegation contract
- **`data/wallet-contracts/rainbow-calibur.ts`** — [RainbowCalibur](https://github.com/uniswap/calibur) (`0x612373d7003d694220f7800eeaf8e3924c0951d3`), Rainbow's EIP-7702 delegation contract

### New entity and contributor

- **`data/entities/rainbow.ts`** — Rainbow corporate entity (`CorporateEntity & WalletDeveloper`) with jurisdiction, privacy policy, farcaster, twitter, and repo URL
- **`data/contributors/danielsinclair.ts`** — New contributor with Rainbow employee affiliation

### Rainbow wallet (`data/software-wallets/rainbow.ts`)

Added comprehensive feature data:

- **EIP-7702**: Supported via RainbowCalibur contract
- **EOA**: `canExportPrivateKey: true`, BIP32/BIP44/BIP39 key derivation with seed phrase export
- **Address resolution**: Non-chain-specific ENS via chain client; ERC-7828 and ERC-7831 not supported
- **Chain abstraction**: Built-in bridging with comprehensive fee display; per-chain ETH and USDC balance views; global and per-chain account value
- **Chain configurability**: L1 RPC endpoint configurable; `nonL1: notSupported`
- **Ecosystem delegation**: Bundled with first EIP-7702 operation; no cross-chain gas or wallet sponsorship
- **Browser integration**: EIP-1193 / EIP-2700 / EIP-6963; `walletCall: notSupported`
- **Hardware wallet support**: Ledger (WebUSB + Bluetooth), Trezor (WebUSB)
- **Scam alerts**:
  - `contractTransactionWarning`: Blockaid transaction simulation; leaks contract address, user address, user IP
  - `scamUrlWarning`: Blockaid dApp scanning; leaks IP and domain only
  - `sendTransactionWarning: notSupported`
[...]

PR #835, opened 2026-06-18
Data: Fill Rainbow chainAbstraction (cross-chain balances + bridging)

ren2140eth:
Fills the previously-`null` \`ecosystem.chainAbstraction\` attribute for Rainbow. Rating: **PARTIAL**.

All values were verified first-hand in the Rainbow **mobile** app (Rainbow is mobile-first), with the fee/risk model cross-checked against Rainbow's support docs. Supporting screenshots are included under \`public/references/wallets/rainbow/screenshots/\` and attached to the relevant refs.

### Cross-chain balances
- **Global account value** — supported: a single portfolio total summing all chains.
- **Per-chain token balances** — supported: the same asset on multiple chains is shown as separate badged rows (e.g. ETH on mainnet + ETH on an L2).
- **Cross-chain sum per token** — *not* supported: Rainbow does not collapse a token's balance into one combined cross-chain total (the asset appears as multiple rows). → this is what caps the rating at PARTIAL.

### Bridging
- **Built-in bridging** — supported (in-app Bridge/Swap).
- **Fees** — `AGGREGATED` by default, `COMPREHENSIVE` after one action: the amount-entry screen shows only a gas estimate by default (and "Hold to Bridge" is available there), so a user can confirm without seeing the Rainbow fee; tapping "Review" reveals the itemized breakdown (destination network, minimum received, "Included Rainbow Fee", max slippage).
- **Risks** — `NOT_IN_UI`: no bridge trust-assumption / cross-chain risk warning in the flow, and the support article omits it too.
- **Suggested bridging** — not supported: sending more of a token than is held on one chain (while holding enough on another) just blocks with "Insufficient Funds" rather than offering to bridge the shortfall.

Supersedes the stale \`chainAbstraction\` data in #607 (which mirrored MetaMask's values); this is verified against the live app instead.


data/software-wallets/rainbow.ts
```diff
@@ -139,7 +139,76 @@ export const rainbow: SoftwareWallet = {
...
 			}),
 		},
-		chainAbstraction: null,
+		chainAbstraction: {
+			bridging: {
+				builtInBridging: supported({
+					ref: [
+						{
+							explanation:
+								'Rainbow has a built-in cross-chain bridge/swap feature. On the amount-entry screen only a gas estimate is shown by default (and "Hold to Bridge" is available there), so the fee shown by default is aggregated. Tapping "Review" (a single action) reveals the itemized breakdown: destination network, minimum received, the "Included Rainbow Fee", and max slippage as separate line items. The UI does not, however, explain the trust assumptions or risks of bridging across chains (no warning about external bridge providers, L2 risk, or possible loss of funds), and the Rainbow support documentation likewise omits these. Verified in the Rainbow mobile app.',
```
polymutex: Once the other comment about putting all this under the `[Variant.Mobile]` subfield, then there is no need to specifically mention "Verified in the Rainbow mobile app" because the `ref` will be scoped to the mobile version only. Same for other `ref.explanation` fields in this PR. It is however OK to still use the phrase "Rainbow mobile app" rather than just "Rainbow app" or "Rainbow" when used in a sentence for clarity, just redundant to have a whole sentence about this.

data/software-wallets/rainbow.ts
```diff
@@ -139,7 +139,76 @@ export const rainbow: SoftwareWallet = {
...
+							lastRetrieved: '2026-06-17',
+							url: 'https://rainbow.me/support/app/bridge-and-swap-tokens',
+						},
+						{
+							explanation:
+								'Rainbow mobile app bridge amount-entry screen: by default only a gas estimate is shown (labelled "Free", roughly 0.0001 ETH), with the "Included Rainbow Fee" and slippage not visible. "Hold to Bridge" is available here, so a user can confirm without ever seeing the itemized fee.',
+							file: 'public/references/wallets/rainbow/screenshots/2026-06-17-chain-abstraction-bridge-input-gas-only.jpg',
```
polymutex: File-based `ref`s should probably always have a `label`. I'll add code to enforce this. In the meantime, please make sure to add a `label` for all these `file` `ref`s. It should be a text description of what the image shows, not an explanation of why the image is relevant (as that's the purpose of the `explanation` field). For example, label: "Rainbow mobile app screenshot of bridging UI".

data/software-wallets/rainbow.ts
```diff
@@ -139,7 +139,76 @@ export const rainbow: SoftwareWallet = {
 				medium: 'CHAIN_CLIENT',
 			}),
 		},
-		chainAbstraction: null,
+		chainAbstraction: {
```
polymutex, line 142: Since all of these findings are mobile-UI-specific, they should all be under the `chainAbstraction.[Variant.MOBILE].[...]` key, not straight under `chainAbstraction` (unless also verified to be the same on the browser extension version, but the `ref`s in this PR don't currently suggest this).

polymutex (changes requested):
Thanks again for the checks :)

polymutex:
(Also, while any screenshot is better than no screenshot, PNGs are always preferred, as JPGs are lossy)

ren2140eth:
sounds good will take a look and revise

ren2140eth:
Thanks for the review! Pushed 408bc228 addressing the three code comments:

- **Variant scoping:** moved the whole `chainAbstraction` block under `[Variant.MOBILE]`, since all findings are mobile-UI-specific and not verified on the extension.
- **Labels:** added a `label` to each file `ref` describing what the screenshot shows (e.g. "Rainbow mobile app screenshot of the bridge amount-entry screen…").
- **Redundant sentences:** dropped the "Verified in the Rainbow mobile app" sentences from the explanations now that the mobile variant scope conveys that.

On JPG vs PNG: these were captured on-device as JPGs and I don't have lossless originals, so re-encoding to PNG would just wrap already-lossy data without quality gain. Leaving them as JPGs for now — happy to re-capture as PNG in a follow-up if you'd prefer.

polymutex (approved):

Issue #848, opened 2026-06-24
wbicons font problems


polymutex:
Icons are broken:

<img width="2913" height="1647" alt="Image" src="https://github.com/user-attachments/assets/0491239b-487f-4dfe-8e21-8bcd6873add5" />

If I revert the Astro font thing and just leave the emoji codepoint thing, icons are still partially broken:

<img width="2700" height="1642" alt="Image" src="https://github.com/user-attachments/assets/3e77bc19-c276-4671-a551-24a3bd314ba9" />

darrylyeo:
Fixed by https://github.com/walletbeat/walletbeat/pull/850

polymutex:
Still looks off-center on both:

<img width="787" height="319" alt="Image" src="https://github.com/user-attachments/assets/6c960a77-f3e2-45db-9dd0-99a4ab43c230" />

And there's the ℹ️ problem on Firefox (though this is with an old version of Firefox):

<img width="787" height="319" alt="Image" src="https://github.com/user-attachments/assets/c66a1155-1207-4a86-89f4-bd2be2298124" />

darrylyeo:
Addressed icon alignment in https://github.com/walletbeat/walletbeat/pull/829

PR #850, opened 2026-06-25
Icons: improve emoji-to-SVG icon mapping, icon usage

darrylyeo:
Follow-up to https://github.com/walletbeat/walletbeat/pull/828 + fixes for issue https://github.com/walletbeat/walletbeat/issues/848.

Screenshot 2026-06-25 at 08 24 49

Screenshot 2026-06-25 at 09 55 16

Screenshot 2026-06-25 at 09 55 30

---

### Fonts
* re-enable `wbicons` SVG icon font for use in navigation menus

### CSS attributes

* `data-icon`:
  * overload with explicit variant tokens for icon font and mapped glyph names
  * replace all usages; set `wbicon` icon font or `emoji` explicitly on a per-case basis

### Icons

* `wbicons`:
  * handle multi-codepoint glyph mappings correctly during font generation
  * update emoji mappings to more closely match SVG icons and historical emoji usage
  * add check to guard against duplicate emoji assignments

### Views

* `<WalletTable>`:
  * show emoji in `title` attributes instead of the mapped icon string
  * restore flower icon `filter` by deleting previously accidentally reverted override
  * reduce icon font size

* `<WalletPage>`:
  * rework icon + circle styling in attribute details and table of contents


src/schema/attributes.ts
```diff
@@ -1013,7 +1013,7 @@ export const formatAttributeTitleText = (
 	attribute: EvaluatedAttribute<OutcomeMetadata>,
 	suffix?: string,
 ) =>
-	`${attribute.evaluation.outcome.icon ?? attribute.attribute.icon} ${
+	`${attribute.evaluation.outcome.icon ?? wbIconEmojiSequences[attribute.attribute.icon]} ${
```
polymutex: Shouldn't this be `wbIconEmojiSequences[attribute.evaluation.outcome.icon ?? attribute.attribute.icon]`?

src/tools/icon-font-generator/icon-font-generator-lib.ts
polymutex, line 1: I don't understand what a lot of this file is doing now. There's new font-ish concepts like sfnt, glyphs, maxp, alignment tables, "cmapFormat14", etc. Could use a good pass of comments.

polymutex (approved):

polymutex (approved):

PR #855, opened 2026-06-27
Data: Rabby

maykelxyz:
## Rabby data
- Added supported hardware wallets
- Filled up remaining null transaction legibility
- Added permissions management for mobile wallets
  - Updated ERC-1155 -> CANNOT_INSPECT. Tested it out and I can't inspect.
## Other changes
- Added CoolWallet to cspell
- Removed desktop variant for Rabby (currently still WIP as mentioned in their website)

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -87,6 +87,7 @@
 		"Consigny",
 		"Containerfile",
 		"contentscript",
+		"CoolWallet",
 		"copyable",
 		"coreutils",
 		"Cormoran",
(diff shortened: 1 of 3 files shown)


.cspell.json
```diff
@@ -87,6 +87,7 @@
 		"Consigny",
 		"Containerfile",
 		"contentscript",
+		"CoolWallet",
```
polymutex, line 90: ?

polymutex (approved):

.cspell.json
polymutex: OK, I see that's a real wallet name and shows up in refs. I was using it as a wallet pseudonym this whole time. Whoops.

PR #858, opened 2026-06-28
Astro: re-enable wbicons font

darrylyeo:
## Astro

* Re-enable the wbicons font registration so nav icons render through the SVG font again


polymutex:
Nice, though the icons are off-center from the circle, and the "About" icon still displays as the emoji on Firefox for some reason

PR #862, opened 2026-06-30
base-app: fill scamUrlWarning (first-hand verified, FULL_URL leak)

ren2140eth:
## What

Fills `security.scamAlerts.scamUrlWarning` for Base App (was `null`).

Base App warns on flagged dapps via a Coinbase blocklist, and — more importantly for the leak fields — it checks **every** in-app-browser navigation by sending the URL to Coinbase.

| field | value |
|---|---|
| `leaksVisitedUrl` | **`FULL_URL`** |
| `leaksUserAddress` | `false` |
| `leaksIp` | `true` |

## How this was verified

Drove **Base App v30.1.0 (`org.toshi`)** in an Android-15 emulator with **mitmproxy installed in the system trust store**, and navigated its in-app dapp browser. On every navigation the app issues a GraphQL persisted query `useBlocklistQueryQuery` to `graphql-base.coinbase.com`:

```
GET https://graphql-base.coinbase.com/query
  ?operationName=useBlocklistQueryQuery
  &variables={"domain":"https://app.uniswap.org/swap"}
```

The `variables.domain` field carries the **entire visited URL — path and query string included** — not just the domain and not a privacy-preserving hash. A second lookup sends the bare origin.

**Proof it's a live decrypted capture:** I typed `https://example.com/path?leaktest=SECRET12345` into the browser, and that exact marker came back in the request:

```
variables={"domain":"https://example.com/path?leaktest=SECRET12345"}
```

So:
- the **full URL** (incl. query string) is sent to Coinbase infrastructure → `leaksVisitedUrl: FULL_URL`, `leaksIp: true`
- **no `0x` address** appears in the request (the check runs before any wallet connection) → `leaksUserAddress: false`

## Note for reviewers

`FULL_URL`. Capture artifacts (raw flow log + screenshots + the `SECRET12345` marker test) are available on request and the setup is fully reproducible.

### DIFF
--- a/data/software-wallets/base-app.ts
+++ b/data/software-wallets/base-app.ts
@@ -20,7 +20,10 @@ import {
 	MultiPartyKeyReconstruction,
 } from '@/schema/features/security/keys-handling'
 import { PasskeyVerificationLibrary } from '@/schema/features/security/passkey-verification'
-import type { ContractTransactionWarning } from '@/schema/features/security/scam-alerts'
+import type {
+	ContractTransactionWarning,
+	ScamUrlWarning,
+} from '@/schema/features/security/scam-alerts'
 import { SpendingApprovalsControl } from '@/schema/features/self-sovereignty/permissions-management'
 import {
 	TransactionSubmissionL2Support,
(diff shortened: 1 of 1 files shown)


ren2140eth:
### Durable identifiers (for independent verification)

- **Endpoint:** `https://graphql-base.coinbase.com/query`
- **Operation:** `useBlocklistQueryQuery`
- **persistedQuery sha256Hash:** `9d984064743db2a6b709145bb7300137a88e93106626b40c0dd62174708c0f79`
- **Leak field:** `variables.domain` = the full visited URL (scheme + host + path + query)

### Reproduce it yourself

1. Put mitmproxy in front of Base App (a phone with the mitmproxy CA trusted is enough — no emulator required).
2. In the in-app browser, open any URL with a distinctive query param, e.g. `https://example.com/?pcheck=hunter2`.
3. Look for the request to `graphql-base.coinbase.com` with `operationName=useBlocklistQueryQuery`; confirm your `pcheck=hunter2` appears verbatim in `variables.domain`. That confirms `FULL_URL`.

polymutex:
So you set up `mitmproxy` on your own (or, with Claude)? For Android? I'm very curious on the setup you've arrived at.

ren2140eth:
Yep - got claude's help with getting it all set up it's an android emulator (AVD) + mitmproxy all on the desktop

polymutex:
In that case, that is yet another amazing finding. However I still think we should hold off on this PR, for a different reason: we should get the full data dump of the wallet through all its UX flows, which will contain the data needed to prove this conclusively. Don't want to publish this without being able to show evidence. Source code would be evidence too, but since Base App is closed-source we can't do that. Let's focus on this case when we next meet to capture this.

ren2140eth:
sounds good can chat on it in the next meeting

PR #863, opened 2026-06-30
Data/zerion

maykelxyz:
### Zerion data added
- Licensing
- Keys handling
- Security best practices
- Release transparency

### Other changes
- Added `cooldown` to cspell


data/software-wallets/zerion.ts
```diff
@@ -321,11 +354,25 @@ export const zerion: SoftwareWallet = {
...
 			releaseTransparency: {
-				artifactSigning: null,
-				dependencyLocking: null,
-				dependencyVulnerabilityScanning: null,
-				hasPublicChangelog: null,
-				hermeticBuilds: null,
+				artifactSigning: supported<ArtifactSigningDetails>({
+					ref: 'https://github.com/zeriontech/zerion-wallet-extension/releases',
+					publication: 'GITHUB_RELEASE',
+					signer: 'DEVELOPER_KEY',
```
polymutex: Is it signed? I don't see a `.asc` or something on the releases page

data/software-wallets/zerion.ts
```diff
@@ -321,11 +354,25 @@ export const zerion: SoftwareWallet = {
...
+				}),
+				dependencyLocking: supported({
+					ref: [
+						{
+							explanation:
+								'The README documents a `.npmrc` `min-release-age=15` supply-chain cooldown that restricts `npm install` to dependency versions published more than 15 days ago, applied to every install "here and in CI."',
```
polymutex: That by itself is not dependency locking. It probably does use dependency locking since it has a [package-lock.json](https://github.com/zeriontech/zerion-wallet-extension/blob/main/package-lock.json) but the PR's current `ref` seems unrelated

data/software-wallets/zerion.ts
```diff
@@ -321,11 +354,25 @@ export const zerion: SoftwareWallet = {
...
+							url: 'https://github.com/zeriontech/zerion-wallet-extension#readme',
+						},
+					],
+				}),
+				dependencyVulnerabilityScanning: notSupported,
+				hasPublicChangelog: supported({
+					ref: 'hhttps://github.com/zeriontech/zerion-wallet-extension/releases',
```
polymutex: Double "h" in "hhttps". Please add a test that catches this in `tests/github-ref-commit-hash.test.ts` or similar

polymutex (changes requested):

.cspell.json
```diff
@@ -193,6 +193,7 @@
 		"HD",
 		"Headshot",
 		"hesterbruikman",
+		"hhttps",
```
polymutex: wat

If this is because the test file itself has `hhttps`, then I think the problem is the test file, not that this non-word should be added to cSpell (which causes it to be tolerated in all source code and text we do intend to be correct).

You can use a different scheme in the test file than `hhttps` that is legit but not in `ALLOWED_SCHEMES`, such as `ftp`. Or remove the tests using this, as per the other comment.

tests/malformed-url-scheme.test.ts
```diff
@@ -0,0 +1,115 @@
...
+		const malformed = findMalformedUrls(
+			"ref: 'hhttps://github.com/zeriontech/zerion-wallet-extension/releases',",
+		)
+
+		expect(malformed).toHaveLength(1)
+	})
+
+	it('the detector accepts a well-formed URL', () => {
+		expect(findMalformedUrls("ref: 'https://github.com/walletbeat/walletbeat'")).toHaveLength(0)
+	})
```
polymutex: These tests are testing the test code itself, which is a bit too meta, no? You can remove.

src/schema/attributes/privacy/app-isolation.ts
```diff
@@ -228,7 +228,7 @@ export const appIsolation: Attribute = {
 		On the web, website \`A\` is not allowed to query your browsing history from
 		website \`B\` by default. This ensures your browsing activity remains private
 		across websites. This principle is enshrined in the HTTP protocol as the
-		[**Same-Origin Policy**](hhttps://en.wikipedia.org/wiki/Same-origin_policy):
+		[**Same-Origin Policy**](https://en.wikipedia.org/wiki/Same-origin_policy):
```
polymutex, line 231: Very nice. Shows the value of the test :)

tests/malformed-url-scheme.test.ts
```diff
@@ -0,0 +1,115 @@
...
+	return malformed
+}
+
+describe('URLs in authored source must be well-formed', async () => {
+	const fileMap = new Map<string, MalformedUrl[]>()
+
+	for (const dir of CHECKED_DIRS) {
+		await crawlCodebase({
+			root: `${REPO_ROOT}${dir}/`,
+			ignore: [],
```
polymutex: You should use `commonExclusion`s here, and you should add your own so that the set of files read is limited to the ones ending in `CHECKED_EXTENSIONS`. This speeds up the test by not needing to read in files that the checker doesn't actually care about. By the time it gets to `fullTraversalFn`, every file has been read, so it is not the appropriate place for filtering.

polymutex (changes requested):

tests/malformed-url-scheme.test.ts
```diff
@@ -81,14 +95,6 @@ describe('URLs in authored source must be well-formed', async () => {
...
-		const malformed = findMalformedUrls(
-			"ref: 'hhttps://github.com/zeriontech/zerion-wallet-extension/releases',",
-		)
-
-		expect(malformed).toHaveLength(1)
-	})
-
 	it('the detector accepts a well-formed URL', () => {
 		expect(findMalformedUrls("ref: 'https://github.com/walletbeat/walletbeat'")).toHaveLength(0)
 	})
```
polymutex: This one is also just a "testing the test" test.

polymutex (approved):

PR #864, opened 2026-06-30
Rate wallet service fee units in fee transparency (#847)

### LINKED ISSUE #847: Fee transparency attribute should require percentage or bps as unit
Probably requires a new feature field, so marking as pre-launch, but should hopefully be easy.

Rationale: Percentages or bps are the only measure of fee-taking that makes sense across all types of orders and are directly comparable across wallets. Alternatives like absolute values can be hard to mentally compute, and wallets could otherwise do tricks to denominate them in unintuitive units.

Inspired by #807

johngrantuk:
Closes #847 

## Summary

Extends fee transparency to distinguish **wallet service fees** (platform fees on built-in swap/bridge flows) from **network and external protocol fees**, and to rate whether those service fees are shown in comparable units (%/bps vs absolute amounts).

- Add `WalletServiceFeeDenomination` to `FeeDisplay` (`PERCENTAGE_OR_BPS`, `ABSOLUTE_OR_OTHER`, `NOT_APPLICABLE`, or `null` when unresearched). Only evaluated on built-in swap and cross-chain bridging flows.
- Replace `comprehensiveFeesShownByDefault` with flow-specific presets: `comprehensiveGasOrExternalFees`, `comprehensiveWalletServiceFeesUnresearched`, and `comprehensiveWalletServiceFeesPercentage`.
- Update fee transparency evaluation: **UNRATED** when denomination is unresearched, **PARTIAL** when wallet service fees use absolute/other units, and use denomination as a tiebreaker when comparing fee displays.
- Move `validateFeeDisplay` to the feature module with new invariants for denomination.
- Update wallet data for 12 wallets and document the new field in `resources/docs/features.md`.


src/schema/features/transparency/fee-display.ts
```diff
@@ -66,21 +88,113 @@ export interface FeeDisplay {
...
+	 *
+	 * - `PERCENTAGE_OR_BPS` — wallet service fee line item(s) use % or bps.
+	 * - `ABSOLUTE_OR_OTHER` — wallet service fee shown in flat $, fixed token
+	 *   amount, etc.
+	 * - `NOT_APPLICABLE` — no wallet service fee on this flow. Use for L1
+	 *   transfers, external app transactions, relayer fees, and other flows
+	 *   where only network or external protocol fees apply.
+	 * - `null` — not yet researched (required on in-scope swap/bridge flows).
+	 */
+	walletServiceFeeDenomination: WalletServiceFeeDenomination | null
```
polymutex: I think this is overfitting to the rating methodology. Features should be able to express what the wallet does and to what someone reviewing a wallet would want to enter to reflect the truth about the wallet, not something that is specifically attuned to the rating methodology. Specifically:

- Wallets probably display either percentage or bps, not both, so they should be separate enum values (treating these two values the same in the attribute is a methdology-side concern)
- Some wallets probably display fees in both raw values and relative terms, so it's important to be able to express this, e.g. by making this an array of fee denominations (dealing with the multiplicity, again, is a methodology-side concern)

.cspell.json
```diff
@@ -463,6 +463,7 @@
 		"UniSat",
 		"uniswap",
 		"unproxied",
+		"Unresearched",
```
polymutex: Lowercase, as it is a common noun. Uppercase and lowercase are treated differently by Harper, which these words are fed into.

src/schema/features/transparency/fee-display.ts
```diff
@@ -66,21 +88,113 @@ export interface FeeDisplay {
...
+	ref: [],
+}
+
+/**
+ * Shorthand for comprehensive built-in swap or cross-chain bridging fee display
+ * pending wallet service fee denomination research. Use on `builtInErc20Swap` and
+ * `feesLargerThan1bps` when display levels are known to be comprehensive but
+ * whether wallet service fees use %/bps vs flat amounts is not yet researched.
+ */
+export const comprehensiveWalletServiceFeesUnresearched: WithRef<FeeDisplay> = {
```
polymutex: This seems like a convenience `const` that is only useful temporarily while the data hasn't been collected yet. It also muddies the property we have today that grepping "null" in the `data/` subdirectory will yield a list of fields that need to be further researched.

I think it is better to delete this const, and either inline it where it's currently used, or use something like

```typescript
export const comprehensiveWalletServiceFeesUnknownDenomination: Omit<WithRef<FeeDisplay>, 'walletServiceFeeDenomination'> = {
	byDefault: FeeDisplayLevel.COMPREHENSIVE,
	afterSingleAction: FeeDisplayLevel.COMPREHENSIVE,
	fullySponsored: false,
	ref: [],
}
```

Then in data files:

```typescript
builtInErc20Swap: supported({
  ...comprehensiveWalletServiceFeesUnknownDenomination,
  walletServiceFeeDenomination: null,
}),
```

polymutex (changes requested):

src/schema/features/transparency/fee-display.ts
```diff
@@ -66,21 +90,89 @@ export interface FeeDisplay {
...
 }
 
-/** Shorthand for fees that are comprehensive by default. */
-export const comprehensiveFeesShownByDefault: WithRef<FeeDisplay> = {
+/**
+ * Shorthand for comprehensive fee display on flows with no wallet service fee
+ * (e.g. L1 transfers, Uniswap approval popups). Do not use on `builtInErc20Swap`
+ * or `feesLargerThan1bps` without explicit denomination research.
+ */
+export const comprehensiveGasOrExternalFees: WithRef<FeeDisplay> = {
```
polymutex: Sorry for the back and forth on this, but with this const name, I worry that a wallet would put this in their data without realizing the implications of what this means for `walletServiceFeeDisplayUnits`. It also prevents the ability to specify a `ref`, useful for screenshotting the fee being displayed.

I suggest either:

- Remove this const, inline it everywhere it is presently used
- Create a partial const that lacks the `walletServiceFeeDisplayUnits` and `ref` fields, to force it to be populated explicitly in the wallet data file
- Turn this into a function that accepts a `WithRef<{ displayUnits: FeeDisplay['walletServiceFeeDisplayUnits'] }` argument and returns a `WithRef<FeeDisplay>`, essentially the same as the previous suggestion but usage in wallet data files would be function-style.

... Or basically anything you can think of that satisfies the properties that what data is being specified in the data file is obvious just by reading the data file (without reading the expanded object here), and that it's obvious if a `ref` is missing (when looking at the data file alone).

`comprehensiveWalletServiceFeesPercentage` below is a bit better because it is more obvious that it means percentage-based fees, though it also has the same `ref` problem.

For the purpose of attribute files needing example feature data as part of their rating scale, it's OK to define a helper const without `ref`s there, but that can be in the attribute file itself so that it remains private and unimportable into wallet data files.

polymutex (approved):
Good to merge after the following :)

src/schema/features/transparency/fee-display.ts
johngrantuk: No worries, makes a lot of sense. I've just removed the const and inlined as I think its clearer and the "helper" const wasn't that helpful!

Issue #591, opened 2026-03-15
Account recovery: Wallet should encourage user to practice recovering their account


polymutex:
From 1TS security benchmark. Wallet should initially then periodically offer a "account recovery drill" feature that has the user walk through the recovery process to exercise their muscle memory.

maykelxyz:
Done through: #867

PR #867, opened 2026-07-01
Feature: Account recovery drills

maykelxyz:
# Feature: Account recovery drills (1TS security benchmark #591)

Wallet should initially then periodically offer an "account recovery drill" feature that has the user walk through the recovery process to ensure that, in case needed, users can actually "recover" their account/s. Folded under `accountRecovery`.

## Schema changes
- Added `AccountRecoveryDrills` interface with three check-ups:
  - `periodicPrivateKeyQuiz` (`boolean`) — periodically confirms the user still has their private key.
  - `periodicSeedPhraseQuiz` (`boolean`) — periodically confirms the user still has their seed phrase.
  - `periodicGuardianAccountCheck` (`boolean | 'NO_GUARDIAN_RECOVERY'`) — confirms the user still has access to their guardian accounts. `NO_GUARDIAN_RECOVERY` lets wallets without guardian recovery opt out of this check.
- Added `drills: Support<AccountRecoveryDrills> | null` to `AccountRecovery`. `null` means not yet rated.

## How Account Recovery is now rated
The attribute now evaluates guardian recovery **and** drills, then takes the worst of the two.

- `drills === null` → **UNRATED**.
- **Drills** on their own:
  - All three check-ups present → **PASS**.
  - Some check-ups present → **PARTIAL** (`howToImprove` lists the missing ones).
  - Drills not supported → **PARTIAL**.
- **Guardian recovery** keeps its existing rating; a wallet with no guardian recovery still **FAILs** that half.
- Final rating = worst of guardian recovery and drills, so a wallet needs both solid guardian recovery and all three drills to fully pass.

## Data
- Added `drills` to the software wallet templates (`completed.tmpl.ts`, `unrated.tmpl.ts`) and filled in drills data across software wallets.
- Regenerated `resources/docs/features.md` to document the new interface and field.


src/schema/features/security/account-recovery.ts
```diff
@@ -473,6 +473,29 @@ export function guardianPolicyMarkdown(guardianPolicy: GuardianPolicy): string {
...
+	 * they still have the seed phrase for their account?
+	 */
+	periodicSeedPhraseQuiz: boolean
+
+	/**
+	 * Does the wallet ask users whether they still have access to their
+	 * guardian accounts to ensure that they can recover their account?
+	 */
+	periodicGuardianAccountCheck: boolean | 'NO_GUARDIAN_RECOVERY'
+}
```
polymutex, line 508: I think this type of schema kind of suggests that a wallet should implement both. Additionally, because it is wrapped in `Support<>` below, this also creates the possibility of:

```typescript
drills: supported({ ... all fields false... })
```

... which wouldn't make sense. And it also forces the `periodicGuardianAccountCheck` to always be present even for wallets that don't have guardian recovery (which we have precedent for, but we should avoid in general when we can help it). Lastly, this doesn't have a `ref` we could attach.

Instead I think this can be something like:

```typescript
export enum AccountRecoveryDrillType {
  SEED_PHRASE,
  GUARDIAN_TEST,
  ...
}
export interface AccountRecoveryDrill {
  type: AccountRecoveryDrillType
  reminderEveryNDays: number
}

// Feature field:
    drills: Support<NonEmptyArray<WithRef<AccountRecoveryDrill>>> | null
```

This way it's either `null`, or unsupported, or supported with at least one type of drill, and each drill type is free to have its own reference.

Yes this makes attribute methodology more complicated to deal with, but that's OK, that's a downstream consequence that shouldn't be an input as to how the feature fields are defined.

polymutex (changes requested):

src/schema/attributes/security/account-recovery.ts
```diff
@@ -117,35 +122,110 @@ function evaluateGuardianRecoveryPolicy(
...
+				{{WALLET_NAME}} does not periodically remind users to verify
+				they can still access their account recovery.
 			`),
-			metadata: {
-				minimumGuardianPolicy: null,
-				outcomes: null,
-			},
+			metadata: { minimumGuardianPolicy: null, outcomes: null },
 		},
 		details: accountRecoveryDetailsContent({}),
```
polymutex: What gets displayed by `accountRecoveryDetailsContent` in this case?

src/schema/attributes/security/account-recovery.ts
```diff
@@ -117,35 +122,110 @@ function evaluateGuardianRecoveryPolicy(
...
+					shortExplanation: sentence(`
+						{{WALLET_NAME}} does not implement guardian-based account recovery.
+						The user will lose access to their account if they lose their seed phrase.
+					`),
+					metadata: {
+						minimumGuardianPolicy: null,
+						outcomes: null,
+					},
+				},
+				details: accountRecoveryDetailsContent({}),
```
polymutex, line 316: Same

src/schema/attributes/security/account-recovery.ts
```diff
@@ -280,11 +362,98 @@ export const accountRecovery: Attribute<AccountRecoveryMetadata> = {
...
+				),
+			),
+		],
+		partial: [
+			exampleRating(
+				paragraph(`
+						The wallet securely distributes a recovery secret across at least 3
+						external services with client-side reconstitution, but does not
+						periodically remind users to verify they can still access their
+						account recovery.
```
polymutex: "access their account recovery" → "recover their account"

src/schema/attributes/security/account-recovery.ts
```diff
@@ -294,6 +463,8 @@ export const accountRecovery: Attribute<AccountRecoveryMetadata> = {
 					the recovery process.
 					The recovery secret is reconstituted on the user's device using
 					2 or more shares from these external services.
+					The wallet also periodically reminds users to verify their recovery
+					setup is intact.
```
polymutex: intact → functional

polymutex (changes requested):

src/schema/attributes/security/account-recovery.ts
maykelxyz: I updated AccountRecoveryMetadata, added drills property

```typescript
export type AccountRecoveryMetadata = {
	minimumGuardianPolicy: GuardianPolicy | null
	outcomes: NonEmptyArray<GuardianScenarioOutcome<GuardianScenarioType>> | null
	drills: {
		configured: AccountRecoveryDrill[]
		missing: AccountRecoveryDrillType[]
	} | null
}
```

Which now shows what is missing and configured in `AccountRecoveryDetails.svelte`

src/schema/attributes/security/account-recovery.ts
```diff
@@ -111,41 +122,144 @@ function evaluateGuardianRecoveryPolicy(
...
-	if (isSupported(accountRecovery.guardianRecovery)) {
-		return evaluateGuardianRecoveryPolicy(
-			ctx,
-			accountRecovery.guardianRecovery.minimumGuardianPolicy,
-		)
+	// Guardian account check-ups are only meaningful for wallets that
+	// actually implement guardian-based recovery.
+	const recommendedDrillTypes: AccountRecoveryDrillType[] = [
+		AccountRecoveryDrillType.PRIVATE_KEY_QUIZ,
+		AccountRecoveryDrillType.SEED_PHRASE_QUIZ,
```
polymutex: Why should a wallet implement both a private key quiz and a seed phrase quiz? Seems redundant.

polymutex: How about wallets that don't use seed phrases, e.g. MPC? Seems like they shouldn't have to implement seed phrase drills, since they can't.

IMO this should look for only one specific type of drill depending on the wallet's default account type.

src/schema/attributes/security/account-recovery.ts
```diff
@@ -167,10 +281,21 @@ export const accountRecovery: Attribute<AccountRecoveryMetadata> = {
 		(also known as "Social recovery") helps make self-custody safe and practical
 		for everyday users. Properly implemented, this keeps users safer while still
 		providing the self-sovereignty benefits of self-custody in the day-to-day.
+
+		But a recovery mechanism that is never exercised can silently become useless.
+		A guardian account gets abandoned. A written-down seed phrase becomes
+		unreadable or gets thrown out. The user simply forgets which of several
+		backup methods they actually set up. None of this is discovered until
+		the user actually needs to recover their account, at which point it is
```
polymutex: Feels AI-written and doesn't really match the writing style of the previous paragraphs

src/schema/features/security/account-recovery.ts
```diff
@@ -514,4 +548,24 @@ export interface AccountRecovery {
...
+	[AccountRecoveryDrillType.PRIVATE_KEY_QUIZ]: 'private key check-ups',
+	[AccountRecoveryDrillType.SEED_PHRASE_QUIZ]: 'seed phrase check-ups',
+	[AccountRecoveryDrillType.GUARDIAN_ACCOUNT_CHECK]: 'guardian account check-ups',
+}
+
+export const accountRecoveryDrillNouns: Record<AccountRecoveryDrillType, string> = {
+	[AccountRecoveryDrillType.PRIVATE_KEY_QUIZ]: 'private key',
+	[AccountRecoveryDrillType.SEED_PHRASE_QUIZ]: 'seed phrase',
+	[AccountRecoveryDrillType.GUARDIAN_ACCOUNT_CHECK]: 'guardian accounts',
 }
```
polymutex: Merge this into a single `export function accountRecoveryDrillWording(type AccountRecoveryDrillType): { label: string, noun: string } { switch(type) { ... } }` function, to group the labels and nouns for each type together.

polymutex (changes requested):

src/schema/attributes/security/account-recovery.ts
maykelxyz: Hmm, I think it would come in for accounts that are imported through private keys. In my testing setup, I have accounts imported through:

1) Seed phrase
2) Private keys

So wallets should verify that users still have the seed phrase for accounts that are imported through the seed phrase and the private keys for accounts that are imported through private keys.

Thinking about it, I can also change the enum type from

`AccountRecoveryDrillType. PRIVATE_KEY_QUIZ ` to `AccountRecoveryDrillType.PRIVATE_KEY_CHECK_UP`.

maykelxyz: That makes sense, refactoring it to:
```ts
// Account support data is also required, to determine which specific
// account recovery drills are expected of the wallet.
if (ctx.features.security.accountRecovery === null || ctx.features.accountSupport === null) {
	return unrated(ctx, { minimumGuardianPolicy: null, outcomes: null, drills: null })
}
```

and a new util function to getRecommendedDrills regarding the accounts supported & if guardian recovery is supported

```ts
function getRecommendedDrillTypes(
	accountSupport: AccountSupport,
	hasGuardianRecovery: boolean,
): AccountRecoveryDrillType[] {
	const keyMaterialDrillTypes: AccountRecoveryDrillType[] = isAccountTypeSupported<AccountTypeEoa>(
		accountSupport.eoa,
	)
		? [
				...(accountSupport.eoa.canExportPrivateKey
					? [AccountRecoveryDrillType.PRIVATE_KEY_QUIZ]
					: []),
				...(accountSupport.eoa.keyDerivation.type === 'BIP32'
					? [AccountRecoveryDrillType.SEED_PHRASE_QUIZ]
					: []),
			]
		: []

	return [
		...keyMaterialDrillTypes,
		...(hasGuardianRecovery ? [AccountRecoveryDrillType.GUARDIAN_ACCOUNT_CHECK] : []),
	]
}
```

src/schema/attributes/security/account-recovery.ts
```diff
@@ -111,41 +127,217 @@ function evaluateGuardianRecoveryPolicy(
...
+	// For empty recommended drill types
+	if (!isNonEmptyArray(recommendedDrillTypes)) {
+		return ctx.build({
+			outcome: {
+				id: 'no_applicable_recovery_drills',
+				rating: Rating.PASS,
+				displayName: 'No account recovery drills applicable',
+				shortExplanation: sentence(`
+					{{WALLET_NAME}}'s account type has no user-held recovery material
+					that periodic check-ups could verify.
```
polymutex, line 215: What kind of wallet would this apply to?

src/schema/attributes/security/account-recovery.ts
```diff
@@ -111,41 +127,217 @@ function evaluateGuardianRecoveryPolicy(
...
+				rating: Rating.PARTIAL,
+				displayName: 'Incomplete account recovery drills',
+				shortExplanation: sentence(`
+					{{WALLET_NAME}} does not run all recommended periodic
+					account recovery check-ups.
+				`),
+				metadata: { minimumGuardianPolicy: null, outcomes: null, drills: { configured, missing } },
+			},
+			details: accountRecoveryDetailsContent({}),
+			howToImprove: sentence(`
```
polymutex: This should be more specifically state that the wallet should have such drill *if* the user created an account by that method (e.g. only do private key recovery drills if the user imported a private key, not a seed phrase for example)

src/schema/attributes/security/account-recovery.ts
```diff
@@ -111,41 +127,217 @@ function evaluateGuardianRecoveryPolicy(
...
+				they can still recover their account.
 			`),
 			metadata: {
 				minimumGuardianPolicy: null,
 				outcomes: null,
+				drills: { configured: [], missing: recommendedDrillTypes },
 			},
 		},
 		details: accountRecoveryDetailsContent({}),
+		howToImprove: sentence(`
```
polymutex: Same

polymutex (changes requested):

src/schema/attributes/security/account-recovery.ts
polymutex: If the answer is "none", then this should be enforced, i.e. `getRecommendedDrillTypes` should return a `NonEmptyArray`, and this branch can stop existing. We should have no known-dead code.

maykelxyz: Well, wallets with no user-held key material and no guardian recovery. I would imagine these are wallets that only support ERC-4337 accounts and __no guardian recovery__.

In our current data, example would be Elytro, Safe, & Daimo being populated with `guardianRecovery: notSupported` & `drills: notSupported`.

polymutex: OK, though that leaves the question of how would users even recover their wallets in this case? But hopefully the other branches of this attribute's logic cover this anyway, so I think we're good.

polymutex (approved):

PR #873, opened 2026-07-01
Add Markdown-based documents to the site

polymutex:


polymutex:
@darrylyeo Is this sane or should I be looking at Astro's native markdown support?

Goal is to still be able to view files as normal on GitHub so there has to be some parsing magic to rewrite URLs to match both cases.

(Note I haven't gone over the CSS to unslopify it yet)

darrylyeo:
Hmm yeah, maybe worth upgrading to [Astro 7](https://astro.build/blog/astro-7/#markdown--mdx-in-rust) and reimplementing as a Markdown AST transform in a Sätteri plugin.

https://docs.astro.build/en/guides/markdown-content/#using-s%C3%A4tteri-plugins-and-features
https://satteri.bruits.org/docs/plugins/

polymutex:
Why is Astro 7 required? Or just to avoid rewriting the thing to Sätteri after the 7.0 jump?

Issue #877, opened 2026-07-02
Data collection: allow multi-entity (operator + intermediaries) mapping per domain


ren2140eth:
## Problem

A single domain can expose request data to **more than one** entity with different privacy meaning. Example surfaced during the Base App capture: `graphql-base.coinbase.com` is served through Cloudflare, so data sent to it is seen by:

- **Coinbase** — the intended recipient / data controller (operator).
- **Cloudflare** — TLS-terminating infrastructure that sees the plaintext in transit, but is not the intended recipient.

These are not equivalent for a privacy assessment: *"the wallet vendor's backend received your address"* is a different statement from *"a CDN terminated the TLS on the way there."* Today `entity-domains` maps each domain to exactly one entity, forcing a lossy choice.

## Current model

- **Data:** `data/entities/domains/entity-domains.json` — `Record<domain, EntityId>`.
- **Resolver:** `data/entities/domains/entity-domains.ts` → `entityForDomain(domain): Entity | null` (exact, else suffix match).
- **CLI:** `mark-domain --domain=... --entity=<EntityId>` writes one pair.
- **Downstream (already multi-entity):** `schema/features/privacy/data-collection.ts` models a flow's recipients as `DataCollectionForFlow.collected: WithRef<DataCollectionByEntity>[]`. So the *output* side already expresses multiple entities receiving data from one flow — the gap is purely upstream (resolution + CLI + a role distinction).

## Proposed model

Keep the JSON backward-compatible via a string shorthand, add a structured form encoding **one operator + zero-or-more intermediaries**:

```ts
type DomainMapping =
  | EntityId                                          // shorthand: sole operator (current behaviour)
  | { operator: EntityId; intermediaries?: EntityId[] }

interface ResolvedDomain { operator: Entity; intermediaries: Entity[] }
function entitiesForDomain(domain: string): ResolvedDomain | null
// entityForDomain kept as a thin wrapper returning `.operator` so existing call sites do not break.
```

Downstream, tag each receipt with its role so the assessment can weight them:

```ts
interface DataCollectionByEntity {
  byEntity: Entity
  role?: "OPERATOR" | "INTERMEDIARY"   // NEW; absent = OPERATOR
  dataCollection: EndpointCollection
  purposes: NonEmptyArray<DataCollectionPurpose>
}
```

[...]

PR #882, opened 2026-07-03
Data collection: multi-entity (operator + intermediaries) domain mappings

ren2140eth:
Implements #877, following the design feedback there. Point-by-point:

- **Single unambiguous representation**: a mapping in `entity-domains.json` is either a bare `EntityId` (sole operator), or `{ operator, intermediaries }` where `intermediaries` is typed `NonEmptyArray<EntityId>` and cannot be absent. The validator rejects empty lists, duplicates, unsorted lists, an operator that is also an intermediary, and unknown entity IDs, with a per-domain error message. The CLI writers canonicalize (dedupe + sort) before writing and re-validate the whole file on save; a mapping whose intermediary list becomes empty is automatically written back in the bare form.
- **`entityForDomain` deleted**: no wrapper kept. The six call sites were updated case-by-case: `check`/`review-requests` gates only test mapping existence; request/referer display now renders `Domain (Operator via Intermediary, ...)`; capture-to-`DataCollection` conversion emits one receipt per resolved entity with its role.
- **`role` unconditionally populated**: `DataCollectionByEntity.role: EntityRole` (`OPERATOR | INTERMEDIARY`) is required; all 29 existing hand-authored entries across Rabby/Daimo/imToken/Ambire/BitBox and the completed template are backfilled as `OPERATOR`.
- **Intermediaries affect the grade**: the converter attributes the same collected data and purposes to each intermediary as to the operator, as separate `collected` entries, so downstream attributes (address correlation, privacy hygiene, ...) weigh them exactly like operators.
- **`mark-domain-update`**: new subcommand with `--set-operator`, `--set-intermediaries`, `--add-intermediaries`, `--remove-intermediaries` (the last two composable, `--set-intermediaries` exclusive with them). Adding an already-present intermediary or removing an absent one errors. `mark-domain` gains `--intermediaries` for first-time marking.
- **Entity catalogue**: added `EntityType.infrastructureProvider` and a Cloudflare entity typed `CorporateEntity & InfrastructureProvider` (multi-type, per existing precedent). Fastly/Akamai can be added the first time a capture actually needs them.
- **Resolution precedence**: exact match first, then the *longest* matching parent domain (previously first-in-key-order), so an intermediary recorded on `sub.example.com` also applies to `deep.sub.example.com` even when `example.com` is mapped. Intermediaries are recorded on the exact fronted hostname, not the apex.
- **Tests**: `tests/entity-domains.test.ts` covers the validator invariants and resolution precedence. `mark-domain`/`mark-domain-update` were also exercised end-to-end against the real JSON file (structured write, add/remove, revert-to-shorthand, operator-in-intermediaries rejection).

Deferred as agreed:
- v2 follow-up: `check` issue category flagging `cf-ray` / `server: cloudflare` response headers on domains whose mapping lacks the corresponding intermediary (no quiesce mechanism).
- v3 follow-up: `detect-intermediaries` subcommand + one-off backfill pass over previously-mapped domains.


data/entities/domains/entity-domains.ts
```diff
@@ -1,87 +1,195 @@
...
-	return entityId
+	return mapping
+}
+
+const domainToEntityIdMapping = assertValidDomainToEntityIdMapping(rawData)
+
+/**
+ * The set of entities that see the data sent to a domain.
+ */
+export interface ResolvedDomain {
```
polymutex: `operator` should be the first field here.

data/entities/domains/entity-domains.ts
```diff
@@ -1,87 +1,195 @@
...
+ * Look up the domain mapping for a domain within a given mapping.
+ * Exact matches take precedence over parent-domain (suffix) matches, so a
+ * CDN-fronted subdomain can carry intermediaries without asserting them for
+ * the whole apex domain.
  */
-function entityIdForDomain(domain: string): EntityId | null {
-	if (Object.hasOwn(domainToEntityIdMapping, domain)) {
-		return domainToEntityIdMapping[domain]
+export function domainMappingIn(
+	mapping: DomainToEntityIdMapping,
```
polymutex: Why pass in `mapping`? Use the global `domainToEntityIdMapping` unless there's a good reason I'm missing

src/tools/wallet-data-collection/wallet-capture-file.ts
```diff
@@ -1980,35 +1984,48 @@ export class WalletCaptureFile {
...
+					role: EntityRole.INTERMEDIARY,
+				})),
+			]
 
-			for (const [info, policy] of userInfos.entries()) {
-				if (policy === null) {
-					maybeThrow(`Cannot figure out collection policy of request: ${request.toString()}`)
-					continue
+			for (const { entity, role } of receivingEntities) {
+				const perEntityKey = `${assertValidEntityId(entity.id)} ${role}`
```
polymutex: Use a `:` as separator, not a space

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
```diff
@@ -854,76 +885,235 @@ export async function handleMarkFlowUnsupported(opts: MarkFlowUnsupportedOptions
...
+	for (const key of Object.keys(mapping).toSorted()) {
+		sortedDomains[key] = mapping[key]
 	}
 
+	// Re-validate to enforce all invariants before writing.
+	assertValidDomainToEntityIdMapping(sortedDomains)
+
+	// Save file with formatting
+	await fs.promises.mkdir(path.dirname(domainFile), { recursive: true })
+	await fs.promises.writeFile(domainFile, JSON.stringify(sortedDomains, null, '\t') + '\n', 'utf8')
```
polymutex: Write to a temp file next to it, then rename to overwrite atomically.

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
```diff
@@ -854,76 +885,235 @@ export async function handleMarkFlowUnsupported(opts: MarkFlowUnsupportedOptions
...
-				`Domain '${targetDomain}' is a subdomain of already marked '${existingDomain}' (${existingDomains[existingDomain]}). You do not need to mark it separately.`,
+			if (domainMappingEquals(existingMapping, targetMapping)) {
+				throw new Error(
+					`Domain '${targetDomain}' is a subdomain of already marked '${existingDomain}' with the same mapping (${domainMappingToString(existingMapping)}). You do not need to mark it separately.`,
+				)
+			}
+
+			if (domainMappingOperator(existingMapping) !== domainMappingOperator(targetMapping)) {
+				throw new Error(
+					`Domain '${targetDomain}' is a subdomain of already marked '${existingDomain}' (${domainMappingToString(existingMapping)}), which has a different operator. Fix the parent domain mapping instead.`,
```
polymutex: "Fix the parent domain mapping instead, or remove it in case it was made in error."

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
```diff
@@ -854,76 +885,235 @@ export async function handleMarkFlowUnsupported(opts: MarkFlowUnsupportedOptions
...
+	}
+
+	if (
+		opts.setOperator === null &&
+		opts.setIntermediaries === null &&
+		opts.addIntermediaries === null &&
+		opts.removeIntermediaries === null
+	) {
+		throw new Error(
+			'Nothing to update; specify at least one of --set-operator, --set-intermediaries, --add-intermediaries, --remove-intermediaries.',
```
polymutex, line 1064: Should also verify that if `--set-intermediaries` is set, then `--add-intermediaries and `--remove-intermediaries` should not (otherwise it's ambiguous what should happen)

polymutex (changes requested):
Fable is hot stuff. This is near perfect. Unlike other agents I've seen, this one is much better at following existing codebase conventions.

data/entities/domains/entity-domains.ts
ren2140eth: Done in 66e203b9 — `operator` now comes first in `DomainMapping`, `ResolvedDomain`, and all object literals. Since the `sort-keys-custom-order` rule in `data/**` would otherwise force alphabetical order, `operator` was added to `firstOrderedKeys` in `eslint.config.mjs` (no other `operator` keys exist under `data/`, so no collateral reordering), which makes operator-first the enforced order rather than an exception.

data/entities/domains/entity-domains.ts
ren2140eth: The reason: `mark-domain` / `mark-domain-update` must resolve against the freshly-read working copy of the JSON file they are editing (e.g. the parent-suffix hint in `mark-domain-update`), which the module-level mapping loaded at import time does not reflect mid-edit. `entitiesForDomainIn` is additionally parameterized so the unit tests can exercise resolution precedence (exact-before-suffix, longest-parent-wins) with synthetic mappings instead of depending on whatever the committed data file happens to contain. Both docstrings now state this (66e203b9), and `entitiesForDomain` remains the API for lookups against committed data. Happy to fold `entitiesForDomainIn` away if you would rather the tests run against the real JSON only.

src/tools/wallet-data-collection/wallet-capture-file.ts
ren2140eth: Done in 66e203b9.

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
ren2140eth: Done in 66e203b9 — temp file next to the target, then `rename`, same pattern as `saveCaptureFileOnly`.

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
ren2140eth: Done in 66e203b9.

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
ren2140eth: This is enforced — it is the first check in `handleMarkDomainUpdate`, at the top of this hunk: `--set-intermediaries` combined with either `--add-intermediaries` or `--remove-intermediaries` throws before anything is read. (`--add` + `--remove` together remain allowed, since adds are applied before removes and both are strict about already-present/absent entries.) Let me know if you would like the check moved into option processing instead.

data/entities/domains/entity-domains.ts
polymutex: Please revert the `eslint.config.mjs` change. It's OK if literals have the `operator` field later, my comment was just regarding the interface definition.

data/entities/domains/entity-domains.ts
polymutex: Run against a global object, have a `updateDomainMapping` function that the subcommands and tests can call as needed.

polymutex (changes requested):

data/entities/domains/entity-domains.ts
ren2140eth: Done in c8d0e642. Reverted the `firstOrderedKeys` change; object literals are back to alphabetical order, and the two type definitions keep `operator` first via targeted `eslint-disable` comments.

One thing this surfaced: `eslint-plugin-sort-keys-custom-order` 2.2.1 hard-crashes under ESLint 10 on *any* sort-keys violation (`context.getSourceCode()` was removed; the crash happens while the report is generated, so `eslint-disable` directives never get a chance to apply). I bumped it to 3.0.1, pinned exactly — 3.0.2's published tarball is missing the `dist/index.d.ts` its package.json declares, which makes the import error-typed. If you'd rather not take the dependency bump in this PR, the fallback is alphabetical key order in the type definitions too (the per-field docs already state which one is the operator) — say the word and I'll switch to that.

data/entities/domains/entity-domains.ts
ren2140eth: Done in c8d0e642. `domainMappingIn` / `entitiesForDomainIn` are gone; `domainMappingForDomain` and `entitiesForDomain` now resolve against the module-level mapping, and `updateDomainMapping()` replaces it: `readDomainMappingFile` calls it after reading the working copy (so the `mark-domain-update` parent-suffix hint sees fresh state), and the tests call it with synthetic mappings, restoring the committed mapping afterwards.

data/entities/domains/entity-domains.ts
ren2140eth: For the record: the 3.0.2 packaging defect (type declarations emitted to `dist/src/` while `types`/`exports` still point to `dist/index.d.ts`) is now reported upstream as hugoattal/eslint-plugin-sort-keys-custom-order#10 — that's the reason for the exact `3.0.1` pin. If a fixed 3.0.3 lands, the pin can be lifted.

ren2140eth: Upstream fixed the packaging in 3.0.3 (verified: types at the declared path, rules/disable-directives/--fix all behave; hugoattal/eslint-plugin-sort-keys-custom-order#10). The exact `3.0.1` pin stays for now regardless — this repo's `minimumReleaseAge: 20160` gates 3.0.3 until ~2026-07-19 — after which the pin can be relaxed back to a caret range.

polymutex: `eslint-disable sort-keys-custom-order/type-keys` should not be necessary. No order interface definition in the codebase needs this exclusion despite having non-sorted fields. e.g. look at `WalletBaseFeatures` for a counterexample.

Please remove the `eslint-disable` directive and revert the package version back to 2.2.1.

Edit: Ah, I see the problem, this is because the rule only applies to .ts files under `data/`. Well, the introduction of this interface means we're about at the point where we have outgrown this file and that this should probably be its own library, e.g. somewhere under `src/schema/entity-domains.ts`. Please create and move the interface type definitions there, as well as all the related functions that don't deal with the global singleton database of entity/domain mappings (e.g. the `assertValidDomainToEntityIdMapping` function definition should move, `updateDomainMapping` should not).

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
```diff
@@ -854,76 +886,243 @@ export async function handleMarkFlowUnsupported(opts: MarkFlowUnsupportedOptions
...
-			existingDomains = assertValidDomainToEntityIdMapping(JSON.parse(content) as unknown)
+			mapping = assertValidDomainToEntityIdMapping(JSON.parse(content) as unknown)
 		} catch (e) {
 			throw new Error(`Failed to parse entity domains file: ${getErrorMessage(e)}`, { cause: e })
 		}
 	}
 
+	// Later lookups (e.g. `domainMappingForDomain`) must resolve against the
+	// working copy just read, not the mapping loaded at import time.
+	updateDomainMapping(mapping)
```
polymutex: A `read` function should not have a side effect like this. Instead, move this call to whenever the mapping data file is overwritten.

polymutex (changes requested):

data/entities/domains/entity-domains.ts
ren2140eth: Done in f86a883b. The type definitions (`DomainMapping`, `DomainToEntityIdMapping`, `ResolvedDomain`) and the validation functions moved to a new `src/schema/entity-domains.ts`; `data/entities/domains/entity-domains.ts` keeps only the singleton machinery (`updateDomainMapping`, `domainMappingForDomain`, `entitiesForDomain`). With the interfaces out of `data/**`, the `eslint-disable` directives are gone and the plugin is reverted to `^2.2.1` — `package.json`/`pnpm-lock.yaml` are byte-identical to `beta` again.

FYI for a possible follow-up outside this PR: 2.2.1 has a latent crash under ESLint 10 — on *any* sort-keys violation in `data/**` it dies with `TypeError: s.getSourceCode is not a function` instead of reporting (fix functions are evaluated at report time and 2.2.1 calls the removed `context.getSourceCode()`). Upstream fixed it in v3; 3.0.3 clears this repo's `minimumReleaseAge` on ~2026-07-19. Happy to send that bump as its own PR then.

src/tools/wallet-data-collection/wallet-data-collection-lib.ts
ren2140eth: Done in f86a883b — `readDomainMappingFile` is a pure read again; `writeDomainMappingFile` now calls `updateDomainMapping` after the atomic rename, so the module-level mapping is refreshed exactly when the file is overwritten. (The `mark-domain-update` parent-suffix hint doesn't need the read-time refresh: within one CLI invocation the import-time mapping and the file on disk are the same data, and any in-process write goes through `writeDomainMappingFile`.)

polymutex (approved):
👍

PR #883, opened 2026-07-03
Data: First-hand-verify Rainbow fee transparency (operationFees + bridge/swap fee units) -> PASS

ren2140eth:
First-hand verification of every input to Rainbow's `feeTransparency` rating, from the Rainbow iOS app and browser extension. Takes the attribute from **UNRATED** to **PASS**.

## What was filled / corrected

**`transparency.operationFees`**
- Filled the three `null` fields (`ethL1Transfer`, `erc20L1Transfer`, `uniswapUSDCToEtherSwap`): aggregate "Estimated fee" by default, comprehensive gas breakdown one tap away, no wallet platform fee (`NOT_APPLICABLE`).
- Corrected `builtInErc20Swap`: questionnaire had `byDefault: NONE` + `units: null`. First-hand it shows a network fee by default (`AGGREGATED`) and itemizes an "Included Rainbow Fee" on Review.

**Wallet-service fee units → `{TOKEN_AMOUNT, PERCENTAGE}`** (built-in swap + cross-chain bridge, both variants). The fee appears as an absolute amount in the line item and is **also disclosed as a percentage**:
- Built-in swap: **0.85%** — mobile: tap the fee line on Review; browser: fee info modal ("Rainbow takes a 0.85% fee from swaps").
- Bridge: **0.25%** — mobile: tap the fee line; browser: fee info modal ("Rainbow takes a 0.25% fee from bridges").

## Rating

The #864 evaluator rates on the *worst* fee type's `walletServiceFeeDisplayUnits`: `null` → UNRATED, absolute-only → PARTIAL, percentage/bps (or N/A) → PASS. With every wallet-service fee disclosing a percentage, no fee type is on an absolute-only unit, so **`feeTransparency` = PASS** (`comprehensive_fees`) on MOBILE, BROWSER, and overall — verified directly against the evaluator.

**Methodology note for review:** on **mobile**, the percentage is gated behind a *second* interaction (tap "Review", then tap the fee line); on the **browser** extension it is one info-icon click. This is recorded in the ref explanations and a code comment. The current evaluator counts a disclosed percentage regardless of click-depth; flagging in case that distinction should be modeled.

## Evidence
First-hand screenshots for every claim (L1 sends, Uniswap swap, gas-options breakdown, built-in swap default + Review, swap % on both variants, browser + mobile bridge Review, bridge % on both variants), each a labelled `file:` ref.

Rebased on latest `origin/beta` (post-#864).


ren2140eth:
**Fee transparency: should a relative fee unit gated behind an extra interaction still PASS?**

Filling Rainbow's fee transparency here surfaced a case worth checking against the #847 / #864 methodology.

Rainbow shows every wallet-service fee (built-in swap, cross-chain bridge) in **two** units: an absolute token amount in the line item, plus the rate as a **percentage** (swap 0.85%, bridge 0.25%). That's exactly the "both raw values and relative terms" case the `walletServiceFeeDisplayUnits` array was designed for, so I've recorded `{TOKEN_AMOUNT, PERCENTAGE}` — and the wallet currently evaluates to **PASS**.

The wrinkle: the percentage sits one interaction *deeper* than the comprehensive breakdown.
- **Browser extension:** breakdown appears on Review; the % is one click away via the fee-line info icon.
- **Mobile:** breakdown appears on Review; the % appears only after *tapping the fee line* — two interactions past the send screen.

The evaluator treats any relative unit as PASS regardless of interaction depth. That's consistent with the field's definition (units are a flat set, no depth dimension) — but worth noting the *display-level* axis **does** model depth (`afterSingleAction` = "at most one additional click"), while the unit axis does not.

**Open question:** is a discoverable-but-deeper % sufficient for PASS, or should there be a depth threshold (e.g. the % must appear within the same one-click breakdown)?
- *For PASS:* the #847 rationale is comparability — the effective rate **is** disclosed and computable; the concern was absolute-*only* denominations, which isn't Rainbow.
- *For a depth gate:* on mobile a two-tap-deep % is arguably not "shown when the comprehensive breakdown is available."

If depth should matter, the cleanest fix — keeping the feature reflecting truth, per the #864 review discussion — is to add a depth dimension to the feature (capture whether the relative unit is in the breakdown vs behind a further interaction) and gate the rating on it, rather than dropping the % from the unit set. Happy to implement whichever you prefer; screenshots for all four cases (swap/bridge x mobile/browser) are in this PR.

polymutex:
cc @johngrantuk Some interesting downstream questions on the fee units changes.

> Open question: is a discoverable-but-deeper % sufficient for PASS, or should there be a depth threshold (e.g. the % must appear within the same one-click breakdown)?

My take: Anything beyond one click should simply not be considered or recorded. If it's two clicks away, it may as well not exist.

Per the feature type documentation, the `walletServiceFeeDisplayUnits` applies only when `FeeDisplayLevel` is `COMPREHENSIVE`, so `{ byDefault: AGGREGATED, afterSingleAction: COMPREHENSIVE, walletServiceFeeDisplayUnits: PERCENTAGE }` means it is shown as a percentage only after 1 interaction. However maybe we should support this differently? @johngrantuk Thoughts?

As for recording the difference in interaction depth of the mobile version vs browser extension version, use the VariantFeature type at the `transparency.operationFees` level.

polymutex:
Filed #884 to refactor the feature types. Love how many edge cases you're encountering...

data/software-wallets/rainbow.ts
```diff
@@ -156,12 +159,31 @@ export const rainbow: SoftwareWallet = {
 								lastRetrieved: '2026-06-26',
 								url: 'https://rainbow.me/support/app/bridge-and-swap-tokens',
 							},
+							{
+								explanation:
+									'First-hand: the Rainbow browser extension "Review & Bridge" panel (reached by clicking "Review", one action) on a USDC cross-chain bridge. It itemizes the fee as minimum received, the bridging provider (Relay), and an "Included Rainbow fee" shown as a token amount (0.000586 USDC), alongside the estimated network fee. This confirms the breakdown is comprehensive after a single action, and that the Rainbow service fee is displayed as a token amount on the browser extension.',
```
polymutex: Not sure where the "First hand" is coming from here, but if I were reading Walletbeat and I see a reference that says "First hand" in it, that implies the other references aren't, so I would immediately start questioning the *rest* of the site as to whether or not it's truly testing wallets :)

polymutex: "This confirms the breakdown is comprehensive after a single action, and that the Rainbow service fee is displayed as a token amount on the browser extension." -> This is explanation text that is not needed; when displayed on the site that just looks repetitive with the explanation of the attribute.

data/software-wallets/rainbow.ts
```diff
@@ -156,12 +159,31 @@ export const rainbow: SoftwareWallet = {
...
+								file: 'public/references/wallets/rainbow/screenshots/2026-07-03-chain-abstraction-browser-bridge-review.png',
+								label:
+									'Rainbow browser extension bridge Review panel itemizing the "Included Rainbow fee" (0.000586 USDC) alongside the network fee',
+								lastRetrieved: '2026-07-03',
+							},
+							{
+								explanation:
+									'Clicking the info icon on the "Included Rainbow fee" line in the browser extension bridge Review surfaces a "Rainbow Fee" note stating "Rainbow takes a 0.25% fee from bridges." So on the browser extension the Rainbow bridge fee is also disclosed as a percentage, not only as a token amount.',
```
polymutex: " So on the browser extension the Rainbow bridge fee is also disclosed as a percentage, not only as a token amount." also unnecessary

data/software-wallets/rainbow.ts
```diff
@@ -225,18 +247,29 @@ export const rainbow: SoftwareWallet = {
 							},
 							{
 								explanation:
-									'Rainbow mobile app bridge "Review" screen (reached by tapping "Review", one action): fees are itemized as the destination network, minimum received, "Included Rainbow Fee", and max slippage, with no warning explaining bridge trust assumptions or cross-chain risk.',
+									'Rainbow mobile app bridge "Review" screen, reached by tapping "Review" (one action). Fees are itemized as separate line items: destination network, minimum received, the "Included Rainbow Fee", and max slippage. The "Included Rainbow Fee" is shown as a token amount (e.g. 0.00125 USDC), so the Rainbow service fee is a token amount. There is no warning explaining bridge trust assumptions or cross-chain risk.',
```
polymutex: "so the Rainbow service fee is a token amount"

same

data/software-wallets/rainbow.ts
```diff
@@ -225,18 +247,29 @@ export const rainbow: SoftwareWallet = {
...
 								file: 'public/references/wallets/rainbow/screenshots/2026-06-17-chain-abstraction-bridge-review-fees.jpg',
 								label:
 									'Rainbow mobile app screenshot of the bridge Review screen with itemized fees',
 								lastRetrieved: '2026-06-17',
 							},
+							{
+								explanation:
+									'On mobile, tapping the "Included Rainbow Fee" line on the bridge Review screen (a second tap, after Review) reveals the fee as a percentage: "Included Rainbow Fee 0.25%". So the bridge fee is disclosed as a percentage on mobile, gated behind one extra tap.',
```
polymutex: "So the bridge fee is disclosed as a percentage on mobile, gated behind one extra tap."

Same. Instead you can just change the first sentence to "On mobile, a tap on the "Included Rainbow Fee" line on the bridge Review screen is required  to reveal [...]"

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+				// NONE), and tapping "Review" (one action) reveals a comprehensive
+				// breakdown including an itemized "Included Rainbow Fee" line, shown as
+				// a token amount (ETH); tapping that fee line (mobile) or its info icon
+				// (browser) reveals the rate as a percentage (0.85%), so the unit set
+				// is {TOKEN_AMOUNT, PERCENTAGE}.
+				builtInErc20Swap: supported<WithRef<FeeDisplay>>({
+					ref: [
+						{
+							explanation:
+								'By default the Rainbow built-in swap screen shows the exchange rate and a single network-fee estimate (e.g. "$0.07"), but does not itemize the Rainbow service fee. So the fee shown by default is aggregated, not hidden.',
```
polymutex: Again consider how this looks on the site. "So the fee shown by default is aggregated, not hidden." seems superfluous, but also asks more questions than answers. Why is "hidden" even in consideration here? It makes sense when filling in the feature type, but when reading the site it seems out of place.

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+							file: 'public/references/wallets/rainbow/screenshots/2026-07-03-operation-fees-builtin-swap-default.png',
+							label:
+								'Rainbow iOS built-in swap screen showing a network-fee estimate by default, with the Rainbow service fee not itemized',
+							lastRetrieved: '2026-07-03',
+						},
+						{
+							explanation:
+								'Tapping "Review" (a single action) reveals a comprehensive breakdown: network, minimum received, an itemized "Included Rainbow Fee" (0.00000262 ETH, i.e. a token amount), max slippage, and the estimated network fee. So a single tap yields a comprehensive breakdown, and the wallet service fee is shown as a token amount.',
```
polymutex: "So a single tap yields a comprehensive breakdown, and the wallet service fee is shown as a token amount"

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+							file: 'public/references/wallets/rainbow/screenshots/2026-07-03-operation-fees-builtin-swap-review.png',
+							label:
+								'Rainbow iOS built-in swap Review panel itemizing the "Included Rainbow Fee" (in ETH) alongside network fee, minimum received, and max slippage',
+							lastRetrieved: '2026-07-03',
+						},
+						{
+							explanation:
+								'On mobile, tapping the "Included Rainbow Fee" line on the swap Review screen (a second tap, after Review) reveals the fee as a percentage: "Included Rainbow Fee 0.85%". So the swap fee is disclosed as a percentage, gated behind one extra tap.',
```
polymutex: "So the swap fee is disclosed as a percentage, gated behind one extra tap"

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+							file: 'public/references/wallets/rainbow/screenshots/2026-07-04-operation-fees-mobile-swap-percentage-fee.png',
+							label:
+								'Rainbow iOS swap Review showing the Included Rainbow Fee as 0.85% after tapping the fee line',
+							lastRetrieved: '2026-07-04',
+						},
+						{
+							explanation:
```
polymutex, line 894: "So the swap fee is disclosed as a percentage on the browser extension too"

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+				// For all three flows below, the transaction screen shows a single
+				// aggregate "Estimated fee" figure by default, and one tap on the gas
+				// control opens a sheet itemizing base fee / miner tip / max
+				// transaction fee. None of these flows carry a Rainbow platform fee,
+				// so walletServiceFeeDisplayUnits is NOT_APPLICABLE.
+				erc20L1Transfer: supported<WithRef<FeeDisplay>>({
+					ref: [
+						{
+							explanation:
+								'By default the USDC (ERC-20) L1 send screen shows only a single aggregate "Estimated fee" figure with no itemized breakdown. The fee shown by default is therefore aggregated.',
```
polymutex: "The fee shown by default is therefore aggregated"

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+							file: 'public/references/wallets/rainbow/screenshots/2026-07-03-operation-fees-erc20-l1-send-screen.png',
+							label:
+								'Rainbow iOS USDC (ERC-20) send screen showing a single aggregate "Estimated fee" figure by default, with no itemized breakdown',
+							lastRetrieved: '2026-07-03',
+						},
+						{
+							explanation:
+								'One tap on the gas control opens this sheet itemizing current base fee, max base fee, miner tip, and max transaction fee. No wallet platform fee applies to a plain ERC-20 transfer, so this network-fee breakdown is the complete breakdown (COMPREHENSIVE). The same sheet is reachable from every transaction type tested.',
```
polymutex: Uppercase "COMPREHENSIVE" is an internal concept. Again, on the site it's not clear to a reader what this means :)

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+							explanation:
+								'By default the ETH L1 send screen shows only a single aggregate "Estimated fee" figure with no itemized breakdown. The fee shown by default is therefore aggregated.',
+							file: 'public/references/wallets/rainbow/screenshots/2026-07-03-operation-fees-eth-l1-send-screen.png',
+							label:
+								'Rainbow iOS ETH send screen showing a single aggregate "Estimated fee" figure by default, with no itemized breakdown',
+							lastRetrieved: '2026-07-03',
+						},
+						{
+							explanation:
+								'One tap on the gas control opens this sheet itemizing current base fee, max base fee, miner tip, and max transaction fee. No wallet platform fee applies to a plain ETH transfer, so this network-fee breakdown is the complete breakdown (COMPREHENSIVE). The same sheet is reachable from every transaction type tested.',
```
polymutex: Same

data/software-wallets/rainbow.ts
```diff
@@ -823,17 +856,135 @@ export const rainbow: SoftwareWallet = {
...
+					afterSingleAction: FeeDisplayLevel.COMPREHENSIVE,
+					byDefault: FeeDisplayLevel.AGGREGATED,
+					fullySponsored: false,
+					walletServiceFeeDisplayUnits: 'NOT_APPLICABLE' as const,
+				}),
+				uniswapUSDCToEtherSwap: supported<WithRef<FeeDisplay>>({
+					ref: [
+						{
+							explanation:
+								'For a USDC-to-ETH swap initiated through the Uniswap frontend (app.uniswap.org), the Rainbow approval popup shows a single aggregate "Estimated fee" by default, alongside the simulated result. The Uniswap protocol fee is reflected in the simulated receive amount, not itemized separately. The fee shown by default is thus aggregated.',
```
polymutex: "The fee shown by default is thus aggregated"

polymutex (changes requested):
Left a bunch of comments below but they are generally all about the same-ish feedback: References need to make sense when seen on the site, not just when seen in the feature data files.

johngrantuk:
> cc @johngrantuk Some interesting downstream questions on the fee units changes.
> 
> > Open question: is a discoverable-but-deeper % sufficient for PASS, or should there be a depth threshold (e.g. the % must appear within the same one-click breakdown)?
> 
> My take: Anything beyond one click should simply not be considered or recorded. If it's two clicks away, it may as well not exist.
> 
> Per the feature type documentation, the `walletServiceFeeDisplayUnits` applies only when `FeeDisplayLevel` is `COMPREHENSIVE`, so `{ byDefault: AGGREGATED, afterSingleAction: COMPREHENSIVE, walletServiceFeeDisplayUnits: PERCENTAGE }` means it is shown as a percentage only after 1 interaction. However maybe we should support this differently? @johngrantuk Thoughts?
> 
> As for recording the difference in interaction depth of the mobile version vs browser extension version, use the VariantFeature type at the `transparency.operationFees` level.

I was assuming we would only consider/record up to a max of one action. In that case I believe the logic is correct but maybe this highlights it's not obvious how to record. Maybe it's worth having `walletServiceFeeDisplayUnits` for default and single action?

data/software-wallets/rainbow.ts
ren2140eth: Makes sense. Removed; the explanation now just describes the panel.

ren2140eth: Agreed. Dropped.

data/software-wallets/rainbow.ts
ren2140eth: Dropped

data/software-wallets/rainbow.ts
ren2140eth: Dropped

data/software-wallets/rainbow.ts
ren2140eth: Rewrote using your suggested form

data/software-wallets/rainbow.ts
ren2140eth: Removed

data/software-wallets/rainbow.ts
ren2140eth: Removed

data/software-wallets/rainbow.ts
ren2140eth: Dropped

data/software-wallets/rainbow.ts
ren2140eth: Dropped

data/software-wallets/rainbow.ts
ren2140eth: Dropped

polymutex (approved):

PR #886, opened 2026-07-04
Test: Create test to ensure all wallet ids are valid paths

maykelxyz:
Motivation: As @ren2140eth reported, path `base-app` returns 404. Walletbeat should have a test to ensure that all wallet ids ex. `base-app` are valid paths.


polymutex (changes requested):
Test appears to be failing

maykelxyz:
> Test appears to be failing 

Yeah, this is before pulling from `beta`. Let me merge, atleast we know now the test works, heh

polymutex (approved):
Looks good but you now probably need to sign your commit.

Issue #839, opened 2026-06-18
`mergeRefs` should deduplicate file-based references


polymutex:
On the per-wallet page, Ambire's Transaction Legibility attribute shows duplicated `ref`s right now. Need to dedup properly.

ren2140eth:
completed in #915

PR #915, opened 2026-07-10
Show image references in an in-page lightbox dialog

### LINKED ISSUE #840: `ReferenceLinks` should better support image-based references
Instead of linking to image files opening in a new tab/window, `ReferenceLinks` should show the screenshot in a modal, or perhaps have a carousel as part of the `ReferenceLinks` section.

ren2140eth:
Fixes #840

Image-based references in `ReferenceLinks` previously opened the raw image file in a new tab. This PR shows them in an in-page lightbox instead:

- Plain left-clicks on an image reference open a native `<dialog>` (`showModal()`), giving focus trapping, `Esc`-to-close, and a backdrop for free. Clicking the backdrop also closes it.
- The lightbox shows the screenshot with its reference label as the caption, previous/next buttons with a position counter (e.g. `2 / 12`), and `←`/`→` arrow-key navigation across all image references in the section.
- An "Open original" link inside the lightbox opens the raw image in a new tab, and modified clicks (middle, ctrl/cmd/shift) on the reference link itself keep default link behavior — so the raw image URL stays reachable and copyable for audits, per the discussion on the issue.
- Image detection is a new `isImageUrl()` helper in `src/schema/url.ts` based on path extension; it handles the root-relative URLs produced by file-based references under `public/`, which the `getDomain()`-based helpers reject.
- Non-image references are unchanged, and the lightbox content only mounts while open, so screenshots are not loaded until requested.

Verified locally on the Rainbow wallet page (12 image references in the fee transparency section): open, arrow-key navigation, Escape close, backdrop close, and "Open original" all work; non-image links behave as before.

### DIFF
--- a/src/schema/url.ts
+++ b/src/schema/url.ts
@@ -219,12 +219,19 @@ export const gitCommitRefPinRegExp = /@(?:[0-9a-f]{40}|[0-9a-f]{7})\b/g
 const imageFileExtensions = ['.png', '.jpg', '.jpeg', '.webp', '.gif', '.avif', '.svg']
 
 /**
- * Whether a URL points to an image file, based on its path extension.
- * Handles root-relative URLs (e.g. file-based references under public/),
- * which `getDomain`-based helpers would reject.
+ * Whether a URL points to an image file hosted in the Walletbeat repo
+ * itself, i.e. the root-relative URL that a file-based reference under
+ * `public/` qualifies to. External image URLs deliberately do not match:
+ * rendering them inline would leak visitor traffic to external hosts.
  */
-export function isImageUrl(url: Url): boolean {
+export function isRepoImageUrl(url: Url): boolean {
 	const path = (isLabeledUrl(url) ? url.url : url).split(/[?#]/)[0].toLowerCase()
+
+	// Root-relative, but not protocol-relative (`//host/...`), URLs are
+	// served from this site's own origin.
+	if (!path.startsWith('/') || path.startsWith('//')) {
+		return false
+	}
 
 	return imageFileExtensions.some(extension => path.endsWith(extension))
 }
--- a/src/schema/url.ts
+++ b/src/schema/url.ts
@@ -221,8 +221,8 @@ const imageFileExtensions = ['.png', '.jpg', '.jpeg', '.webp', '.gif', '.avif', 
 /**
  * Whether a URL points to an image file hosted in the Walletbeat repo
  * itself, i.e. the root-relative URL that a file-based reference under
- * `public/` qualifies to. External image URLs deliberately do not match:
- * rendering them inline would leak visitor traffic to external hosts.
+ * `public/` qualifies to. External (absolute or protocol-relative) image
+ * URLs do not match.
  */
 export function isRepoImageUrl(url: Url): boolean {
 	const path = (isLabeledUrl(url) ? url.url : url).split(/[?#]/)[0].toLowerCase()
--- a/src/schema/url.ts
+++ b/src/schema/url.ts
@@ -215,6 +215,20 @@ function getDefaultUrlLabel(url: string): string {
  */
 export const gitCommitRefPinRegExp = /@(?:[0-9a-f]{40}|[0-9a-f]{7})\b/g
 
+/** File extensions of URLs that browsers render as images. */
+const imageFileExtensions = ['.png', '.jpg', '.jpeg', '.webp', '.gif', '.avif', '.svg']
+
+/**
+ * Whether a URL points to an image file, based on its path extension.
+ * Handles root-relative URLs (e.g. file-based references under public/),
+ * which `getDomain`-based helpers would reject.
+ */
+export function isImageUrl(url: Url): boolean {
+	const path = (isLabeledUrl(url) ? url.url : url).split(/[?#]/)[0].toLowerCase()
+
+	return imageFileExtensions.some(extension => path.endsWith(extension))
+}
+
 /** Return the label for a URL. */
 export function getUrlLabel(url: Url): string {
 	if (isLabeledUrl(url)) {
--- a/src/views/ReferenceLinks.svelte
+++ b/src/views/ReferenceLinks.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	// Types/constants
 	import type { FullyQualifiedReference } from '@/schema/reference'
+	import { isImageUrl } from '@/schema/url'
 
 
 	// Props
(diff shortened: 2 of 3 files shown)


ren2140eth:
Petals are missing when looking at the site via dev server but this is unrelated to this PR
Screenshot From 2026-07-10 15-17-51

polymutex (changes requested):
Looks good, but:

- Can you make sure this follows the CSS attribute system (see `src/styles/css-attributes.css` and other files in that directory)
- For the case of a ReferenceLinks section with a single image reference, have the image be displayed inline (with a max width/height, preserving ratio, similar to the layout the lightbox has but inline)
- For multi-image ReferenceLinks, have the image be displayed next to each reference (maybe floating on the left or the right, whatever you think looks nicer); make sure that each picture takes the same horizontal and vertical space as one another if laid out this way.

ren2140eth:
Thanks for the review — all three addressed in 6e4783a2:

- **CSS attribute system**: the lightbox now uses `data-card`, `data-column`/`data-row`, and `data-icon="circle"` primitives. Custom CSS remains only where no primitive covers it: the dialog's fixed dimensions, `::backdrop`, and image `object-fit`/sizing.
- **Single-image sections**: the image renders inline at a bounded size (max-height 20em, ratio preserved), with the labeled source link as its caption — same layout as the lightbox, which still opens on click for a larger view.
- **Multi-image sections**: each reference gets its thumbnail on the right, all in identical fixed-size boxes (`object-fit: cover`), clicking through to the lightbox at that image.

One fix that came out of testing this: the same image file can back multiple references within a section (Ambire's transaction legibility does this today), and the lightbox previously counted duplicates — stepping through identical images and inflating its counter. It also would have prevented single-image inline display from ever triggering in that case. The section's image list is now deduplicated by URL.

polymutex (changes requested):
Looks good, but two more requests:

- The image deduplication thing is good but is preferably done at an earlier stage in `mergeRefs`. Look at https://github.com/walletbeat/walletbeat/issues/839 about this. Perhaps this PR can solve both problems at once. Merging earlier allows the UI code to not have to deduplicate. Currently the Ambire Transaction Legibility attribute is a good test case for this deduplication; right now the lightbox view correctly dedupes but the reference list in the page doesn't, which is a weird mismatch.
- Please refactor the lightbox stuff in its own component file; this ReferenceLinks.svelte thing is getting heavy.

cc @darrylyeo Do you want to review this further?

ren2140eth:
Both done:

- Dedup moved into mergerefs, also fixes 839 so refs with no explanation and identical URL sets are dedduped. Same URL refs wtih different explanations are left since they back distinct claims (see rainbow - fee transparency). I diffed every evaluated reference site-wide before/after: only ambire trx legibility and a doubled keycard duress-resistance link changed. One implementation note: the dedup branch deliberately avoids mergeLabeledUrls, which crashes on the root-relative URLs that file-based references qualify to (same getDomain limitation as isImageUrl)
- Lightbox extracted to src/components/ImageLightbox.svelte

src/schema/reference.ts
polymutex, line 1: Fixing the issues in `reference.ts` in a different way in #927. Please revert these changes and rebase, it should dedupe them.

src/views/ReferenceLinks.svelte
```diff
@@ -13,9 +14,47 @@
...
 
+	// Internal state
+	let lightbox = $state<{ open: (url: string) => void }>()
+
+	// (Derived)
+
+	// References arrive deduplicated by `mergeRefs`, so every image URL
+	// appears at most once here.
+	const imageUrls = $derived(
+		references.flatMap(ref => ref.urls).filter(url => isImageUrl(url.url)),
```
polymutex: We should only display image URLs for images that are in the repo (i.e. file-based references), otherwise we leak Walletbeat traffic info to whatever other server the image is hosted on. Please replace the use of `isImageUrl` with something more specific.

src/views/ReferenceLinks.svelte
```diff
@@ -36,51 +75,138 @@
 
 		<ul class="references-list" data-list="gap-2">
 			{#each references as ref, index (index + '::' + ref.urls.map(url => url.url).toSorted().join('|'))}
+				{@const refImages = ref.urls.filter(url => isImageUrl(url.url))}
```
polymutex: Same

tests/merge-refs.test.ts
polymutex, line 1: Please revert this as well given the revert to `reference.ts`

polymutex (changes requested):

src/schema/reference.ts
ren2140eth: Done - rebased onto beta with our `mergeRefs` commit dropped. Verified on the Ambire Transaction Legibility attribute: the refs arrive deduped.

src/views/ReferenceLinks.svelte
ren2140eth: Replaced with `isRepoImageUrl`, which only matches the root-relative URLs that file-based references under `public/` qualify to (it also rejects protocol-relative `//host/...` URLs). External image URLs now render as regular external links, so wallet pages never load images from external hosts.

src/views/ReferenceLinks.svelte
ren2140eth: Done — same `isRepoImageUrl` check here.

tests/merge-refs.test.ts
ren2140eth: Done

src/schema/url.ts
```diff
@@ -215,6 +215,27 @@ function getDefaultUrlLabel(url: string): string {
...
 export const gitCommitRefPinRegExp = /@(?:[0-9a-f]{40}|[0-9a-f]{7})\b/g
 
+/** File extensions of URLs that browsers render as images. */
+const imageFileExtensions = ['.png', '.jpg', '.jpeg', '.webp', '.gif', '.avif', '.svg']
+
+/**
+ * Whether a URL points to an image file hosted in the Walletbeat repo
+ * itself, i.e. the root-relative URL that a file-based reference under
+ * `public/` qualifies to. External image URLs deliberately do not match:
+ * rendering them inline would leak visitor traffic to external hosts.
```
polymutex: This last sentence is a detail that doesn't make sense in the pure context of this file (which is just code to handle URLs in general), it only makes sense in the context of its current usage in ImageLightbox.svelte, where the comment should be moved to.

polymutex (approved):
@darrylyeo Please review, otherwise please merge this after addressing the comment below and waiting 24h @ren2140eth

src/schema/url.ts
ren2140eth: moved the rationale to where isRepoImageUrl is actually applied (ReferenceLinks.svelte) vs. ImageLightbox.svelte

PR #878, opened 2026-07-02
feat: eip directory ui improvement

pheobeayo:
## Add attribute icons and wallet implementation status to `/wallet-eips`

Closes #872 

### What this adds

Two enhancements to the `/wallet-eips` page as requested in #872:

1. **Attribute icons per EIP** — each EIP row now shows the wbicons font icon matching its thematic attribute (e.g. 🧾 for signing EIPs, 🌐 for provider EIPs, 👤 for account abstraction EIPs)
2. **Wallet implementation status in the expanded view** — when a user expands an EIP, they see which wallets support it, partially support it, or don't support it yet, with wallet icons linking to the per-wallet page

### Files changed

**Schema**
- `src/schema/eips.ts` — adds two fields to the `Eip` interface:
  - `icon: WBIconID` — the wbicons icon representing the EIP's theme
  - `relatedAttributeIds: string[]` — walletbeat attribute IDs used to derive wallet implementation status

**EIP data files** (all 12 in `data/eips/`) — each populated with `icon` and `relatedAttributeIds`:

| EIPs | icon | relatedAttributeIds |
|---|---|---|
| 712, 7730, 8213 | `transaction_legibility` | `['transactionLegibility']` |
| 1193, 2700, 6963 | `browser_integration` | `['browserIntegration']` |
| 4337, 7702 | `account_abstraction` | `['accountAbstraction']` |
| 5564 | `private_token_transfers` | `['privateTransfers']` |
| 5792 | `transaction_batching` | `['transactionBatching']` |
| 7828, 7831 | `address_resolution` | `['addressResolution']` |

**Component**
- `src/components/WalletEipList.svelte` — updated to:
  - Render the EIP icon using `data-icon="wbicons {eip.icon}"` in each row
  - Derive wallet ratings from `ratedSoftwareWallets` via `wallet.overall[groupId][attrId].evaluation.outcome.rating`
  - Display wallets grouped as Supports / Partial / Not supported with icons linking to per-wallet pages; failing wallets shown greyscale

### Implementation notes

- Wallet implementation status is derived at render time from the existing `ratedSoftwareWallets` evaluation tree — no new data files needed
- The `relatedAttributeIds` array allows one EIP to map to multiple attributes if needed in future
- Per-EIP wallet tracking pages (`/wallet-eips/7730` etc.) are out of scope for this PR, as noted in the issue

Issue #872, opened 2026-07-01
EIP directory UI improvements


polymutex:
Related but not exactly the same as #871.

For the EIP directory at `/wallet-eips`, we should:

- Have the attribute icons related to it (we may need to add a `WBIconID` field to the EIP interface definition)
- List the wallet icons that implement it in the expanded view, linking to the per-wallet page about it (same implementation derivation logic as needed by #871)

polymutex:
cc @pheobeayo if you'd like to work on further improvements to this page :)

pheobeayo:
I would like to work on it @polymutex

Issue #871, opened 2026-07-01
Per-EIP wallet implementation tracking


maykelxyz:
Continuation of #759 

Like 7702 tracker, Walletbeat should have a per-EIP wallet implementation tracker

<img width="1702" height="884" alt="Image" src="https://github.com/user-attachments/assets/87a3fee2-9de0-4d6d-bbae-e7ea3f53a674" />

Will mark as enhancement as I think this is a "nice to have" and a less prio implementation

PR #919, opened 2026-07-11
Add walletEipSupport() to derive per-EIP wallet support from resolved features

ren2140eth:
## Summary

Adds `walletEipSupport()` in `src/schema/eip-support.ts`: a derivation function that takes a wallet's `ResolvedFeatures` and returns which tracked EIPs the wallet implements, as `Record<EipNumber, WithRef<Support> | null>`.

This is the common building block discussed with @polymutex for #871 (per-EIP wallet support tracker) and #872 / #878 (EIP directory UI improvements): both UI surfaces can consume the same derived record, so they can never disagree about a wallet's EIP support. No UI changes in this PR.

## Semantics

- `SUPPORTED` = the wallet implements the EIP, at least partially.
- `NOT_SUPPORTED` = verified not to implement it.
- `null` = unknown, or not applicable to this wallet type (e.g. browser integration EIPs for wallets with no browser variant). The return type cannot distinguish these two cases; if a UI needs to render "N/A" differently from "unknown", that distinction can be layered on later without breaking this API.
- References are normalized onto each entry from wherever they live in the feature tree: on the `Supported` payload itself (account types, wallet call), on the enclosing record (browser integration, address resolution, transaction legibility), with `refNotNecessary` as the fallback.

Per-EIP derivation:

| EIP | Derived from |
| --- | --- |
| 712 | `security.transactionLegibility` → ERC-8213 message signing legibility: supported iff any EIP-712 signing detail (decoded struct, domain/message hash, digest) is surfaced |
| 1193 / 2700 / 6963 | `integration.browser` record |
| 4337 | `accountSupport.rawErc4337` |
| 5564 | `privacy.transactionPrivacy` stealth address support |
| 5792 | `walletCall` |
| 7702 | `accountSupport.eip7702` |
| 7730 | `security.transactionLegibility.erc7730`: supported iff at least one complex benchmark transaction is decoded (hardware: on-device or via companion app) |
| 7828 / 7831 | `addressResolution.chainSpecificAddressing` |
| 8213 | `security.transactionLegibility.erc8213`: supported iff the calldata digest is shown, or the EIP-712 signing display requirement is met (EIP-712 digest, or domain hash + message hash together, per the alternative the ERC explicitly accepts) |

The 712/7730/8213 logic follows the shape of the `walletSupport` derivations in #878, collapsed to binary (PASS/PARTIAL → `SUPPORTED`, FAIL → `NOT_SUPPORTED`), with two deliberate deviations: (1) unassessed message-signing/calldata data (`null` fields inside a `supported(...)` payload) maps to `null` (unknown) rather than `NOT_SUPPORTED`, per the null-over-false convention; (2) for ERC-8213, showing the domain hash and message hash together is credited as equivalent to showing the EIP-712 digest, matching the alternative the ERC summary in `data/eips/erc-8213.ts` explicitly accepts (#878's 8213 derivation only credits the digest, though its 712 derivation does honor this alternative). Graded implementation-quality assessments stay in attributes; this function only answers "does the wallet implement this EIP at all?".

Hardware wallets are handled wherever the data supports it (7702/4337 account types, hardware-shaped 7730/8213 records); note that `walletCall` resolves to `notSupported` for non-software wallets upstream of this function, so hardware wallets show `NOT_SUPPORTED` rather than N/A for 5792.

## Tests

`tests/eip-support.test.ts`:

- Unknown-features fixture → all 12 entries `null` (null never becomes ❌).
- Targeted unit tests for browser integration ref fan-out, account type ref preservation, wallet call, and software + hardware transaction legibility derivations.
- Smoke test across every software and hardware wallet × variant: the derivation never throws, always returns exactly the 12 tracked EIPs, and every non-null entry is a well-formed ref-annotated `Support` value.

`pnpm check:quick` and `pnpm check:vitest` pass locally.

### DIFF
--- a/src/schema/eip-support.ts
+++ b/src/schema/eip-support.ts
@@ -221,8 +221,9 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 		}),
 	'7828': features => addressResolutionEipSupport(features, 'erc7828'),
 	'7831': features => addressResolutionEipSupport(features, 'erc7831'),
-	// A wallet implements ERC-8213 if it exposes the calldata digest or the
-	// EIP-712 digest when the user is asked to sign.
+	// A wallet implements ERC-8213 if it displays the calldata digest, or
+	// satisfies the ERC's EIP-712 signing display requirement: the EIP-712
+	// digest, or the domain hash and message hash together.
 	'8213': features =>
 		transactionLegibilityEipSupport(features, transactionLegibility => {
 			const erc8213 = transactionLegibility.erc8213
@@ -241,12 +242,16 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 				return null
 			}
 
-			return (
-				(calldataDisplay !== null &&
-					displayEntryIsShown(calldataDisplay[CallDataDisplay.CALLDATA_DIGEST])) ||
-				(messageSigningLegibility !== null &&
-					displayEntryIsShown(messageSigningLegibility[MessageSigningDetails.EIP712_DIGEST]))
-			)
+			const calldataDigestShown =
+				calldataDisplay !== null &&
+				displayEntryIsShown(calldataDisplay[CallDataDisplay.CALLDATA_DIGEST])
+			const signatureDigestShown =
+				messageSigningLegibility !== null &&
+				(displayEntryIsShown(messageSigningLegibility[MessageSigningDetails.EIP712_DIGEST]) ||
+					(displayEntryIsShown(messageSigningLegibility[MessageSigningDetails.DOMAIN_HASH]) &&
+						displayEntryIsShown(messageSigningLegibility[MessageSigningDetails.MESSAGE_HASH])))
+
+			return calldataDigestShown || signatureDigestShown
 		}),
 }
 
--- a/src/schema/eip-support.ts
+++ b/src/schema/eip-support.ts
@@ -22,8 +22,8 @@ import { hasRefs, mergeRefs, refTodo, type WithRef } from './reference'
 /**
  * Whether a wallet implements a specific EIP.
  *
- * - `SUPPORTED` means the wallet implements the EIP, at least partially.
- * - `NOT_SUPPORTED` means the wallet was verified not to implement the EIP.
+ * - `Supported` means the wallet implements the EIP, at least partially.
+ * - `NotSupported` means the wallet was verified not to implement the EIP.
  * - `'UNKNOWN'` means the implementation status has not been assessed yet.
  * - `'NOT_APPLICABLE'` means the EIP does not apply to this type of wallet
  *   (e.g. browser integration EIPs for wallets that have no browser variant).
@@ -50,11 +50,6 @@ function eipSupport(
 	}
 }
 
-/** The references attached to `value` itself, if any. */
-function ownRefs(value: unknown): WithRef<unknown>['ref'] | undefined {
-	return hasRefs(value) ? value.ref : undefined
-}
-
 /** EIP support based on the browser integration record. */
 function browserIntegrationEipSupport(
 	features: ResolvedFeatures,
@@ -73,7 +68,7 @@ function browserIntegrationEipSupport(
 	}
 
 	// The browser integration record's references document all browser EIPs.
-	return eipSupport(isSupported(support), ownRefs(support), browser.ref)
+	return eipSupport(isSupported(support), hasRefs(support) ? support.ref : undefined, browser.ref)
 }
 
 /** EIP support based on chain-specific address resolution. */
@@ -94,7 +89,11 @@ function addressResolutionEipSupport(
 	}
 
 	// The address resolution record's references document both address ERCs.
-	return eipSupport(isSupported(support), ownRefs(support), addressResolution.ref)
+	return eipSupport(
+		isSupported(support),
+		hasRefs(support) ? support.ref : undefined,
+		addressResolution.ref,
+	)
 }
 
 /**
@@ -166,8 +165,8 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 		if (isSupported(rawErc4337) || delegateIsErc4337 === true) {
 			return eipSupport(
 				true,
-				isSupported(rawErc4337) ? ownRefs(rawErc4337) : undefined,
-				delegateIsErc4337 === true ? ownRefs(eip7702) : undefined,
+				isSupported(rawErc4337) ? rawErc4337.ref : undefined,
+				delegateIsErc4337 === true && isSupported(eip7702) ? eip7702.ref : undefined,
 			)
 		}
 
@@ -186,7 +185,10 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 
 		const stealthAddresses = transactionPrivacy[PrivateTransferTechnology.STEALTH_ADDRESSES]
 
-		return eipSupport(isSupported(stealthAddresses), ownRefs(stealthAddresses))
+		return eipSupport(
+			isSupported(stealthAddresses),
+			hasRefs(stealthAddresses) ? stealthAddresses.ref : undefined,
+		)
 	},
 	'5792': features => {
 		const walletCall = features.walletCall
@@ -195,7 +197,7 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 			return 'UNKNOWN'
 		}
 
-		return eipSupport(isSupported(walletCall), ownRefs(walletCall))
+		return eipSupport(isSupported(walletCall), hasRefs(walletCall) ? walletCall.ref : undefined)
 	},
 	'6963': features => browserIntegrationEipSupport(features, '6963'),
 	'7702': features => {
@@ -207,7 +209,7 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 
 		const eip7702 = accountSupport.eip7702
 
-		return eipSupport(isSupported(eip7702), ownRefs(eip7702))
+		return eipSupport(isSupported(eip7702), isSupported(eip7702) ? eip7702.ref : undefined)
 	},
 	// ERC-7730 is registry-based: a wallet that implements it decodes every
 	// transaction in the registry, so it must decode *all* of the complex
--- a/src/schema/eip-support.ts
+++ b/src/schema/eip-support.ts
@@ -160,9 +160,7 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 
 		if (isSupported(eip7702)) {
 			delegateIsErc4337 =
-				eip7702.contract === 'UNKNOWN'
-					? null
-					: isSupported(eip7702.contract.methods.validateUserOp)
+				eip7702.contract === 'UNKNOWN' ? null : isSupported(eip7702.contract.methods.validateUserOp)
 		}
 
 		if (isSupported(rawErc4337) || delegateIsErc4337 === true) {
@@ -232,21 +230,19 @@ const eipSupportResolvers: Record<EipNumber, (features: ResolvedFeatures) => Eip
 			return eipSupport(false, transactionLegibility.ref)
 		}
 
-		const decoded = Object.values(ComplexBenchmarkTransactions).map(
-			(benchmark): boolean | null => {
-				const entry = erc7730[benchmark]
+		const decoded = Object.values(ComplexBenchmarkTransactions).map((benchmark): boolean | null => {
+			const entry = erc7730[benchmark]
 
-				if (entry === null) {
-					return null
-				}
+			if (entry === null) {
+				return null
+			}
 
-				if (typeof entry === 'string') {
-					return entry === DataLocation.ON_DEVICE || entry === DataLocation.OFF_DEVICE
-				}
+			if (typeof entry === 'string') {
+				return entry === DataLocation.ON_DEVICE || entry === DataLocation.OFF_DEVICE
+			}
 
-				return isShown(entry.decoded)
-			},
-		)
+			return isShown(entry.decoded)
+		})
 
 		if (decoded.includes(false)) {
 			return eipSupport(false, transactionLegibility.ref)
--- a/src/schema/eip-support.ts
+++ b/src/schema/eip-support.ts
@@ -1,0 +1,263 @@
+import { remap } from '@/types/utils/remap'
+
+import type { EipNumber } from './eips'
+import type { ResolvedFeatures } from './features'
+import { AccountType } from './features/account-support'
+import type { BrowserIntegrationEip } from './features/ecosystem/integration'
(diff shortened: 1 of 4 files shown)


src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+import { hasRefs, refNotNecessary, type WithRef } from './reference'
+
+/**
+ * Whether a wallet implements a specific EIP.
+ *
+ * - `SUPPORTED` means the wallet implements the EIP, at least partially.
+ * - `NOT_SUPPORTED` means the wallet was verified not to implement the EIP.
+ * - `null` means the implementation status is unknown, or that the EIP does
+ *   not apply to this type of wallet (e.g. browser integration EIPs for
+ *   wallets that have no browser variant).
```
polymutex: Good point; this means there are two cases for `null` right now, which we need to be able to distinguish.

So let's replace this with `export type EipSupport = WithRef<Support> | 'UNKNOWN' | 'NOT_APPLICABLE'`.

src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+export type WalletEipSupport = Record<EipNumber, EipSupport>
+
+/** A `Support` value annotated with references. */
+function supportWithRef(
+	implemented: boolean,
+	ref: WithRef<unknown>['ref'] | undefined,
+): WithRef<Support> {
+	return {
+		...(implemented ? featureSupported : notSupported),
+		ref: ref ?? refNotNecessary,
```
polymutex: Should be `refTodo` instead of `refNotNecessary` here.

src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/**
+ * Normalize a feature-level `Support` value into an `EipSupport`, preserving
+ * references attached to the value itself and falling back to `fallbackRef`
+ * (typically the references of the enclosing feature) otherwise.
+ */
+function normalizeSupport(
+	support: Support<object> | null,
+	fallbackRef?: WithRef<unknown>['ref'],
```
polymutex: Too easy to run into unintended ref attachment here. All `ref`s should be hand-selected on a per-EIP case-by-case basis depending on what makes sense. If multiple do, they should all be taken into consideration (using `mergeRefs`), though in practice I don't think any EIP needs multiple refs just yet.

src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+ * `derive` returns whether the EIP is implemented, or `null` if the
+ * transaction legibility data does not answer that question.
+ */
+function transactionLegibilityEipSupport(
+	features: ResolvedFeatures,
+	derive: (
+		transactionLegibility:
+			| HardwareTransactionLegibilityImplementation
+			| SoftwareTransactionLegibilityImplementation,
+	) => boolean | null,
```
polymutex: Should return `EipSupport` now that it needs to distinguish the unknown vs not applicable case and to carry references properly, and at that point it may mean that having `transactionLegibilityEipSupport` as a middleman function no longer makes sense.

src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/**
+ * Whether a display entry is shown to the user.
+ * Handles both the software wallet shape (a bare `DataDisplayOptions`) and
+ * the hardware wallet shape (a `DisplayCapability` object).
+ */
+function displayEntryIsShown(entry: DataDisplayOptions | { display: DataDisplayOptions }): boolean {
+	return isShown(typeof entry === 'object' ? entry.display : entry)
+}
```
polymutex: Belongs to `src/schema/features/security/transaction-legibility.ts`

src/schema/eip-support.ts
```diff
@@ -0,0 +1,268 @@
...
+
+			if (erc7730 === null) {
+				return null
+			}
+
+			if (!isSupported<HardwareWalletErc7730 | SoftwareWalletErc7730>(erc7730)) {
+				return false
+			}
+
+			return Object.values(ComplexBenchmarkTransactions).some(benchmark => {
```
polymutex: 7730 is registry-based, so a wallet that actually supports it would support *all* the transactions in the registry, not just any.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+	notSupported,
+	supported,
+} from '@/schema/features/support'
+import { refs } from '@/schema/reference'
+import { Variant } from '@/schema/variants'
+import type { BaseWallet } from '@/schema/wallet'
+import { WalletType } from '@/schema/wallet-types'
+
+/** A resolved feature set where every feature is unknown. */
+function unknownResolvedFeatures(): ResolvedFeatures {
```
polymutex: This should just use `data/software-wallets/unrated.tmpl.ts`. No need for this function.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+				eip7702: supported({
+					ref: { url: 'https://example.com/7702' },
+					contract: 'UNKNOWN',
+				}),
+				safe: notSupported,
+			},
+		})
+
+		expectSupported(eipSupport['7702'])
+		expectNotSupported(eipSupport['4337'])
```
polymutex: A 7702-supporting wallet can probably be said to support 4337 too, in the sense that it supports a smart-account-based account type.
The 4337 support attribution code should use 4337 and 7702 account support data to determine the support status (supported if either is, and use the ref from whichever branch is supported, or both if both are).
Probably deserves its own test case.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+			...unknownResolvedFeatures(),
+			walletCall: supported({
+				ref: { url: 'https://example.com/5792' },
+				atomicMultiTransactions: featureSupported,
+			}),
+		})
+
+		expectSupported(eipSupport['5792'])
+		expect(refUrls(eipSupport['5792'])).toContain('https://example.com/5792')
+	})
```
polymutex, line 219: Should have another test case with atomicMultiTransactions not supported.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+				},
+			},
+		})
+
+		// Per ERC-8213, displaying the domain hash and message hash together is
+		// an accepted alternative to displaying the EIP-712 digest.
+		expectSupported(eipSupport['8213'])
+	})
+
+	it('derives ERC-7730 from hardware transaction legibility', () => {
```
polymutex: Should have separate test cases for the all-transactions-shown vs partial-only.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+								continue
+							}
+
+							const eipSupport = walletEipSupport(
+								resolveFeatures(wallet.features, wallet.variants, variant),
+							)
+
+							// One entry per tracked EIP, each either unknown or a
+							// reference-annotated Support value.
+							expect(Object.keys(eipSupport).sort()).toEqual(Object.keys(eips).sort())
```
polymutex: Useless, already guaranteed by typesystem.

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+		})
+
+		// The wallet decodes at least one benchmark transaction on-device.
+		expectSupported(eipSupport['7730'])
+		// The ERC-8213 data was not assessed.
+		expect(eipSupport['8213']).toBeNull()
+		expect(eipSupport['712']).toBeNull()
+	})
+
+	describe('handles all wallets', () => {
```
polymutex: This is good to have but should move to a subtest in `tests/wallet-integrity.test.ts`

tests/eip-support.test.ts
```diff
@@ -0,0 +1,342 @@
...
+							)
+
+							// One entry per tracked EIP, each either unknown or a
+							// reference-annotated Support value.
+							expect(Object.keys(eipSupport).sort()).toEqual(Object.keys(eips).sort())
+
+							for (const support of Object.values(eipSupport)) {
+								if (support !== null) {
+									expect(isMaybeSupported(support)).toBe(true)
+									expect(Object.hasOwn(support, 'ref')).toBe(true)
```
polymutex: Already guaranteed by typesystem

polymutex (changes requested):

polymutex:
(Please run `pnpm check:all` instead of just the quick tests when verifying changes.)

src/schema/eip-support.ts
ren2140eth: Done in 51fdd191: `EipSupport` is now `WithRef<Support> | 'UNKNOWN' | 'NOT_APPLICABLE'`. `'NOT_APPLICABLE'` is currently only produced for browser integration EIPs on wallets with no browser variant; everything else unassessed resolves to `'UNKNOWN'`.

src/schema/eip-support.ts
ren2140eth: Done: falls back to `refTodo` when no reference source applies.

src/schema/eip-support.ts
ren2140eth: Done: removed the generic fallback. Each resolver now hand-picks its reference sources and merges them with `mergeRefs()` (e.g. browser EIPs merge the support value's own refs with the browser integration record's refs; account-type EIPs use only the account type support's refs).

src/schema/eip-support.ts
ren2140eth: Went with dropping the middleman entirely: the 712/7730/8213 resolvers are now inlined and return `EipSupport` directly, distinguishing unknown from not-supported at each step and attaching `transactionLegibility.ref` explicitly.

src/schema/eip-support.ts
ren2140eth: Done: moved to `transaction-legibility.ts` (next to `isShown`), which also lets it use the module-private `DisplayCapability` type instead of an inline shape.

src/schema/eip-support.ts
ren2140eth: Done: all complex benchmark transactions must now decode. Any benchmark verified as not decoding yields NOT_SUPPORTED; if the assessed ones decode but some are unassessed (`null`), the result is `'UNKNOWN'` since full registry coverage can't be confirmed either way. Added test cases for all three outcomes.

tests/eip-support.test.ts
ren2140eth: Done: tests now resolve `data/software-wallets/unrated.tmpl.ts` (and `data/hardware-wallets/unrated.tmpl.ts` for the hardware cases) via `resolveFeatures()`.

tests/eip-support.test.ts
ren2140eth: Done: the 4337 resolver now reports supported if either raw ERC-4337 or EIP-7702 account support is present, merging the refs from whichever branch(es) are supported. Added a dedicated test case asserting a 7702-only wallet gets ERC-4337 credit with the 7702 refs.

tests/eip-support.test.ts
ren2140eth: Added. I kept the expectation as supported-for-5792 in that case, since the wallet still implements the wallet call API and `atomic` is a capability within it

tests/eip-support.test.ts
ren2140eth: Done: split into an all-benchmarks-decode case (supported), a partial-only case (not supported), and an assessed-decode-but-some-unassessed case ('UNKNOWN').

tests/eip-support.test.ts
ren2140eth: Removed.

tests/eip-support.test.ts
ren2140eth: Done: moved into the per-wallet block in `tests/wallet-integrity.test.ts` as a 'derives EIP support for each variant' subtest.

tests/eip-support.test.ts
ren2140eth: Removed.

ren2140eth:
All review comments addressed in 51fdd191. Verified with `pnpm check:all` this time — all checks pass.

tests/eip-support.test.ts
ren2140eth: Refined this in b8b6dd38 after a second look: EIP-7702 alone doesn't guarantee ERC-4337 — the delegate contract may not be an ERC-4337 account at all. Since `SmartWalletContract` already records `methods.validateUserOp`, the resolver now credits ERC-4337 from the 7702 branch only when the delegate contract is known to implement `validateUserOp`; an unknown delegate contract yields `'UNKNOWN'`, and a known non-4337 delegate (with no raw 4337 support) yields NOT_SUPPORTED. Test cases cover all three. Happy to go back to the broader smart-account interpretation if you prefer it for this row.

src/schema/eip-support.ts
```diff
@@ -0,0 +1,308 @@
...
+	return {
+		...(implemented ? featureSupported : notSupported),
+		ref: mergedRefs.length > 0 ? mergedRefs : refTodo,
+	}
+}
+
+/** The references attached to `value` itself, if any. */
+function ownRefs(value: unknown): WithRef<unknown>['ref'] | undefined {
+	return hasRefs(value) ? value.ref : undefined
+}
```
polymutex: Just use `hasRefs` directly

src/schema/eip-support.ts
```diff
@@ -0,0 +1,308 @@
...
+	type SoftwareWalletErc8213,
+} from './features/security/transaction-legibility'
+import { featureSupported, isSupported, notSupported, type Support } from './features/support'
+import { hasRefs, mergeRefs, refTodo, type WithRef } from './reference'
+
+/**
+ * Whether a wallet implements a specific EIP.
+ *
+ * - `SUPPORTED` means the wallet implements the EIP, at least partially.
+ * - `NOT_SUPPORTED` means the wallet was verified not to implement the EIP.
```
polymutex: These should be `Supported` and `NotSupported`, not enum-ish values like the other below.

tests/wallet-integrity.test.ts
```diff
@@ -105,6 +109,21 @@ describe('wallets', () => {
...
+					if (wallet.variants[variant] !== true) {
+						continue
+					}
+
+					const eipSupport = walletEipSupport(
+						resolveFeatures(wallet.features, wallet.variants, variant),
+					)
+
+					// One entry per tracked EIP.
+					expect(Object.keys(eipSupport).sort()).toEqual(Object.keys(eips).sort())
```
polymutex: Again this is already guaranteed by the typesystem. The only thing this test needs to do is to verify that the `walletEipSupport` function runs without throwing.

polymutex (approved):
Good to merge after the below. Notify #878 once merged

src/schema/eip-support.ts
ren2140eth: Done in 8e1a68f8.

src/schema/eip-support.ts
ren2140eth: Done.

tests/wallet-integrity.test.ts
ren2140eth: Done: the subtest now only asserts `walletEipSupport()` does not throw.

PR #921, opened 2026-07-12
Add per-EIP wallet adoption tracker pages

### LINKED ISSUE #871: Per-EIP wallet implementation tracking

ren2140eth:
## Summary

Builds the UI half of #871 on top of `walletEipSupport()` from #919: one wallet adoption tracker page per tracked EIP, generated at `/wallet/eip/<number>/`.

## What this adds

**Schema** (`src/schema/eip-support.ts`)
- `ratedWalletEipSupport(wallet, eipNumber)`: aggregates a single EIP's support across a rated wallet's variants, returning both per-variant detail and an overall value. Semantics: supported if any variant supports the EIP; verified unsupported only when no variant might support it (i.e. none is unknown); not applicable only when the EIP applies to no variant.

**UI**
- `src/views/EipSupportTable.svelte`: generic table of wallets by EIP support — status, per-platform breakdown for multi-variant wallets, and reference links, sorted supported-first.
- `src/pages/wallet/eip/[eip].astro`: one page per tracked EIP with the EIP's summary/why-it-matters header, followed by software / hardware / embedded wallet tables. Each table is rendered only when the EIP applies to at least one wallet of that kind (e.g. no hardware table on the EIP-6963 page).
- `src/utils/eip-url.ts`: `getEipTrackerUrl()` routes EIP-7702 to its existing richer tracker at `/wallet/7702/` instead of generating a duplicate generic page.
- The `/wallet-eips` directory now links each EIP entry to its tracker page.

**Tests**
- Aggregation semantics covered in `tests/eip-support.test.ts` (ref merging from supporting variants only, unknown-beats-not-supported, not-applicable propagation).
- `[eip].astro` added to the path-naming whitelist alongside `[attrGroupId].astro`.

## Notes

- Pages for sparsely-implemented EIPs (e.g. ERC-5564) will be mostly ❌/unknown at first; "no tracked wallet implements this" is itself the finding, and the pages fill in as wallet data does.
- #878 currently derives per-EIP wallet support from a parallel `relatedAttributeIds` mechanism for the same directory; once this and #919 are in, it can rebase to consume `walletEipSupport()` / `ratedWalletEipSupport()` so the two surfaces cannot disagree.

Closes #871


src/constants/navigation.ts
```diff
@@ -27,14 +27,41 @@ function walletImageIcon(id: string, ext: string): WalletImageNavigationIcon {
...
+		.filter(eip =>
+			wallets.some(
+				wallet => ratedWalletEipSupport(wallet, eip.number).overall !== 'NOT_APPLICABLE',
+			),
+		)
+		.map(eip => ({
+			id: `${idPrefix}-eip-${eip.number}-tracker`,
+			title: `${eipShortLabel(eip)} Tracker`,
+			href: getEipTrackerUrl(eip),
+			icon: 'ICON_CHART_BAR' as const,
```
polymutex: We should probably add an `icon` field to the `Eip` type, which would be optional and this can use `ICON_BAR_CHART` if unset. This would make the navigation list a bit less boring, right now it looks kind of flat with the same icon being used for all of the EIPs.

src/schema/eip-support.ts
```diff
@@ -71,6 +75,15 @@ function browserIntegrationEipSupport(
 	return eipSupport(isSupported(support), hasRefs(support) ? support.ref : undefined, browser.ref)
 }
 
+/**
+ * Interpret a `null` software-wallet-only feature: `resolveFeatures`
+ * hard-nulls such features for other wallet kinds, so for those `null` means
+ * the feature (and its EIPs) cannot apply rather than "unassessed".
+ */
```
polymutex: Instead of this, I think the `Eip` type should have a `appliesTo: NonEmptySet<WalletType>` field that contains this information. Then there's no need for putting that logic here, it would be localized to the individual EIP definitions.

src/utils/eip-url.ts
polymutex, line 1: Not sure we need this library at all... Maybe this and `src/utils/wallet-url.ts` can be merged into a single `urls.ts` library?

tests/eip-support.test.ts
```diff
@@ -85,6 +92,22 @@ describe('walletEipSupport', () => {
...
+		expect(eipSupport['5792']).toBe('NOT_APPLICABLE')
+		expect(eipSupport['7828']).toBe('NOT_APPLICABLE')
+		expect(eipSupport['7831']).toBe('NOT_APPLICABLE')
+
+		// Account type support is a base feature: hardware wallets can
+		// meaningfully support signing EIP-7702 authorizations and ERC-4337
+		// user operations, so unassessed data stays unknown.
+		expect(eipSupport['4337']).toBe('UNKNOWN')
+		expect(eipSupport['7702']).toBe('UNKNOWN')
+	})
```
polymutex: This test should be removed if moving the EIP applicability down into the Eip interface.

tests/eip-support.test.ts
```diff
@@ -388,3 +411,74 @@ describe('walletEipSupport', () => {
...
+
+/** A supported EIP support value referencing the given URL. */
+function supportedEip(url: string): WithRef<Support> {
+	return { ...featureSupported, ref: { url } }
+}
+
+/** A verified-unsupported EIP support value referencing the given URL. */
+function notSupportedEip(url: string): WithRef<Support> {
+	return { ...notSupported, ref: { url } }
+}
```
polymutex: Trivial, can inline.

tests/eip-support.test.ts
```diff
@@ -388,3 +411,74 @@ describe('walletEipSupport', () => {
 		expect(eipSupport['7730']).toBe('UNKNOWN')
 	})
 })
+
+/** A supported EIP support value referencing the given URL. */
+function supportedEip(url: string): WithRef<Support> {
+	return { ...featureSupported, ref: { url } }
```
polymutex: `supported({ ref: { url } })`

tests/eip-support.test.ts
```diff
@@ -388,3 +411,74 @@ describe('walletEipSupport', () => {
...
+}
+
+/** A verified-unsupported EIP support value referencing the given URL. */
+function notSupportedEip(url: string): WithRef<Support> {
+	return { ...notSupported, ref: { url } }
```
polymutex: Use `notSupportedWithRef`

polymutex (changes requested):
Good start, but I think it's a bit too dry/academic. I think we should go for something punchier.

Look at @PatrickAlphaC 's https://erc8213.eth.limo/#/wallets and the "Wallets" section of https://7702beat.eth.sh/ . The wallet list is usually more "screenshottable". Then there's also the "wall of shame" idea, also made for easy screenshottability: https://7702beat.eth.sh/#wall-of-shame

Some ideas on how to achieve this in practice:

- Put a pie chart in the top right (e.g. floating next to the EIP details) showing the breakdown of wallets supporting / not supporting / unrated.
- Add a countdown timer since EIP finalization (may require modifying the EIP type and adding some code to fetch it to the EIP status updater tool)
- Don't use a table. Perhaps just use a grid of wallet icons and their names underneath, then label them as either "supported", "unknown", or "not supported". Show references on hover.
- Maybe show the supported / unsupported sets side by side (could work with the grid approach, have 2 grids)
- Don't sort alphabetically; sort by earliest-support-first (you can figure this out by the `min` of the associated `ref`s' `lastRetrieved` field, though it's sparsely populated; if unset, fallback to sorting by the wallet stage and then internal score as tiebreaker, so that the order is consistent with the Walletbeat homepage).

All of the above are just ideas. Please do **not** implement them in this PR; I think the current page looks good as-is, and I'd prefer merging in this version as-is than holding up this PR. These improvements, if you think they'd be worthwhile, should be done in separate PRs. In the present, feel free to file the ones you agree with as issues on the issue tracker.

@darrylyeo Want to review this for UI change consistency?

ren2140eth:
Thanks will take a look

src/constants/navigation.ts
ren2140eth: Done in 8baecffc — added an optional `icon` field to the `Eip` type (navigation falls back to `ICON_CHART_BAR` when unset) and gave each tracked EIP its own Lucide icon (signature for 712, plug/bell/radar for the provider EIPs, eye-off for 5564, at-sign for 7828, hash for 8213, etc.).

src/schema/eip-support.ts
ren2140eth: Done in 8baecffc — `Eip` now carries `appliesTo: NonEmptySet<WalletType>`, declared in the individual EIP definitions (software-only for the browser EIPs, 5792, 7828 and 7831; all types otherwise). The null-interpretation helper is gone; `eip-support.ts` gates on `appliesTo` in one place. Bonus: the navigation now filters tracker entries by `appliesTo` directly instead of computing per-wallet support.

src/utils/eip-url.ts
ren2140eth: Done in b20ee5b9 — merged `eip-url.ts` and `wallet-url.ts` into a single `src/utils/urls.ts` and updated all importers.

tests/eip-support.test.ts
ren2140eth: Removed in 8baecffc. Test was confirming that 4337 and 7702 stay applicable to hardware wallets. Let me know if you want to keep the regression test that would catch this in.

tests/eip-support.test.ts
ren2140eth: Inlined in 8baecffc.

tests/eip-support.test.ts
ren2140eth: Done in 8baecffc — used `supported({ ref: { url } })`.

tests/eip-support.test.ts
ren2140eth: Done in 8baecffc — used `notSupportedWithRef({ ref: { url } })`.

ren2140eth:
Filed the review's UI ideas (plus an OG share-card proposal that takes the screenshottability angle to its conclusion) as #955 — recorded as a menu of follow-ups worth exploring after this merges, not a commitment to all of them.

src/schema/eips.ts
```diff
@@ -54,6 +59,18 @@ export interface Eip {
...
+	 * The wallet types this EIP can apply to.
+	 * The EIP is treated as not applicable for wallets of any other type.
+	 */
+	appliesTo: NonEmptySet<WalletType>
+
+	/**
+	 * Navigation icon for this EIP.
+	 * Navigation entries fall back to a generic chart icon when unset.
+	 */
+	icon?: LucideNavigationIcon
```
polymutex: Well, since you've now provided it for all EIPs, there's no need to have it optional. You can remove the `?`.

I'd also add `| WBIconID` to the type, to allow these icons to be replaced by proper Walletbeat icons later.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+	import { Variant } from '@/schema/variants'
+
+	const EipSupportStatus = {
+		SUPPORTED: 'SUPPORTED',
+		NOT_SUPPORTED: 'NOT_SUPPORTED',
+		UNKNOWN: 'UNKNOWN',
+		NOT_APPLICABLE: 'NOT_APPLICABLE',
+	} as const
+
+	type EipSupportStatus = (typeof EipSupportStatus)[keyof typeof EipSupportStatus]
```
polymutex: This should just be an `enum`.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+		>
+			{#snippet Cell({ row: wallet, column, value })}
+				{@const walletSupport = supportByWalletId.get(wallet.metadata.id)!}
+
+				{#if column.id === 'wallet'}
+					<div class="wallet-info" data-row>
+						<span class="row-count" data-row="center"></span>
+
+						<img
+							src={`/images/wallets/${wallet.metadata.id}.svg`}
```
polymutex: Use the wallet's `iconExtension`.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+							alt={wallet.metadata.displayName}
+							class="wallet-icon"
+							width="36"
+							height="36"
+							onerror={event => {
+								if (event.currentTarget instanceof HTMLImageElement)
+									event.currentTarget.src = '/images/wallets/default.svg'
+							}}
```
polymutex: No need.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+						<ul class="variant-list" data-list="unstyled gap-1">
+							{#each perVariant as { variant, support } (variant)}
+								{@const status = eipSupportStatus(support)}
+
+								<li data-variant-status={status}>
+									<span class="status-icon">{eipSupportStatusIcon[status]}</span>
+									{variantLabel(variant)}
+								</li>
+							{/each}
+						</ul>
```
polymutex, line 238: This should special-case the (likely) cases that either all variants support it or all variants don't support it, and in either of these cases, just how a single Supported / Not supported checkbox to make it simpler.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+										rel="noopener noreferrer"
+										class="source-link"
+									>
+										{url.label}
+										<span>{@html ExternalLinkIcon}</span>
+									</a>
+								</li>
+							{/each}
+						</ul>
+					{/if}
```
polymutex, line 266: Would be nice to reuse the nice lightbox thing you contributed.
Maybe `ReferenceLinks` needs a "mini" version that could render here?
Can be a follow-up PR.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+
+				{:else}
+					{value}
```
polymutex: When would this ever happen?
If never, then throw error or remove.

src/views/EipSupportTable.svelte
```diff
@@ -0,0 +1,360 @@
...
+				content: counter(TableRowCount);
+			}
+
+			:global([data-disabled]) &::before {
+				content: '–';
+			}
+		}
+
+		.wallet-icon {
+			filter: drop-shadow(rgba(255, 255, 255, 0.1) 0px 0px 4.66667px);
```
polymutex: Seems opinionated and probably doesn't work equally well on light/dark mode.
Check what we do for wallet icons on other pages and use that here, or bring that to other pages. This should probably be a shared CSS filter in a common CSS attribute thing.
cc @darrylyeo

polymutex (changes requested):

src/schema/eips.ts
ren2140eth: Done

src/views/EipSupportTable.svelte
ren2140eth: I actually think its more helpful to have these written out as it is now if a casual user where to see this page and see "platform" only they might not understand what the check means without explicitly labeling the browser/mobile variants. Also separate issue I noticed is that Base as a single platform wallet is labelled "-" i think this is also confusing it should have the green check and the mobile label. Can fix in a follow up commit if you agree

src/views/EipSupportTable.svelte
ren2140eth: Done in 81d43bda.

src/views/EipSupportTable.svelte
ren2140eth: Done in 81d43bda.

src/views/EipSupportTable.svelte
ren2140eth: Removed in 81d43bda.

src/views/EipSupportTable.svelte
ren2140eth: Agreed — filed #957 for the mini `ReferenceLinks` variant + lightbox treatment, as a follow-up PR after this merges.

src/views/EipSupportTable.svelte
ren2140eth: Removed in 81d43bda — all four columns are handled explicitly, so the branch was unreachable.

src/views/EipSupportTable.svelte
ren2140eth: Done in c7cf6bfd — that filter was actually copied verbatim from `WalletTable`/`Eip7702Table`, so I hoisted it into a shared `[data-wallet-icon]` attribute in `css-attributes.css` and pointed all three tables at it. It's now `light-dark()`-aware: unchanged white glow in dark mode, subtle dark shadow in light mode (where the white glow previously did nothing). Happy for @darrylyeo to adjust the values there — it's one definition now.

darrylyeo: This is better as a new variant token on the existing `[data-icon]`, something like `&[data-icon~='shadow']`

darrylyeo: This is better as a new variant token on the existing `[data-icon]`, something like `&[data-icon~='shadow']`

src/pages/wallet/eip/[eip].astro
```diff
@@ -0,0 +1,154 @@
...
+		font-size: 1.05rem;
+	}
+
+	footer {
+		p {
+			color: var(--text-secondary);
+			font-size: 0.9rem;
+		}
+
+		.github-link {
```
Comment (darrylyeo) [inline src/pages/wallet/eip/[eip].astro]: In general, use `[data-icon]` and other existing CSS attributes to reduce the amount of local CSS across the board.

Working guidelines for creating CSS attributes:
- defines one or more local overridable CSS variable parameters
- defines one or more variant tokens for opinionated but commonly reusable presets
- represents a net-new primitive that can't be covered by combos of existing CSS attributes/variants, HTML tag default styles in `global.css`, and primitive Svelte UI components in `src/components`

src/views/EipSupportTable.svelte
ren2140eth: Done in 8738ac5a — `[data-wallet-icon]` is gone; it's now a `shadow` variant token on `[data-icon]` (`&[data-icon~="shadow"]`), with the wallet logos wrapped in a `<span data-icon="shadow">` sized via `--icon-size`. Codegen artifacts (`css-attributes.d.ts`, `.vscode/walletbeat.css-attributes.json`) regenerated.

src/pages/wallet/eip/[eip].astro
Comment (ren2140eth) [reply src/pages/wallet/eip/[eip].astro]: Done in 8738ac5a — the footer's hand-rolled `.github-link` block is gone: the link now uses `data-input` + `data-link="camouflaged"` with a `data-icon` span for the GitHub glyph, so the only local footer CSS left is the paragraph color/size. Applied to both this page and `7702.astro` (which it was copied from).

src/views/EipSupportTable.svelte
polymutex: OK

ren2140eth: updated in 55d705e

polymutex (approved):

PR #926, opened 2026-07-13
CSS: add workaround for stripped `&` selectors during ESBuild minification

### LINKED ISSUE #918: Table of contents right sidebar is hidden behind horizontal scrolling
Loading a per-wallet page in Chromium:

Image

Looks nice, but where's the right sidebar? It seems to be there, but need to horizontally scroll for it (and the horizontal scrollbar is hidden):

Image

On Firefox, same page, the content is cut off on the right side:

Image

The sidebar can also be seen by scrolling it into view, but doing so requires going to a level of horizontal scroll where the content becomes cut off on the left side:

Image

darrylyeo:
Follow-up to https://github.com/walletbeat/walletbeat/pull/829.

Rewrite `&` selectors as `:is(&)` so that [ESBuild does not strip them during CSS minification](https://github.com/evanw/esbuild/issues/4497) and incorrectly change the meaning of `[data-sticky-container]` `:not(& ...)` rules in the production build.

Fixes https://github.com/walletbeat/walletbeat/issues/918.

### DIFF
--- a/src/styles/css-attributes.css
+++ b/src/styles/css-attributes.css
@@ -1236,7 +1236,7 @@ ol {
 	--sticky-sizeInline: var(--sticky0-sizeInline);
 
 	@scope (&) to ([data-scroll-container] > *) {
-		& [data-sticky-container]:not(& [data-sticky-container] [data-sticky-container]) {
+		:is(&) [data-sticky-container]:not(:is(&) [data-sticky-container] [data-sticky-container]) {
 			--sticky-level: 1;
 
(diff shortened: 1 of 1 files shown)


polymutex (approved):
Do you want to add a unit test to ensure `&` isn't accidentally used on its own in this file in the future?

darrylyeo:
No need, I think this is a pretty niche situation. Just reported the issue so hopefully we can revert this if they fix it upstream.

PR #929, opened 2026-07-13
Add ERC-4361 (Sign in with Ethereum) to transaction legibility attribute

### LINKED ISSUE #924: Add Sign in with Ethereum (ERC-4361) to transaction legibility
Last item on the grant proposal we had promised.

https://eips.ethereum.org/EIPS/eip-4361

polymutex:
Fixes #924.

Rates 3 wallets on it.


tests/eip-support.test.ts
```diff
@@ -239,6 +239,7 @@ describe('walletEipSupport', () => {
 				...unratedFeatures().security,
 				transactionLegibility: {
 					ref: { url: 'https://example.com/legibility' },
+					erc4361: null,
```
maykelxyz: Why null? maybe you can change this test from:

`	it('derives EIP-712, ERC-7730 and ERC-8213 from software transaction legibility', () => {`
to:
`	it('derives EIP-712, ERC-7730, ERC-4361, and ERC-8213 from software transaction legibility', () => {`

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1075,8 +1165,17 @@ function evaluateHardwareWalletTransactionLegibility(
...
+		const { erc4361 } = hardwareTransactionLegibility
+
+		if (erc4361 === null) {
+			return unrated(ctx)
+		}
+
+		if (!isSupported(erc4361)) {
+			return hardwareMissingErc4361TransactionLegibility(ctx, hardwareTransactionLegibility)
+		}
+
```
maykelxyz: I suggest moving this to line 1118, so that would look like:

```ts
	const { erc7730, detailsDisplayed, dataExtraction, erc8213, erc4361 } = hardwareTransactionLegibility

	if (
		erc7730 === null ||
		detailsDisplayed === null ||
		dataExtraction === null ||
		erc8213 === null ||
		erc4361 === null
	) {
		return unrated(ctx)
	}
```

src/schema/attributes/security/transaction-legibility.ts
```diff
@@ -1186,8 +1285,17 @@ function evaluateSoftwareWalletTransactionLegibility(
...
+		const { erc4361 } = transactionLegibilitySupport
+
+		if (erc4361 === null) {
+			return unrated(ctx)
+		}
+
+		if (!isSupported(erc4361)) {
+			return softwareMissingErc4361TransactionLegibility(ctx, softwareTransactionLegibility)
+		}
+
```
maykelxyz: Same

maykelxyz (commented):
Looks good, comments are just nits.. Feel free to merge anyway

src/schema/attributes/security/transaction-legibility.ts
polymutex: But if we have all the data but ERC-4361, and that the rating isn't dependent on ERC-4361 support (it almost never is), isn't it nice to produce the rating regardless?

maykelxyz: If that's the case, then I guess we can make it an exception.

polymutex: No, you're right, it's really easy to populate it and all the other ERCs are done at the outset so might as well be consistent. Done.

maykelxyz (approved):
Looks good now :)

PR #931, opened 2026-07-13
<Pie>: centered slice alignment; translucent "unrated" material

### LINKED ISSUE #912: Flower rating chart should always have one petal facing upwards
To match the orientation of the emojis and of typical representation of cherry blossom flowers, the flower rating chart should always have one petal facing straight up, and that petal is the "first" of the set.

For the odd-number-of-petals case, this means the vertical axis should cut through the lower half of the flower, whereas right now it cuts through the top half:

Image

_(Pictured: 5 petals, so needs a clockwise rotation of 36 degrees)_

For the even-number-of-petals case, this means the vertical axis should *not* cut between flower petals (right now it does), and the horizontal axis *should* cut in between flower petals (right now it doesn't):

Image

_(Pictured: 6 petals, so needs a clockwise rotation of 30 degrees)_

Note that this can't be a dumb rotation filter applied to the whole flower pie chart, as the icons on the petals needs to stay rightside-up.

darrylyeo:
* Default pies to have first slice centered north instead of bordering north
  * fixes https://github.com/walletbeat/walletbeat/issues/912
* Fix regression where slice base background colors were no longer neutral
* Add backdrop blur behind "unrated" slices and related badges instead of a solid gray fill

Screenshot 2026-07-13 at 03 50 43

Screenshot 2026-07-13 at 03 38 25 Screenshot 2026-07-13 at 03 57 23

## Components

* `<Pie>`:
  * center the first slice at the top by default
  * remove the wallet page's pie rotation offset
  * drop the solid fill behind transparent gradient stops
  * blur unrated slices against content behind the pie

## Views

* `<ScoreBadge>` / `<WalletStageBadge>`:
  * apply backdrop blur to unrated badge fills

### DIFF
--- a/src/components/Pie.svelte
+++ b/src/components/Pie.svelte
@@ -153,7 +153,11 @@
 			}))
 			.filter(({ weight }) => weight > 0)
 
-		if (colorWeights.length <= 1) return colorWeights[0]?.color ?? slice.color
+		if (colorWeights.length <= 1) {
+			const color = colorWeights[0]?.color ?? slice.color
+
+			return color === slice.gradient.transparentStopColor ? 'transparent' : color
+		}
 
 		const areaRadiusStops = slice.gradient.areaRadiusStops
 		const totalWeight = colorWeights.reduce((sum, entry) => sum + entry.weight, 0)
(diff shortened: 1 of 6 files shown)


polymutex (approved):
🌸

PR #933, opened 2026-07-13
Thread: create ERC-7730 update thread

maykelxyz:


polymutex (changes requested):
I think this is incomplete. Should also:

- Acknowledge current centralization tradeoffs of the repo
- Link to "Wallets that implement clear signing" page (if not our own, Patrick's)
- Reference to your workshop, include slide screenshots and video link

maykelxyz:
> * Acknowledge current centralization tradeoffs of the repo

I was thinking of adding this, but the point I want to make through the thread is why wallets should adopt ERCs (not just clear signing).. So less focus on clear signing itself, but more on ERC adoption in general, and just citing the clear signing as an example.

Will add the other details as I think it helps to prove the point

polymutex (approved):

polymutex:
Looks good but please make vitest happy.

maykelxyz:
Posted here:

https://x.com/walletbeat/status/2077114039042441573

PR #935, opened 2026-07-13
EIP tracker: refresh stale statuses against upstream specs

ren2140eth:
## What

Re-verified the `status` field of all 12 EIPs in the tracker against their upstream front-matter (`eips.ethereum.org` for EIPs, `ethereum/ERCs` for ERCs). Three had drifted:

| EIP | was | now | source |
|-----|-----|-----|--------|
| EIP-7702 | Last Call | **Final** | finalized with Pectra |
| ERC-4337 | Draft | **Final** | `ethereum/ERCs` front-matter |
| ERC-7828 | Draft | **Review** | `ethereum/ERCs` front-matter |

ERC-7828's real status is **Review**, which the `EipStatus` enum did not model — added `EipStatus.REVIEW` and its `Review` label. The other nine statuses were already accurate and are unchanged.

The status chip in `WalletEipList.svelte` colors each status via a `data-status` rule but had none for `REVIEW`, so the new chip would render uncolored — added a violet-400 variant, consistent with the component's existing green/sky/yellow-400 palette, in EIP-process order between Draft and Last Call.

## Verification

- `tsc --noEmit`: clean
- `prettier --check` / `cspell` / `eslint` on the changed files: clean
- Status chip now has a distinct color for every `EipStatus` value.

PR #936, opened 2026-07-14
EIP tracker: add warning-only upstream status checker

ren2140eth:
Follow-up to #935

## What

Adds `pnpm check:eip-status` — a standalone tool that fetches each tracked EIP's upstream `status:` front-matter and reports any drift from what Walletbeat records.

- Checks both `ethereum/EIPs` and `ethereum/ERCs` master per EIP (specs occasionally move between the two repos).
- A spec that 404s in **both** repositories (e.g. an as-yet-unmerged ERC like ERC-8213) is reported as **unverifiable**
- Flags: `--json` (machine-readable, for automation), `--strict` (exit non-zero on drift), `--quiet` (drop the per-EIP "ok" lines). **Default exits 0 — warning-only.**

## Why a script, not a networked vitest test

The vitest suite is fully hermetic (no test touches the network). A live-fetch test would break that and be flaky in CI, so the network call lives in this standalone tool instead — deliberately **not** wired into `check:all`/`check:quick`. A companion **offline** unit test (`tests/eip-status-checker.test.ts`) covers the pure parse/map/URL logic.

A follow-up PR will add a scheduled GitHub Action that runs this on a timer and opens a drift-tracking issue (same pattern as `helios-refresh.yaml`, but it only files an issue).

## Verification

`pnpm check:eip-status` against live upstream: 11 match (including the three refreshed in #935), ERC-8213 unverifiable (pre-merge), **0 drift**. 8 offline unit tests pass.


src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
+import { type Eip, type EipNumber, EipPrefix, EipStatus } from '@/schema/eips'
+import { getErrorMessage } from '@/types/errors'
+
+/** Outcome of comparing one EIP's Walletbeat status against its upstream spec. */
+export enum EipStatusCheckKind {
+	/** Walletbeat's status matches the upstream spec's front-matter. */
```
polymutex: frontmatter (one word). Add to cspell if not recognized.

src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
...
+	MATCH = 'MATCH',
+	/** Both statuses are known but disagree — actionable drift. */
+	DRIFT = 'DRIFT',
+	/**
+	 * The upstream status could not be determined: network error, a 404 in both
+	 * the EIPs and ERCs repositories (e.g. an as-yet-unmerged spec), or a
+	 * front-matter status Walletbeat does not model. Surfaced as a warning,
```
polymutex: Same

src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/** Result of checking a single EIP's status. */
+export interface EipStatusCheckResult {
+	number: EipNumber
+	prefix: EipPrefix
+	friendlyName: string
+	/** Status recorded in Walletbeat. */
+	ours: EipStatus
+	/** Status read from upstream, or null when unverifiable. */
+	upstream: EipStatus | null
```
polymutex: same, replace throughout

src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
...
+	note: string | null
+}
+
+/** Aggregate report over every checked EIP. */
+export interface EipStatusReport {
+	results: EipStatusCheckResult[]
+	matches: EipStatusCheckResult[]
+	drift: EipStatusCheckResult[]
+	unverifiable: EipStatusCheckResult[]
+}
```
polymutex: Highly redundant and ambiguous types. Restructure as:

```typescript
export interface EipCheckResultBase {
  // basic fields common to each check
}

export interface EipCheckResultMatch extends EipCheckResultBase {}

export interface EipCheckResultDrift extends EipCheckResultBase {
  // drift-only fields
}

export interface EipCheckResultUnverifiable extends EipCheckResultBase {
  // "unverifiable"-only fields
}

export interface EipStatusReport {
    // Note the removal of `results` which is redundant
	matches: EipCheckResultMatch[]
	drift: EipCheckResultDrift[]
	unverifiable: EipCheckResultUnverifiable[]
}
```

src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
...
+
+function ercsRepoUrl(number: EipNumber): string {
+	return `${ERCS_REPO_BASE}/erc-${number}.md`
+}
+
+/**
+ * Candidate upstream URLs for an EIP, most-likely repository first. EIPs and
+ * ERCs occasionally move between the two repositories, so both are tried
+ * regardless of the recorded prefix.
+ */
```
polymutex, line 137: If so, this is a form of drift and should be reported as such.

src/tools/eip-status-checker/eip-status-checker-lib.ts
```diff
@@ -0,0 +1,273 @@
...
+	const frontMatter = /^---\r?\n([\s\S]*?)\r?\n---/.exec(markdown)
+
+	if (frontMatter === null) {
+		return null
+	}
+
+	const status = /^status:[ \t]*(.+?)[ \t]*$/m.exec(frontMatter[1])
+
+	return status === null ? null : status[1]
+}
```
polymutex, line 155: Use `parseMarkdownWithFrontmatter` from markdown-utils.ts

tests/eip-status-checker.test.ts
polymutex, line 1: Not a very useful test on its own, since this isn't actually checking anything. Useful for the agent to build the tool, but not really worthwhile ongoing-ly. You can delete.

polymutex (approved):

src/tools/eip-status-checker/eip-status-checker-lib.ts
polymutex: > If so, this is a form of drift and should be reported as such.

Still needs solving

polymutex (changes requested):

src/tools/eip-status-checker/eip-status-checker-lib.ts
ren2140eth: This should be resolved in c9159763 and f1cdb2b. It derives the published EIP/ERC prefix from the destination spec’s category per EIP-1 and compares it to our recorded prefix and reports any mismatch. Missing categories implies EIP. I also added focused regression coverage in 8443a3f4 for stale prefix detection, following Moved tombstones, unmodeled upstream statuses, and missing-category prefix drift.

polymutex (approved):

PR #941, opened 2026-07-15
Security audits bug bounty

### LINKED ISSUE #865: Security audits & bug bounty program
Wallets that aren't hardware wallets are exempted from the bug bounty program attribute. Walletbeat should also evaluate software wallet's bug bounty program:

```ts
// This attribute is only applicable for hardware wallets
// For software wallets, we exempt them from this attribute
if (ctx.features.type !== WalletType.HARDWARE) {
	return exempt(ctx, sentence('This attribute is only applicable for hardware wallets.'))
}
```

maykelxyz:
Attribute: Merge security audits and bug bounty into one attribute

Software wallets were exempted from `bugBountyProgram` entirely (it was hardware-wallet-only), and hardware wallets were exempted from `securityAudits`. Per #865, folds the two into one attribute (`securityAuditsAndBugBounties`, `security-audits-bug-bounty.ts`) so every wallet type is rated on both fronts.

## Changes
- Rating is the worse of the two sub-ratings — a wallet needs to pass both audits and bug bounty to pass overall
  - Either sub-part missing data → whole attribute `UNRATED`
- Deleted `security-audits.ts` and `bug-bounty-program.ts` attribute files, replaced with `security-audits-bug-bounty.ts`
  - `noAudits()`/`audited()` carried over from the old audits file
  - `evaluateBugBountyProgram()` (renamed from `bugBountyAvailable`) carried over from the old bug bounty file, minus the hardware-only exemption
  - New `combineEvaluation()` merges the two sub-results and concatenates `howToImproveMarkdown` from whichever side is lacking
- `attribute-tree.ts` — removed the standalone `bugBountyProgram` entry, security group now just imports `securityAuditsAndBugBounties`
- `software-wallet-stages.ts` — both `security_audit_1y` (stage 1) and `bug_bounty_program` (stage 2) now gate on the combined attribute
- `SecurityAuditsDetailsProps` gained `bugBountyDetails: string`, rendered as a new "Bug bounty program" section in `SecurityAuditsDetails.svelte`
- `completed.tmpl.ts` — added a sample `activeBugBountyProgram` and set `bugBountyProgram: supported(...)` (was `null`), since software wallets are now rated on it too

Closes #865


data/software-wallets/completed.tmpl.ts
```diff
@@ -195,6 +201,36 @@ const recentAudit: SecurityAudit = {
...
+/**
+ * Comprehensive bug bounty program.
+ *
+ * The bug bounty part of security-audits requires:
+ * - availability = ACTIVE
+ * - coverageBreadth = 'FULL_SCOPE'
+ * - non-zero minimum and maximum rewards
+ * - legalProtections supported (ref must be a real URL, not refTodo)
+ */
+const activeBugBountyProgram: BugBountyProgramSupport = {
```
polymutex: Inline?

src/schema/attributes/security/security-audits-bounties.ts
```diff
@@ -0,0 +1,681 @@
...
+ */
+function combineEvaluation(
+	ctx: EvaluationContext<SecurityAuditsMetadata>,
+	auditsPart: SecurityAuditsSubResult,
+	bugBountyPart: BugBountyProgramSubResult,
+): Evaluation<SecurityAuditsMetadata> {
+	const rating =
+		compareExplicitRatings(auditsPart.rating, bugBountyPart.rating) <= 0
+			? auditsPart.rating
+			: bugBountyPart.rating
```
polymutex: This code compares the two parts twice, once here and the other on line 395 (with conflicting preference in the "equal rating" case btw). It would be easier and safer to groups this logic and do the comparison once:

```typescript
const { rating, displayName, shortExplanation } = ((): Pick<...> {
  if (both pass) { return { ... } }
  if (audits is worse) { return { ... } }
  // Otherwise, they are equal or bug bounty is worse
  return { ... }
})()
```

src/schema/attributes/security/security-audits-bug-bounty.ts
```diff
@@ -0,0 +1,681 @@
...
+	auditor: exampleSecurityAuditor,
+	ref: 'https://example.com/audit.pdf',
+	unpatchedFlaws: 'ALL_FIXED',
+	variantsScope: 'ALL_VARIANTS',
+}
+
+export const securityAuditsAndBugBounties: Attribute<SecurityAuditsMetadata> = {
+	id: 'securityAuditsAndBugBounties',
+	icon: 'security_audits',
+	displayName: 'Security Audits & Bug Bounties',
```
polymutex: Thinking about this again, maybe we can drop "Bug" from the name. "Security Audits & Bounties" implies the "bounties" is "security bounties", which is understandable.

src/schema/attributes/security/security-audits-bounties.ts
polymutex, line 1: "audits" is plural, but "bounty" is singular. They should either both be singular, or both plural.

polymutex (changes requested):

src/schema/attributes/security/security-audits-bounties.ts
maykelxyz: Refactored to:

```ts
if (auditsPart.rating === Rating.PASS && bugBountyPart.rating === Rating.PASS) {
	return {
		rating: Rating.PASS,
		displayName: 'Recent security audit and active bug bounty program',
		shortExplanation: sentence(
			'{{WALLET_NAME}} has undergone a recent security audit with all faults addressed, and maintains an active bug bounty program.',
		),
	}
}

return compareExplicitRatings(bugBountyPart.rating, auditsPart.rating) < 0
	? bugBountyPart
	: auditsPart
})()
```

I don't think there's a need for the `if (audits is worse)` branch..

src/views/attributes/security/SecurityAuditsDetails.svelte
```diff
@@ -28,9 +27,7 @@
 
 	const flawStatuses = {
 		FIXED: {
-			label: 'Fixed',
-			color: 'var(--rating-pass)',
-		},
+@/schema/attributes/security/security-audits-bounties
```
polymutex: ?

src/schema/stages/software-wallet-stages.ts
```diff
@@ -547,7 +570,13 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
 					rationale: sentence(
 						'This aligns incentives for security exploits to be reported to the wallet developer, rather than exploited.',
 					),
-					evaluate: variantsMustPassAttribute(softwareWalletVariants, bugBountyProgram),
+					evaluate: variantsMustPassAttribute(softwareWalletVariants, securityAuditsAndBounties, {
+						allowPartial: false,
+						ifUnverifiable: sentence(
+							"{{WALLET_NAME}}'s bug bounty program cannot be publicly verified.",
+						),
```
polymutex: Should be THROW, since this should never happen

src/schema/stages/software-wallet-stages.ts
```diff
@@ -547,7 +570,13 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
...
+						ifNoVariantInScope: null,
+					}),
```
polymutex: This should probably look for the bug bounty part of this, no?
Otherwise it can be confusing when looking at a wallet's stage progress, if the wallet has no audits but does have bounties. This ladder check would incorrectly be shown as fail, even though the wallet has bounties.
I understand that the wallet would be gated on an earlier stage anyway, but the user can still look at each stage's checks independently on the wallet page.
Like the other comment, the logic can be shared.

src/schema/stages/software-wallet-stages.ts
```diff
@@ -186,11 +186,34 @@ export const softwareWalletStageOne: WalletStage<SoftwareAttributeGroupId> = {
...
+							}
+
+							return {
+								rating: StageCriterionRating.PASS,
+								explanation: sentence(
+									'{{WALLET_NAME}} has undergone a security audit within the last year.',
+								),
+							}
+						},
+					),
```
polymutex, line 219: This logic can be defined in an exported function in the attribute, so that both the attribute rating logic and this file can share it.

polymutex (changes requested):

src/views/attributes/security/SecurityAuditsDetails.svelte
maykelxyz: Yeah, that was an import issue, fixed

src/schema/stages/software-wallet-stages.ts
maykelxyz: Alright, I was originally thinking this too:

> I understand that the wallet would be gated on an earlier stage anyway, but the user can still look at each stage's checks independently on the wallet page.

polymutex (approved):

PR #946, opened 2026-07-16
<WalletPage>: scroll-driven flower navigation, native scroll controls, sticky breadcrumb positioning

### LINKED ISSUE #938: Replace per-attribute-group pie chart with global wallet rating pie chart
In #915, @ren2140eth left the following comments:

> Petals are missing when looking at the site via dev server but this is unrelated to this PR
> Screenshot From 2026-07-10 15-17-51
> 
> _Originally posted by @ren2140eth in https://github.com/walletbeat/walletbeat/issues/915#issuecomment-4939978382_

In actuality, the chart is here in his screenshot:

Image

This shows that the pie chart behavior as the user scrolls is non-obvious. I believe lucila also had the same point of confusion.

Two small bugs in addition:

- The spinning is currently incorrect after #931, as the attribute being shown is shown pointing up, whereas it should be pointing leftwards to make sense when the pie chart is on the left edge of the attribute box:
  Image
- The pie chart moves out of view very early in the scroll, for example this is where it started jumped out of its spot for me, where the gut feeling that "something is missing" is strongest (this is the entire verticalness of the browser viewport):
  Image

... But IMO this is actually **best left unfixed** if going with the suggestions that follows.

Now that we have a flower-shaped 2-layer pie chart in the wallet summary page, having a single-layer pie chart in the per-wallet page where the *outer* layer of the other-page's pie chart now becomes the *inner* (and only) layer of the wallet-page pie chart is also confusing: suddenly there's more pie charts and they've swapped places and levels of the hierarchy compared to the previous page.

Last argument would be that it's a bit of a shame that we got this fancy flower control on the wallet comparison page but we don't take advantage of it on the per-wallet page. Wouldn't it be the best place to show a large version of it with the large attribute icons we got?

So: I suggest we simply get rid of the per-attribute-group pie charts. Instead, we can replace them with a single global pie chart at the top of the page. This pie chart can scroll around as needed too, e.g. it can jump into the sticky header as the user scrolls, and highlight the attribute group within the large pie chart that is in view. Crude mockup (I used an HSV filter -35% saturation & -10% lightness, then 40% opacity, feel free to play around with it):

Image

This may require making the attribute group header a bit bigger than this, adding more margins etc. until it looks nice. It also works naturally on mobile as well. A more elaborate version of this may **dynamically adjust the strength of the color filter as a function of what percentage of it is visible in the viewport**. Would probably be really cool and naturally smoothly transition as the page scrolls. Of course, the flower itself would be fully clickable and scroll to the section in question (either to the attribute group section when clicking an inner petal, or the specific attribute when clicking one of the outer "seeds").

Breadcrumb-y header optional, it does have the problem of sometimes being ambiguous as to which attribute to show when there are multiple visible in the viewport. Maybe the whose center is closest to the center of the viewport, maybe the one that is at the highest on the page while still being ≥80% visible, something like that.

Mockup of what the top of the page would look like. Doesn't it look like the large version of the pie chart totally belongs there? (Maybe pushed further inwards so that the content vertically aligns, I didn't squish the header section enough to do so in this crude rendition):

Image

Let me know your thoughts.

darrylyeo:
Follow-up to #829 and #931.

* One wallet report pages, replace per-attribute group pies with one sticky-positioned two-level rating flower that doubles as navigation.
* Reuse flower petal shapes and icons for individual attribute group and attribute explanations.
* Add sticky floating controls for bulk opening/closing collapsibles and scrolling to anchor sections and other points of interest.
* In supporting browsers, introduce dynamic sticky floating breadcrumb headings and badge placement via scroll-driven animations.

Fixes [https://github.com/walletbeat/walletbeat/issues/938](https://github.com/walletbeat/walletbeat/issues/938).

---

## CSS attributes

* `[data-sticky-breadcrumb]`
  * add recursive page / group / attribute breadcrumbs
  * preserve normal-flow heading origins
  * derive positions / exits from shared anchors, timelines and sticky insets
  * exit headings with their containing sections
  * retain normal-flow fallback without anchor positioning / scroll timelines

* `[data-scroll-container]` / `[data-sticky]`
  * derive anchor clearance from sticky insets
  * enable proximity scroll snapping
  * add native previous / next scroll controls
  * stick detail summaries below global header
  * contain controls within respective scroll areas

* `[data-row-item]`
  * bound flexible item sizing

* `[data-list-item-marker]`
  * read marker content from item attributes

* `<details>`
  * normalize content padding
  * remove redundant wrappers

## Components

* `<Pie>`
  * share slice geometry / flower configuration
  * contain slice paint
  * enlarge group icons while preserving label positions
  * honor reduced-transparency preference

* New – `<ScrollAngleSteps>`
  * synchronize both flower levels through shared start-angle steps

* `<NavigationItems>`
  * accept slice geometry
  * support trailing score / stage badges
  * share hover / focus / interest states across matching links
  * preserve existing aside behavior

* `<Background>`
  * simplify CSS value types / mobile effects

## Views

* `<WalletPage>`
  * replace attribute overview pies with sticky `<NavigationItems>` flower
  * match `<WalletTable>` radii, offsets, gaps, corners and label positions
  * place flower in aside on desktop / sticky header on mobile
  * rotate both levels toward current content
  * keep wrapper transparent; paint only slices
  * render matching slice icons in TOC / summaries
  * reuse slice icons in breadcrumbs without slice shapes
  * link flower / TOC / inline-summary hover and focus states
  * show score / stage badges using existing visibility rules
  * restore wallet logo layering, heading markers and title flow origin
  * animate `<h1>` / `<h2>` / open `<h3>` as recursive breadcrumbs
  * move summary badges / trailing content to sticky-header end
  * stick `<h4>` summaries below global header
  * place attribute descriptions under headings within summary rows
  * open attribute / stage details by default
  * add page / current-group disclosure controls

* `<WalletTable>`
  * consume shared flower configuration

* `<WalletStageSummary>`
  * split stage titles / descriptions

## Global styles

* use Space Grotesk for headings, badges and navigation
* inherit Avenir for body content
* preserve monospace styles


polymutex:
FYI #956 bumps to Astro 7 (needed for #873) and a few other packages. I don't think it has bearing on this PR but it did cause some reformatting changes. If you get merge conflicts, pick yours unconditionally and then reformat.

polymutex (commented):
Ready for review?

polymutex:
We'll demo the site to the EF in 48 hours, would love to merge this by then even if incomplete :) Let me know if ready to review/merge.

polymutex (approved):

PR #951, opened 2026-07-17
Tests: no support spread test

### LINKED ISSUE #948: Test: Prevent `{ ...featureSupported }` and `{ ...notSupported }`
Makes data entry more consistent to have only one style of expressing support/non-support.

maykelxyz:
Closes #948

Created test to make data entry more consistent to have only one style of expressing support/non-support.


tests/no-support-spread.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+				fileSet.add(fileMatch.filePath)
+			}
+		},
+		ignore: commonExclusions.concat([
+			'src/schema/features/support.ts',
+
+			// Exclude test files
+			/\.test\.ts$/i,
+		]),
```
polymutex, line 61: Should ignore every non-`.ts` file.

tests/no-support-spread.test.ts
```diff
@@ -0,0 +1,83 @@
...
+		name: '...featureSupported',
+		regexp: /\.\.\.\s*featureSupported\b/,
+		explanation:
+			'Do not spread `featureSupported` to build a supported feature object; use the `supported({ ... })` helper from `@/schema/features/support` instead.',
+	},
+	{
+		name: '...notSupported',
+		regexp: /\.\.\.\s*notSupported\b/,
+		explanation:
+			'Do not spread `notSupported` to build an unsupported feature object; use the `notSupportedWithRef({ ... })` helpers from `@/schema/features/support` instead.',
```
polymutex: The `WithRef` part is only if the notSupported is paired with a `ref` field.

I think checking whether this is the case would be too complex, so I think just changing the error message to something like "if needing a `ref, use the `notSupportedWithRef({ ... })`helper from [....], otherwise use `notSupportedWith({ ... })`".

polymutex (approved):

polymutex (approved):

Issue #958, opened 2026-07-19
2026 roadmap


polymutex:
We need an update on the 2025 roadmap as part of our grant funding search. Going in that search without a roadmap means we wouldn't know what we're asking to be funded.

PR #960, opened 2026-07-20
Add 2026 roadmap.

polymutex:
Fixes issue #958.


maykelxyz (commented):
Looks good to me.

A bit surprised that finishing hardware wallet methodology would be a non-goal for 2026.. Maybe because I was kinda looking forward to the thought of Walletbeat covering it

darrylyeo (approved):

governance/roadmap/2026/2026-roadmap.md
```diff
@@ -0,0 +1,263 @@
...
+- Reach out to at least **3 potential funding sources**. Top candidates:
+  - Ethereum Foundation 1TS, as continuation of existing grant funding.
+    - Rationale for funding: Continuation of Clear Signing and other wallet security initiatives, e.g. ERC-8213 tracking.
+  - Other parts of the Ethereum Foundation worth reaching out to: Ecosystem Support, or Access Layer.
+    - Rationale for funding: Renewed EF focus on other CROPS-y aspect of wallets, especially privacy. Good synergy with EF's Kohaku initiative.
+  - Ludlow Institute.
+    - Rationale for funding: Valuable privacy research for web3 wallets.
+    - Blocked on additional maturity of the wallet data collection tool, and fully-rated wallets using this tool.
+- Create **non-grant-based sources of funding**.
+  - Accept crypto donations. This may include creating "badge NFTs" (commemorative only, absolutely no governance implications) similar to e.g. Rotki per-release funding NFT badges.
```
johngrantuk, line 74: Before committing to the NFT badges maybe it would be possible to quantify the expected benefit and resource required to implement? Feels like the demand could be pretty small (e.g. as demonstrated by the shut down of POAP) and anyone donating would likely do it without that incentive so could end up being a bit of a waste of valuable resources.

johngrantuk (approved):

maykelxyz (approved):

ren2140eth:
A very lightweight version of this would simply be linking our Giveth page to our website since we can take donations on Giveth at any time

ren2140eth:
This generally looks good to me. My only feedback would be is it worth to convert the treasury 100% into USDC today? This would allow us to be a lot more concrete in terms of what runway we do have and the price of ETH has rebounded significantly off of the recent lows. Given the existing runway is already minimal it might make sense to just "lock it in" now.

ren2140eth (approved):

polymutex:
@johngrantuk Agree in general that we should expect the revenue from this to probably be in the 0.1ETH to 1ETH range.
I still think it's worth pursuing because most of the work involved there would be by @tesera5 as the artist behind the NFTs, not blocking other development work, and not blocking other branding resources/artwork per this timeline.

I also agree with @ren2140eth that a minimal version of this may be to just link to Giveth. One thing I'm a bit concerned about is the lack of attribution of incoming funds this way, which seems like something that needs solving regardless of the funding mechanism.

polymutex:
> This generally looks good to me. My only feedback would be is it worth to convert the treasury 100% into USDC today? This would allow us to be a lot more concrete in terms of what runway we do have and the price of ETH has rebounded significantly off of the recent lows. Given the existing runway is already minimal it might make sense to just "lock it in" now.

Well, fellow co-treasurer... In my opinion, as an Ethereum steward, IMO we should never go below 50% ETH. So the question might be "why not lock it now by selling 50% ETH and buying USDC with it". To which I would say that we've been historically absolutely terrible at timing the market, and that if we assume this streak of bad luck will continue, then we may as well spread it out to get something closer to the average price.
Not a strongly held opinion though. I suck at market timing anyway. If you feel strongly confident, then we could push for one big 50% transaction. The timeline of the sales are intended as a "should have done at least X% by that date", not necessarily a "do exactly X% by that time".

polymutex:
> A bit surprised that finishing hardware wallet methodology would be a non-goal for 2026.. Maybe because I was kinda looking forward to the thought of Walletbeat covering it

I just don't see it realistically happening by then, if we consider the time we'll spend on grant-hunting, finishing up all the loose ends for launch, increased social media focus, and all the prep we may have to do for Devcon.
But we should definitely aim to make a solid dent in it!

ren2140eth:
Don’t feel super strongly either was thinking it could be easier from a planning perspective b/c we could then say we must complete x by x date based on the stablecoin runway alone and the eth portion just serves as buffer beyond that so even if eth trades down 50% by the time we are out of stables we are sort of prepared for that scenario.

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

lucilapastore:
This generally looks good to me too. I also think we should convert 100% of the treasury into USDC today. This would make us a lot more stable in terms of our runway.

I agree with all the priorities and the tasks assigned to me, but I think the donations page should be done and available ASAP so I can promote it on our social media channels.

I am aware that our resources are very limited, but we have to pursue a strong presence during Devcon Mumbai, leading initiatives like the privacy hub and the cypherpunk congress. It will be the event of the year, and the stronger our presence is there, the higher our chances will be to get a new grant, partner or source of funding.

PR #961, opened 2026-07-20
Transaction legibility: evaluation functions

### LINKED ISSUE #939: Transaction legibility: Use higher-level evaluation functions
When implementing #929, I found that the `ratingScale` examples for the transaction legibility attributes all use "leaf"-level evaluation functions that only ever return a static value. This is sufficient for displaying the scale, but means we don't get the free testing advantage that the `ratingScale` doubles as in `tests/attribute-integrity.test.ts`. The `ratingScale` examples should call the highest-level function possible that doesn't take a full `ResolvedFeatures` in order to exercise more of the attribute evaluation logic.

(Why I noticed: I added ERC-4361 and the tests didn't fail when I had it set to `null` while expecting it to return `UNRATED`)

maykelxyz:
### Chore: Use higher level eval function

`transactionLegibility` examples called the rating-specific helpers directly with hand-picked (often `null`) fields, bypassing the real dispatch logic.

Switched them to `evaluateHardwareWalletTransactionLegibility` / `evaluateSoftwareWalletTransactionLegibility`, the same functions `evaluate()` uses, and filled in the fields needed for each to make sure it's on its intended rating.

Closes #939


src/schema/attributes/security/transaction-legibility.ts
polymutex, line 1: Please add some helper `const`s for fully-supported subfields (e.g. `erc7730FullySupported`, same for the other ERC) to make the rating scale easier to review.

polymutex (changes requested):

polymutex (approved):

PR #962, opened 2026-07-20
Feat: Stage updates

maykelxyz:
Stages: Add more stage 1 and 2 criteria

### Stage 1
- `account_recovery_drills` — wallet must periodically prompt users to verify their recovery methods are still accessible
- `chain_abstraction` — wallet must abstract away cross-chain balance/bridging complexity (`chainAbstraction` attribute, `allowPartial: true`)

### Stage 2
- `account_recovery` — wallet must implement guardian-based recovery that covers all likely catastrophic scenarios
- `full_l1_provider_independence` — wallet must not depend on external providers for basic L1 operations, even with a self-hosted node configured
- `hardware_wallet_interoperability` — wallet must directly support hardware wallets from at least three manufacturers, not just reach them via WalletConnect
- `app_isolation` — wallet must default to per-app accounts when connecting to apps
- `orderflow_transparency` — wallet must transparently disclose how it monetizes or shares transaction data before onchain inclusion

`account-recovery.ts` added two new exported helpers used by the new criteria:
```ts
export function hasAccountRecovery(accountRecovery: AccountRecovery): boolean
export function hasAccountRecoveryDrills(
	accountRecovery: AccountRecovery,
	accountSupport: AccountSupport,
): boolean | null
```

## Other changes
- Registered `orderflowTransparency` in the attribute tree (`src/schema/attribute-tree.ts`).


src/schema/stages/software-wallet-stages.ts
```diff
@@ -892,6 +1050,26 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
 					evaluate: variantsMustPassAttribute(softwareWalletVariants, transactionBatching),
 					displayName: 'Transaction Batching',
 				},
+				{
+					id: 'hardware_wallet_interoperability',
```
polymutex: This should replace `hardware_wallet_subset`.

src/schema/stages/software-wallet-stages.ts
```diff
@@ -892,6 +1050,26 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
...
+					description: sentence(
+						'The wallet must directly support hardware wallets from at least three major manufacturers.',
+					),
+					rationale: sentence(`
+						Counting supported hardware wallet manufacturers alone is not sufficient: a wallet that only reaches most of them through WalletConnect still leaves users dependent on an intermediary and a separate external application.
+						True hardware wallet interoperability requires direct integration, which preserves a competitive, open hardware wallet market.
```
polymutex: This explanation makes no sense on the ladder on its own, it only makes sense in relation to what this entry replaces

src/schema/stages/software-wallet-stages.ts
```diff
@@ -313,6 +322,52 @@ export const softwareWalletStageOne: WalletStage<SoftwareAttributeGroupId> = {
 					}),
 					displayName: 'Standard Security Practices',
 				},
+				{
+					id: 'account_recovery_drills',
```
polymutex: Account recovery should come before drills, no?

src/schema/stages/software-wallet-stages.ts
```diff
@@ -562,6 +617,21 @@ export const softwareWalletStageOne: WalletStage<SoftwareAttributeGroupId> = {
 					}),
 					displayName: 'Browser Integration',
 				},
+				{
+					id: 'chain_abstraction',
```
polymutex: IMO we should make this stage 2. It's not super critical anymore vs the time it was written, most wallets already do something here, and we should keep stage 1 ladder short to the extent we can help it.

src/schema/stages/software-wallet-stages.ts
```diff
@@ -646,6 +716,44 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
 					}),
 					displayName: 'Duress Resistance',
 				},
+				{
+					id: 'account_recovery',
```
polymutex: Being able to properly recover one's account definitely doesn't strike me as a stage-2 thing... IMO it should be stage 1 if not stage 0.5, and if it's stage 1, we could have an easier variant at stage 0.5 for something like EOA drills, **but with this requirement waived if guardian recovery is implemented**.

src/schema/stages/software-wallet-stages.ts
```diff
@@ -646,6 +716,44 @@ const softwareWalletStageTwo: WalletStage<SoftwareAttributeGroupId> = {
...
+						'The wallet must implement guardian-based account recovery that lets users recover their account in all likely catastrophic scenarios.',
+					),
+					rationale: sentence(`
+						Self-custody is only practical for everyday users if losing a
+						device, a guardian, or a single external provider does not mean
+						permanently losing access to one's account. Guardian-based
+						recovery provides this safety net without reintroducing a single
+						party that can take over the account.
+					`),
+					evaluate: stageCriterionEvaluationPerVariant(
```
polymutex: Why not simply make it attribute-based here?

polymutex (changes requested):

src/schema/stages/software-wallet-stages.ts
```diff
@@ -187,6 +193,66 @@ export const softwareWalletStageZeroFive: WalletStage<SoftwareAttributeGroupId>
...
+
+							return {
+								rating: StageCriterionRating.PASS,
+								explanation: sentence(
+									'{{WALLET_NAME}} periodically prompts users to verify their recovery material is still accessible.',
+								),
+							}
+						},
+					),
+					displayName: 'Account Recovery Drills',
```
polymutex: Let's call it "Basic Account Recovery Assurance", since it can also be fulfilled by guardian recovery. This title would cover the rough intent that both pathways would satisfy. Feel free to come up with a better one or ask claude for a better one though... just has to be short enough to pass as a title

src/schema/stages/software-wallet-stages.ts
```diff
@@ -187,6 +193,66 @@ export const softwareWalletStageZeroFive: WalletStage<SoftwareAttributeGroupId>
...
+							}
+
+							if (isSupported(variantWallet.features.security.accountRecovery.guardianRecovery)) {
+								return {
+									rating: StageCriterionRating.PASS,
+									explanation: sentence(
+										'{{WALLET_NAME}} implements guardian-based account recovery.',
+									),
+								}
+							}
```
polymutex: This should be based on the attribute evaluation; just "supporting" guardian-based recovery can mean lots of things which aren't all really equal to one another (hence the attribute logic being the monster it is)

src/schema/stages/software-wallet-stages.ts
```diff
@@ -187,6 +193,66 @@ export const softwareWalletStageZeroFive: WalletStage<SoftwareAttributeGroupId>
...
+
+							const passes = hasAccountRecoveryDrills(
```
polymutex: `const hasDrills` to disambiguate with the above.

polymutex (changes requested):

src/schema/stages/software-wallet-stages.ts
maykelxyz: Ah yeah. Replaced with `hasAccountRecovery`.

polymutex: Can you use the full attribute logic instead? i.e. just call `variantsMustPassAttribute` directly within the function. This ensures the logic between this and the "pure" `variantsMustPassAttribute`-based attribute later are exactly consistent, avoid the possibility of helper function drift (e.g. if we add some other top-level check in the Account Recovery `evaluate` function that doesn't go through the `hasAccountRecovery` function)

src/schema/attributes/security/account-recovery.ts
```diff
@@ -58,23 +58,53 @@ export type AccountRecoveryMetadata = {
...
 			recovery: AccountRecoveryOutcomeCannotBeRecovered
 		} => !isAccountRecoverable(outcome.recovery),
 	)
+}
+
+/**
+ * Whether the wallet implements guardian-based account recovery that lets
+ * the user recover their account in all likely catastrophic scenarios.
+ */
+export function hasAccountRecovery(accountRecovery: AccountRecovery): boolean {
```
polymutex: (Obsolete)

polymutex (changes requested):

src/schema/stages/software-wallet-stages.ts
```diff
@@ -241,47 +314,23 @@ export const softwareWalletStageOne: WalletStage<SoftwareAttributeGroupId> = {
 					displayName: 'Security Audits',
 				},
 				{
-					id: 'hardware_wallet_subset',
+					id: 'hardware_wallet_interoperability',
 					description: sentence(
-						'The wallet must support hardware wallets from at least three manufacturers.',
+						'The wallet must directly support hardware wallets from at least three major manufacturers.',
```
polymutex, line 319: 👍

src/schema/stages/software-wallet-stages.ts
```diff
@@ -241,47 +314,23 @@ export const softwareWalletStageOne: WalletStage<SoftwareAttributeGroupId> = {
...
 					),
 					rationale: sentence(`
-            By letting you offload your private key to a separate hardware
-            device, the wallet developer demonstrates that they are serious
-            about ensuring the security of your private key.
-          `),
-					evaluate: stageCriterionEvaluationPerVariant(
+						Direct integration with hardware wallets, rather than reliance on an intermediary and a separate external application, keeps users independent of any single go-between and preserves a competitive, open hardware wallet market.
```
polymutex: Emphasizing the "direct" in the description is good, but again here this rationale alone sounds like it's as a counterfactual to the version of this ladder item that was removed, which will not make sense when read in the context of the `/stages` page without the history. Needs to be a rationale of the hardware wallet interop attribute in isolation.

polymutex (approved):

polymutex (approved):

PR #966, opened 2026-07-22
Fix/transaction batching support

### LINKED ISSUE #832: Transaction batching support should fail if wallet doesn't support atomic batching and doesn't populate the `receipts` array correctly
Found during manual playground experimentation. This should be a FAIL case, not a PARTIAL case.

maykelxyz:
Fix: transaction batching & Connect emit test 

## Changes
- Poll `wallet_getCallsStatus` instead of reading it once -  bundles are frequently still pending right after `wallet_sendCalls` returns, so a single read was catching wallets mid-flight and reporting them as failing `atomic`/`receipts` checks
  - Polls every `2s` up to a `60s` timeout via `pollCallsStatus`
  - Handles both EIP-5792 status shapes: v2.0.0 numeric codes (`100` = pending, `>=200` terminal) and v1.0.0 string statuses (anything other than `"PENDING"` is terminal)
- Fix early connection detection: the `connect` event usually fires at provider-discovery time (page load), long before the user reaches step 2, so a step-scoped listener was missing it and reporting `connectEventFired: false` for wallets that connect early
  - Added `attachEarlyConnectListener` in `WalletTest.svelte`, wired up both on EIP-6963 discovery and as a fallback on `window.ethereum`
  - `runStep2Connect` now prefers the early-captured event via `ctx.getEarlyConnectEvent()`, falling back to its own step-scoped listener for wallets that connect lazily on account authorization

## Other changes
- Explicit `$state<T>()` type params for `transactionState`, `signatureState`, `stepTestState`, `scamAlertState` in `WalletTest.svelte`

Closes #832


polymutex (commented):
Change looks OK but I don't see how this solves the issue described in #832, i.e. the fact that a wallet doesn't populate `receipts` is treated as `PARTIAL`.

maykelxyz:
> Change looks OK but I don't see how this solves the issue described in #832, i.e. the fact that a wallet doesn't populate `receipts` is treated as `PARTIAL`.

Updated `Valid receipts array` to required. Yeah. The changes earlier were for ensuring the tests actually populate the receipts. Glad to also report that Ambire passes **all required and optional** specs for 6963, 1193, 2700, & 5792 :)

polymutex (approved):

PR #971, opened 2026-07-26
Data: Fill Rainbow scamAlerts (mobile)

ren2140eth:
Fills `security.scamAlerts` for the Rainbow mobile app. Scoped to `[Variant.MOBILE]`, with `[Variant.BROWSER]: null` pending separate testing of the extension.

## Results

| Field | Value |
| --- | --- |
| `scamUrlWarning` | supported, `DOMAIN_ONLY` |
| `contractTransactionWarning` | supported (`contractRegistry`) |
| `sendTransactionWarning` | not supported |
| `unlimitedApprovalWarning` | not supported |

Rating: **FAIL** on mobile (`leak_domain` — warns about scams, but leaks browsed domains). Overall stays unrated while the browser variant is null.

## Notes for review

**Rainbow's own test dapp made the warnings observable.** `bx-test-dapp.vercel.app` is flagged in Rainbow's registry, so both warnings could be captured firing without visiting a live phishing site. Found via their e2e flow `e2e/flows/screens/MaliciousDappTransactionWarning.yaml`.

**The assessment is per-request, not inherited from the requesting site.** A send to the burn address, requested by that same flagged app, drew no warning. That is why `unlimitedApprovalWarning` is false: Rainbow warns when a request is assessed malicious, never because an allowance is unlimited.

**These values use the Walletbeat testing suite as it stands today.** This is specifically for the recent contract warning test, which we can revisit once refreshed.

**Evidence for the two unsupported fields does not render.** `rateLeakBasedWarning` replaces the ref with `refNotNecessary` on its not-supported branch, and `ScamAlertDetails.svelte` gates each `ReferenceList` on `isSupported`. So the four screenshots and explanations backing those verdicts are visible only in this diff. Would it be worth adjusting the schema so we can surface this evidence?

**The address-poisoning capture shows only the second address.** The look-alike relationship is not visible in the image itself. The Etherscan history for both addresses confirms the ordering, and can be shared on request.

Source references are pinned to `a37815f4`.


data/software-wallets/rainbow.ts
```diff
@@ -731,7 +736,119 @@ export const rainbow: SoftwareWallet = {
...
+									'Connecting to an app that Rainbow flags shows this warning in practice. The app\'s address appears in red beside a warning symbol, above an alert reading "This app is likely malicious".',
+								file: 'public/references/wallets/rainbow/screenshots/2026-07-26-scam-alerts-malicious-dapp-connect-warning.png',
+								label:
+									'Rainbow mobile connection screen for a flagged app, showing its address in red and an alert reading "This app is likely malicious".',
+								lastRetrieved: '2026-07-26',
+							},
+						],
+						leaksUserAddress: false,
+						leaksUserIp: true,
+						leaksVisitedUrl: 'DOMAIN_ONLY',
```
polymutex, line 814: Are you sure this is DOMAIN_ONLY, not the full URL? The GraphQL link above has the variable name `url`, unclear if that's set to domain only or full URL.

polymutex:
> **Evidence for the two unsupported fields does not render.** `rateLeakBasedWarning` replaces the ref with `refNotNecessary` on its not-supported branch, and `ScamAlertDetails.svelte` gates each `ReferenceList` on `isSupported`. So the four screenshots and explanations backing those verdicts are visible only in this diff. Would it be worth adjusting the schema so we can surface this evidence?

Yes, as a separate PR.

data/software-wallets/rainbow.ts
ren2140eth: Yes this should be confirmed by the ref in line 795

polymutex: Perfect

polymutex (approved):

PR #984, opened 2026-07-28
Data: Add rainbow browser extension scam alerts

maykelxyz:
## Rainbow data
- Filled `scamAlerts[Variant.BROWSER]`, mirroring the mobile evidence where the same Blockaid-backed checks apply
  - `contractTransactionWarning`: supported, leaks contract address, user address, and IP (server-side simulation/scan)
  - `scamUrlWarning`: supported, leaks full URL and IP but not the user address
  - `sendTransactionWarning`: not supported
  - `unlimitedApprovalWarning`: not supported

## Other changes
- Ignore the "dApp" site-convention lint when it's immediately followed by a file extension don't trigger the grammar check

### DIFF
--- a/tests/utils/grammar.ts
+++ b/tests/utils/grammar.ts
@@ -352,6 +352,22 @@ export async function grammarLintMessages(
 		lint => lint.lint_kind_pretty() !== 'Word Choice' || lint.get_problem_text() !== 'lockdown',
 	)
 
+	// Ignore the "dApp" site convention lint when it's immediately followed by a source-file
+	// extension, e.g. the upstream filename `dapp.ts` inside an auto-generated GitHub blob
+	// label such as "dapp.ts"
+	lints = lints.filter(lint => {
+		if (
+			lint.lint_kind_pretty() !== 'Site convention' ||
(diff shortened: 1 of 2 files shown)


tests/utils/grammar.ts
```diff
@@ -352,6 +352,22 @@ export async function grammarLintMessages(
...
+			lint.lint_kind_pretty() !== 'Site convention' ||
+			lint.get_problem_text().toLowerCase() !== 'dapp'
+		) {
+			return true
+		}
+
+		const after = trimmedText.slice(lint.span().end, lint.span().end + 10)
+
+		return !/^\.[a-z]{1,10}\b/.test(after)
+	})
```
polymutex, line 369: This is well-intentioned but it is overfitted to the particular case here. It would break if `dapp` wasn't the final word in the filename, e.g. `https://example.com/dapp/manifest.js`. The underlying problem is the need to ignore words that are parts of URLs, and words that are part of URLs that get rendered as reference labels.

Added a fix that properly ignores URLs and GitHub ref labels.

polymutex (approved):

PR #1010, opened 2026-08-02
Data: Add Rainbow RNG source evidence

ren2140eth:
## Summary

- cite Rainbow browser wallet generation and its browser-provided CSPRNG
- cite Rainbow mobile wallet generation and the Android/iOS OS CSPRNG providers
- leave the existing OS_CSPRNG feature values and ratings unchanged

## Context

Rainbow already passes the secure RNG sub-evaluation, but its existing references only established key storage. This adds variant-scoped, immutable source evidence following the precedent used for other software wallets.

### DIFF
--- a/data/software-wallets/rainbow.ts
+++ b/data/software-wallets/rainbow.ts
@@ -86,7 +86,7 @@ export const rainbow: SoftwareWallet = {
 		`),
 		contributors: [polymutex, mattmatt, ren2140],
 		iconExtension: 'svg',
-		lastUpdated: '2026-08-01',
+		lastUpdated: '2026-08-02',
 		urls: {
 			androidManifestXml:
 				'https://raw.githubusercontent.com/rainbow-me/rainbow/develop/android/app/src/main/AndroidManifest.xml',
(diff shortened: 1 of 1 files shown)


data/software-wallets/rainbow.ts
```diff
@@ -969,6 +981,24 @@ export const rainbow: SoftwareWallet = {
...
+								"In production, the random-values provider used by Rainbow's mobile app obtains random bytes from Java `SecureRandom` on Android.",
+							label: 'Rainbow mobile Android CSPRNG provider',
+							url: 'https://github.com/LinusU/react-native-get-random-values/blob/1bb6c57a29bf197aebd31b1327d7bba58ab296fe/android/src/main/java/org/linusu/RNGetRandomValuesModule.java#L25-L34',
+						},
+						{
+							explanation:
+								"In production, the random-values provider used by Rainbow's mobile app obtains random bytes from Security.framework `SecRandomCopyBytes` on iOS and fails if that OS call is unsuccessful.",
+							label: 'Rainbow mobile iOS CSPRNG provider',
+							url: 'https://github.com/LinusU/react-native-get-random-values/blob/1bb6c57a29bf197aebd31b1327d7bba58ab296fe/ios/RNGetRandomValues.m#L12-L20',
+						},
```
polymutex: Line numbers are off in both of these refs.
Also, these point to the underlying libraries that I'm sure Rainbow uses, but we need evidence of that too, otherwise these refs don't mean much.
The right way to do this is to have multiple code pointers; note that `url` can be set to an array of URLs, which is useful for this case.

polymutex (changes requested):

data/software-wallets/rainbow.ts
ren2140eth: leaving this to point to evidence in the Rainbow repo only

polymutex (approved):

PR #1037, opened 2026-08-06
Data: Add rainbow data collection & data collection CLI improvements

maykelxyz:
Data: Generate Rainbow's dataCollection from capture data

## Changes
- Populated `privacy.dataCollection` for Rainbow `BROWSER` from the reviewed capture. `MOBILE` stays `null` — no real mobile capture yet, so it's honest instead of borrowing browser data like the old manual block did.
- Added `generate-data-collection.ts` + `parse-data-collection.ts`. Parser validates the generated JSON into a real `VariantFeature<DataCollection>` (`Enum.is()`, `entityById`) 
- Fixed `unreviewableRequests()`  - its filter was inverted, so it returned reviewable requests instead of the ones still blocked on unclassified strings.
- Non-ASCII strings (e.g. binary Sentry payloads) now get base64-encoded before being written into `.annotations.json`, instead of as raw text that can corrupt the file.
- `review-strings` now hides binary payloads (`looksBinary`, already used for request `content`) instead of dumping raw bytes to the terminal. Display-only, doesn't affect review/classification.

## Other changes
- Added `cloudinary`, `imgix`, `microsoft`, `rudderstack` entities and domain mappings.
- Updated `data/collection/global.annotations.json` from reviewing Rainbow's requests.


polymutex (changes requested):
Please split this into multiple PRs, this is too much to review at once.

maykelxyz:
Split into multiple PRs: #1040 #1042 #1043

PR #1040, opened 2026-08-07
Fix: Decode binary strings, unreviewed filter, and ISO timestamp as benign strings

maykelxyz:
# Fix: Decode binary strings, unreviewed filter, and ISO timestamp as benign strings

## Changes
- Fixed `unreviewedRequests()`/`prioritizedUnreviewedRequests()` filtering with the inverted condition
- Binary payloads (e.g. Sentry `replay_recording` envelopes) no longer get stuffed into `EncodedUserDataString.str` as raw text. Added `strBase64` as an alternate encoding, chosen automatically in `encode()` when the string isn't ASCII
  - Added `looksBinary()` in `wallet-data-collection-lib.ts` to detect these before printing, so `displayRequestInfo`.
- Added ISO-8601 timestamps to `GLOBAL_BENIGN_REGULAR_EXPRESSIONS`


polymutex:
Importing some of these ideas in #1041. I think the only surviving thing left here is the ISO-8601 timestamps thing.

polymutex (approved):

PR #1042, opened 2026-08-07
Feat: create generator and parser for dataCollection

maykelxyz:


src/tools/wallet-data-collection/generate-data-collection.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const id = args.id
+
+if (id === undefined || id === '') {
+	throw new Error('Missing required --id=<walletId> argument')
+}
+
+const type = args.type ?? 'software'
+
+const repoDir = path.join(path.dirname(fileURLToPath(import.meta.url)), '../../..')
```
polymutex, line 41: `getRepositoryRoot`

polymutex (commented):
This PR highlights a real problem (import cycle) but I don't think this is the right solution. I will work on a different solution, which is static pre-rendered data collection JSONs, similar to what we do for manifests. No need for re-parsing/re-typechecking code this way.

PR #1063, opened 2026-08-11
Fix: Recognize user data in reviewing strings

### LINKED ISSUE #1062: Data Collection CLI: Entropy detection incorrectly flags strings containing wallet addresses
https://github.com/walletbeat/walletbeat/pull/1043#issuecomment-5253857379

maykelxyz:
Fix: Recognize user data in reviewing strings

`review-strings` kept flagging strings with an embedded wallet address (e.g. `eip155:1:0x41c36c21...`) as unclassified, even after the address was already declared. Two bugs: the substring cross-check only looked at strings already discovered standalone, not the full `userData` store, and the match was case-sensitive so checksummed vs. lowercase addresses never matched.

Fixes #1062

### DIFF
--- a/src/tools/wallet-data-collection/wallet-capture-file.ts
+++ b/src/tools/wallet-data-collection/wallet-capture-file.ts
@@ -349,6 +349,22 @@ function _isEthereumAddress(s: string): boolean {
 	return s.length === 42 && s.startsWith('0x', 0)
 }
 
+/**
+ * Splits `str` on occurrences of `existingStr`, like `String.prototype.split`,
+ * except that when `existingStr` is an Ethereum address, the match is done
+ * case-insensitively: EIP-55 checksum casing is a cosmetic convention, so
+ * `0xABC...` and `0xabc...` refer to the same address and should be treated
+ * as the same substring. Non-address substrings are matched as-is (case
+ * sensitively). Returns `null` when there is no match.
+ */
+function splitOnKnownSubstring(str: string, existingStr: string): string[] | null {
+	const bits = _isEthereumAddress(existingStr)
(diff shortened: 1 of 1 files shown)


src/tools/wallet-data-collection/wallet-capture-file.ts
```diff
@@ -349,6 +349,22 @@ function _isEthereumAddress(s: string): boolean {
...
+ * as the same substring. Non-address substrings are matched as-is (case
+ * sensitively). Returns `null` when there is no match.
+ */
+function splitOnKnownSubstring(str: string, existingStr: string): string[] | null {
+	const bits = _isEthereumAddress(existingStr)
+		? str.split(new RegExp(escapeRegExp(existingStr), 'i'))
+		: str.split(existingStr)
+
+	return bits.length > 1 ? bits : null
+}
```
polymutex, line 366: This needs to be cached on a per-UserDataString basis to avoid the overhead of recomputing this on every split

src/tools/wallet-data-collection/wallet-capture-file.ts
```diff
@@ -1866,13 +1880,32 @@ export class WalletDataStrings {
...
+				if (s.str.pieces.size > 0) {
+					candidates.add(s.str.str)
+				}
+			}
+
+			for (const ud of this.captureFile.userData._index.values()) {
+				if (ud.pieces.size > 0) {
+					candidates.add(ud.str)
+				}
+			}
```
polymutex, line 1906: This seems wrong. They should already be loaded as part of the capture file loading. I'll have to investigate...

PR #1066, opened 2026-08-11
Wallet page: declutter TOC and collapse defaults

### LINKED ISSUE #1024: Wallet page: Remove bottom sidebar navigation buttons
They replicate scrolling functionality, which isn't super useful when the sidebar itself serves as a scrollable table of contents. We can remove the buttons and reclaim some space.

### LINKED ISSUE #1025: Wallet page: Remove "Table of contents" header label
Easy vertical space reclaim.

### LINKED ISSUE #1029: Wallet page: Right sidebar: Remove stage labels
They are visually heavy and take a lot of horizontal space within the sidebar, causing some attributes to span 2 lines. We can just remove them; this information is already present in two other places of the page (the top stage progress, and the attribute box in the main page body).

(They are also incorrect: off-by-one, due to being named by index rather than by stage label, but that will be fixed by the act of removing them...)

lucilapastore:
## Summary
- Remove bottom TOC sidebar chrome (list/rows toggles, scroll markers, ↑↓ buttons)
- Drop the visible “Table of contents” header treatment and collapse stage/attribute details by default
- Stop auto-expanding the left nav for the current page
- Hide stage badges in the wallet page TOC

## Test plan
- [ ] Open a wallet page and confirm the TOC sidebar no longer has bottom nav controls
- [ ] Confirm stages/attributes start collapsed
- [ ] Confirm left nav no longer auto-expands the current section
- [ ] Confirm TOC items no longer show stage badges

### DIFF
--- a/src/views/NavigationItems.svelte
+++ b/src/views/NavigationItems.svelte
@@ -82,11 +82,6 @@
 	let effectiveSearchValue = $derived(searchValue.trim().toLowerCase())
 
 	// Functions
-	const hasCurrentPage = (item: NavigationItem) => (
-		currentHref === item.href
-		|| (item.children?.some(hasCurrentPage) ?? false)
-	)
-
 	const fuzzyMatch = (text: string, query: string): [number, number][] | undefined => {
 		const ranges: [number, number][] = []
 		let textIndex = 0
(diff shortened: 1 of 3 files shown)


polymutex (changes requested):
I think this is more stripped down than what we had described, e.g. the stage ladder shouldn't be collapsed for the not-yet-passed stage, and the section in the anchor part of the URL should remain expanded. (I recommend leaving these to @darrylyeo)
There are parts we can use as-is from this PR though, e.g. the removal of the stage badges from the right sidebar and the navigation buttons, so maybe split those into their own PR

lucilapastore:
Per review feedback, split the as-is pieces into a focused PR:

https://github.com/walletbeat/walletbeat/pull/1072 — remove TOC nav buttons + hide stage badges

Leaving collapse-by-default / URL-section expand behavior for @darrylyeo. Closing this broader PR in favor of #1072.

PR #1071, opened 2026-08-13
Test: Restore url-check test with a generator script

### LINKED ISSUE #1058: Tests: URL anti hallucination test
Create test to ensure all urls in the repository exists and doesn't resolve to 404.

https://github.com/walletbeat/walletbeat/pull/1043#discussion_r3755635798

maykelxyz:
Test: Restore url-check test with a generator script

`url-check.test.ts` was deleted in 08e68096 because the inline `knownValidUrls`
list had become high-maintenance. This PR brings the test back and adds
`pnpm validate-urls` (`src/tools/url-validator/url-validator.ts`) to generate
that list automatically instead of hand-editing it. It's not fully hands-off
though: URLs that 403 or hit a Cloudflare challenge can't be auto-verified,
so those still need to be checked in a browser and added by hand..

## Changes
- `tests/utils/known-urls.ts`: `knownValidUrls` + `URLS_TO_SKIP` now live here
  instead of inline in the test, generated/kept in sync by the script rather
  than typed by hand
- `src/tools/url-validator/url-validator.ts`: walks all wallet ref URLs,
  fetches anything not already known-valid, appends successes, and drops
  entries that are no longer referenced by any wallet
- `tests/url-check.test.ts`: same checks as before (valid websites/docs/
  repos/socials/refs, exhaustive + no-extraneous + no-duplicate + valid-hash
  checks on `knownValidUrls`), just importing the list instead of declaring it
- Fixed non-existing URLs found while regenerating the list: `keycard-shell`,
  `base-app`, `elytro`, `zerion`

Closes #1058


tests/utils/known-urls.ts
```diff
@@ -0,0 +1,2285 @@
...
+	'tiktok.com',
+	'web3.bitget.com',
+	'gridplus.io',
+	'coinbase.com',
+	'sec.gov',
+	'defillama.com',
+	'defillama.com',
+	'coingecko.com',
+	'api.github.com',
+]
```
polymutex, line 32: This is quite the list now, and I worry we will miss a lot. Especially with things like `api.github.com` or `linkedin.com` (prevents checking the LinkedIn URLs of entities etc.) Also, `defillama.com` is there twice.

If we bring back this test, I think we need a more robust way to determine URL validity than single-shot direct fetch. For example, it can try to fetch it through `pure.md` as an alternative, or some URL checkers [like these](https://duckduckgo.com/?q=url%20tester%20service). The goal should be that no URL would need to be skipped.

If something is truly unfetchable, then at least this needs to be refactored to be a bit more rigorous than substring-based.

src/tools/url-validator/url-validator.ts
```diff
@@ -0,0 +1,266 @@
...
+			resolve({ ok: false, detail: err instanceof Error ? err.message : String(err) })
+		}
+	})
+}
+
+function serializeEntry(entry: KnownValidUrl): string {
+	return `\t{\n\t\turl: '${entry.url}',\n\t\turlHash: '${entry.urlHash}',\n\t\tretrieved: '${entry.retrieved}',\n\t},`
+}
+
+/** Rewrite the `knownValidUrls` array in tests/utils/known-urls.ts with the given entries. */
```
polymutex: This would be much cleaner if we just keep the known URLs set as a JSON file, not TypeScript.

src/tools/url-validator/url-validator.ts
```diff
@@ -0,0 +1,266 @@
...
+		for (const qualRef of toFullyQualified(x.ref)) {
+			for (const qualRefUrl of qualRef.urls) {
+				urls.push(qualRefUrl)
+			}
+		}
+	}
+}
+
+/** Collect every URL the URL check test would check, across all wallets. */
+function collectUrls(): Url[] {
```
polymutex, line 80: In order to verify that this URL collection is comprehensive, please add a post-build test (see `tests/postbuild`) that crawls all HTML files, looks for links to external websites, and ensures that every external URL we link to is in the known URL set.

polymutex (changes requested):

tests/utils/known-urls.ts
maykelxyz: If there are urls that can't be fetched through the script, it will encourage the user to manually verify and input the url in the `known-urls.json` file. So I was about to say that I suggest we remove urls to skip one by one, but it could take a while to ensure all is valid.

I think we can merge this for now, and file an issue as a follow up. Thoughts?

polymutex: I insist that if we keep this list, we should at least make it a real domain-name-based match (unless you already have the follow-up PR ready)

maykelxyz:
Also, updated pnpm check:all to be sequential, rather than parallel. Issue before: https://github.com/walletbeat/walletbeat/actions/runs/32406442454/job/96548130661

Issue #736, opened 2026-05-16
Import data from Coinspect: Automated data import (2/4)


polymutex:
Per #215, Coinspect is now sharing its data under a permissive license: https://github.com/coinspect/wallet-security-ranking

A lot of what Coinspect looks for in a wallet is also what Walletbeat looks for (and that's not surprising, since we both share the objective of pushing wallets on security). To that end, it would great to import Coinspect's data to enrich Walletbeat's own dataset. (Sharing data in the reverse direction is already implemented in #167).

There are four steps to making this work:

1. Creating a mapping from Coinspect wallet IDs to Walletbeat's wallet IDs (#735)
2. Importing Coinspect's raw data and ensuring automatic syncs (this issue)
3. Adding UI code to credit Coinspect (and link back to it), on attribute boxes where the data used to evaluate the attribute comes from Coinspect's dataset (#737)
4. Creating data transformation functions to import Coinspect's data into Walletbeat's own data structures (#738) (We should not implement this until the UI code for crediting Coinspect exists.)

This issue tracks (2): Importing Coinspect's raw data and ensuring automatic syncs.
This includes:
- Having a clone of the repository's `current-reports` subfolder, along with a `LICENSE` file (CC 4.0) and a `README` file pointing to the original repo. Having a clone (and not e.g. a git submodule) is important to keep Walletbeat's build process hermetic.
  - To keep the repository total weight light, we can probably skip images, though they maybe useful after #551 is implemented. Out of scope for now.
- Having a test that verifies that the clone is in sync with the remote repo.
- Having a pipeline similar to the [Helios checkpoint update pipeline](https://github.com/walletbeat/walletbeat/blob/beta/.github/workflows/helios-refresh.yaml) which automatically checks and updates the data if there is anything new. This can work by storing the latest-checked commit hash as a text file in the repo, to avoid doing a deep diff in the common case where nothing has changed.

Issue #735, opened 2026-05-16
Import data from Coinspect: Wallet ID mapping (1/4)


polymutex:
Per #215, Coinspect is now sharing its data under a permissive license: https://github.com/coinspect/wallet-security-ranking

A lot of what Coinspect looks for in a wallet is also what Walletbeat looks for (and that's not surprising, since we both share the objective of pushing wallets on security). To that end, it would great to import Coinspect's data to enrich Walletbeat's own dataset. (Sharing data in the reverse direction is already implemented in #167).

There are four steps to making this work:

1. Creating a mapping from Coinspect wallet IDs to Walletbeat's wallet IDs (this issue)
2. Importing Coinspect's raw data and ensuring automatic syncs (#736)
3. Adding UI code to credit Coinspect (and link back to it), on attribute boxes where the data used to evaluate the attribute comes from Coinspect's dataset (#737)
4. Creating data transformation functions to import Coinspect's data into Walletbeat's own data structures (#738) (We should not implement this until the UI code for crediting Coinspect exists.)

This issue tracks (1): Creating a mapping.

PR #1078, opened 2026-08-14
735 coinspect mapping

johngrantuk:
This PR maps Coinspect wallet IDs to Walletbeat wallet IDs #735. 

Marked as draft as follow up stacked PRs will be required before merging.

# Coinspect data source

Coinspect shares data in repo: https://github.com/coinspect/wallet-security-ranking

Coinspect publishes one report per wallet–platform pair under [current-reports](https://github.com/coinspect/wallet-security-ranking/tree/main/current-reports). Each report JSON includes:

- **`walletMakerUID`** — wallet-level ID (e.g. `"metamask"`, `"ambire"`)
- **`platform`** — one of `"Browser"`, `"Android"`, `"Ios"`

This PR does not clone those reports. It snapshots the `(walletMakerUID, platform)` pairs into `data/coinspect/reports-snapshot.json` so the mapping can be defined and tested now. Cloning Coinspect data into Walletbeat as the live source will be handled in #736 and I'd propose we should use `current-reports` .

## Mapping

The mapping is an object keyed by Coinspect `walletMakerUID`, each value pointing at a Walletbeat `WalletName` (the registry key in [`data/wallets.ts`](https://github.com/johngrantuk/walletbeat/blob/735-coinspect-mapping/data/wallets.ts) — e.g. `baseApp`, `uniswapWallet`).

Each entry is one of:

1. **Mapped** — `{ walletbeatId: '<WalletName>' }`  
   Example: `metamask → metamask`, `gem → gemwallet`, `coinbase-wallet → baseApp`, `bridge-wallet → mtpelerin`.
2. **Unmapped** — `{ walletbeatId: null, reason: NOT_IN_WALLETBEAT, note? }`  
   Used when Coinspect rates a wallet Walletbeat does not track yet (e.g. `zengo`, `exodus`).
3. **Mapped with skips** — same as (1), plus `skipPlatforms: [...]`  
   Used when Walletbeat only tracks a subset of the platforms Coinspect rates (e.g. Ambire skips `Android`; Base App skips `Browser`).

Separately, Coinspect `platform` strings map to Walletbeat `Variant`s via `COINSPECT_PLATFORM_TO_VARIANT`:

- `Browser` → `BROWSER`
- `Android` / `Ios` → `MOBILE`

Because Walletbeat has only one `MOBILE` variant, Coinspect’s Android and iOS reports for the same wallet both target that slot. How to merge them is deferred to #738.

The integrity test asserts every `(walletMakerUID, platform)` in the snapshot is either mapped (and the target Walletbeat wallet has that variant) or explicitly nulled, and that no two Coinspect wallets map to the same Walletbeat wallet. When #736 lands a real clone the integrity test should then read the clone the same way.

Also had some issues with a failing test and fix suggested was:
- `GitIgnoredFiles()` now matches gitignore directory patterns (`foo/`) against both the directory and its contents.
- Stops the filename-consistency test from failing on local ignored dirs such as `.pnpm-store`.


data/coinspect/mapping.ts
```diff
@@ -0,0 +1,60 @@
+import { type CoinspectMapping, CoinspectUnmappedReason } from '@/schema/data-sources/coinspect'
+
+/** Keyed by Coinspect `walletMakerUID`. */
```
polymutex: It seems like a hassle to keep a separate file for this mapping. How about putting the Coinspect ID In `WalletMetadata`?

tests/utils/codebase.ts
```diff
@@ -103,11 +103,11 @@ export async function GitIgnoredFiles(): Promise<PathPredicate> {
...
 			const normalizedDirPrefix = normalizePath(dirPrefix)
+			const dirPath = normalizedDirPrefix.endsWith('/')
+				? normalizedDirPrefix.slice(0, -1)
+				: normalizedDirPrefix
 
-			if (
-				normalizedPath === normalizedDirPrefix ||
-				normalizedPath.startsWith(normalizedDirPrefix)
-			) {
+			if (normalizedPath === dirPath || normalizedPath.startsWith(`${dirPath}/`)) {
```
polymutex, line 110: What about this PR requires this change?

src/schema/data-sources/coinspect.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Variant } from '@/schema/variants'
+
+/** How each Coinspect `platform` string resolves to a Walletbeat `Variant`. */
+export const COINSPECT_PLATFORM_TO_VARIANT = {
+	Browser: Variant.BROWSER,
+	Android: Variant.MOBILE,
+	Ios: Variant.MOBILE,
+} as const satisfies Record<string, Variant>
+
+export type CoinspectPlatform = keyof typeof COINSPECT_PLATFORM_TO_VARIANT
```
polymutex: Use a native TypeScript enum, and use the helper from `enum.ts` to get some useful functions for free.

For mapping them to variants, better to define it as a function with a `switch(coinspectPlatform)` statement, as that forces the branches to be exhaustive.

data/coinspect/reports-snapshot.json
polymutex, line 1: Are you planning to have this file contain all of the Coinspect data eventually?

If possible I think it'd be better to keep them in separate per-wallet files, similar to the data-collection JSON files. I think it would make any updates (wallet-specific or otherwise) easier to understand as well.

polymutex (changes requested):

data/coinspect/mapping.ts
johngrantuk: I felt the mapping file made it easy to see which Coinspect wallets Walletbeat does/doesn't track. 
I also thought it was useful that the test would highlight any Coinspect wallet that we didn't have an explicit mapping for (e.g. they add more).
If we want that kind of test I think we need a separate file to keep track of unmapped wallets. Maybe we don't need that test though, WDYT?

tests/utils/codebase.ts
johngrantuk: Locally check:all had started to fail for me and I couldn't see a valid reason why. This was a suggested fix from my agent.

data/coinspect/reports-snapshot.json
johngrantuk: No. My plan was to clone [current-reports](https://github.com/coinspect/wallet-security-ranking/tree/main/current-reports). The coinspect-integrity test would then be updated to check each folders report JSON file.

data/coinspect/mapping.ts
polymutex: I'm more worried about the case where we add a wallet and we forget to check if Coinspect already tested it, and end up forgetting to update this mapping file. Having a `coinspectId` field in `WalletMetadata` makes it impossible to forget this.

The test you mention is also definitely useful to have, but maybe I think we can satisfy that too by having a `knownUnmappedCoinspect` string set in `coinspect-integrity.test.ts`. Then have the test fail if either:
 - It finds a Coinspect wallet ID that isn't in `knownUnmappedCoinspect` and not in any wallet's `WalletMetadata` field. This covers the "Coinspect added a wallet" case, and forces us to think about whether this new wallet matches one we have an entry for
 - It finds that a Walletbeat wallet's `WalletMetadata` is mapped to a Coinspect ID that is in `knownUnmappedCoinspect` (ensures that the wallets we have a mapping for and those we don't are mutually exclusive)
 - It finds that `knownUnmappedCoinspect` has an ID that isn't in the current-reports set (i.e. Coinspect _removed_ a wallet)

I think this is basically describing the "separate file to keep track of unmapped wallets" idea you mention, so if so I think we're in agreement.

data/coinspect/reports-snapshot.json
polymutex: OK. In that case maybe we'll need a bit more metadata on top here, namely the commit hash and the date that this is from. May be useful to flag out-of-date/stale data

polymutex (changes requested):

data/coinspect/mapping.ts
johngrantuk: Made the change and definitely looks better and easier to maintain.

data/coinspect/reports-snapshot.json
johngrantuk: Makes sense. Thought this would be better handled in the next PR for #736

src/schema/wallet.ts
```diff
@@ -84,6 +84,11 @@ export interface WalletMetadata {
 	 */
 	id: string
 
+	/**
+	 * Coinspect wallet-maker ID for this Walletbeat wallet, if any.
+	 */
+	coinspectId?: string
```
polymutex: As with all wallet data fields, please make it non-`undefined`, i.e. it _must_ be explicitly set in all data files. Then `null` means "we don't know", and we can have a special sentinel value like `NO_COINSPECT_ID` when there is no Coinspect ID.

tests/coinspect-integrity.test.ts
```diff
@@ -0,0 +1,141 @@
...
+		walletsByCoinspectId.set(coinspectId, [wallet.metadata.id])
+	} else {
+		existing.push(wallet.metadata.id)
+	}
+}
+
+describe('coinspect mapping', () => {
+	// Coinspect added a wallet we have not tracked yet: it is neither mapped
+	// on a Walletbeat wallet nor listed in knownUnmappedCoinspect.
+	it('has a WalletBeat mapping or known-unmapped entry for every Coinspect wallet', () => {
```
polymutex: `Walletbeat` not `WalletBeat`.
If no test catches this spelling mistake, would be great to add one...

polymutex (changes requested):

src/schema/wallet.ts
polymutex: If `NO_COINSPECT_ID` collapses to `string`, then `string | null | { type: 'NO_COINSPECT_ID' }` works as a non-ambiguous type. It's a bit ugly but we use this in some other places, e.g. check the `Entity` fields in `src/schema/entity.ts`

johngrantuk: Sorry this wasn't obvious to me. I saw some other optional fields and copied those. I've made the change and also added some updates to the various docs.

polymutex (approved):
🎊

polymutex:
Can you rebase?

PR #1079, opened 2026-08-14
Feat: Create custom transaction tester

maykelxyz:
Adds a way to send your own transaction from the wallet tester. you only input a contract address and calldata, everything else reuses the existing transaction flow. (Great for testing wallet comparisons like I've been doing a lot lately)

## Changes

- Added a "Custom Transaction" item at the bottom of the Transactions sidebar in `WalletTest.svelte` (id `custom-tx`, tracked in `transactionState.hashes` like the preset tests)
- Added a custom form mode to `TransactionsTab.svelte` behind a new `isCustom` prop:
  - Contract address input + calldata textarea, validated on submit (`isValidAddress` / `isValidCalldata` type guards)
  - Builds a `TestTransaction` on the fly and sends it through the existing `handleSendTransaction`, so the mainnet chain-switch modal, pending state, and Blockscout link all work unchanged
- Form follows the same input pattern already used in `ScamAlertsTab` / `TransactionSimulationsTab` (`.address-input` / `.address-error` classes, error cleared on input) instead of inventing a new one

### DIFF
--- a/src/components/Tabs/TransactionsTab.svelte
+++ b/src/components/Tabs/TransactionsTab.svelte
@@ -3,6 +3,7 @@
 
 	let {
 		selectedTx,
+		isCustom = false,
 		transactionState,
 		account,
 		onSendTransaction,
@@ -10,6 +11,7 @@
 		formatValue,
 	}: {
 		selectedTx: TestTransaction | undefined
+		isCustom?: boolean
 		transactionState: {
 			activeId: string | null
 			isPending: boolean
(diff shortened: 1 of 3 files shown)


src/components/Tabs/TransactionsTab.svelte
```diff
@@ -21,14 +23,151 @@
...
+		isCustom
+			? transactionState.hashes[customTxId]
+			: selectedTx
+				? transactionState.hashes[selectedTx.id]
+				: undefined
+	)
+
+	function isValidAddress(addr: string): addr is `0x${string}` {
+		return /^0x[0-9a-fA-F]{40}$/.test(addr)
+	}
```
polymutex: `isErc55Address` is a more accurate version of this function (and the `Erc55Address` type is a narrower type than `0x${string}`)

polymutex (approved):

PR #1080, opened 2026-08-14
Fix: Only render name, icon, wallet URL, support status in the EIP tables

### LINKED ISSUE #1076: EIP tracker pages are huge
e.g. a pre-rendered `dist/wallet/eip/6963/index.html` is over 30 megabytes. All the other EIP trackers are in the same ballpark.

Even the index page is over 10x smaller than that despite having more content. Something's broken.

maykelxyz:
Every EIP tracker page was 31-42MB. `[eip].astro` passed full `RatedWallet` objects to the hydrated `EipSupportTable` islands, and Astro serializes all island props into the HTML — one `props` attribute alone was 31.8MB. The component also imported `getWalletUrl`.

## Changes

- Added `EipSupportRow` (`src/schema/eip-support-row.ts`):

```ts
export interface EipSupportRow {
	id: string
	displayName: string
	iconExtension: string
	url: string
	overall: EipSupportStatus
	variants: Array<{ variant: Variant; status: EipSupportStatus }>
	sourceUrls: Array<{ url: string; label: string }>
}
```

- `[eip].astro` precomputes the rows at build time; `EipSupportTable.svelte` takes `rows` instead of `wallets` and no longer imports any wallet data

Pages are now a lot smaller.

Also it turns out other tables (`WalletTable`, `Eip7702Table`, etc.) still ship the `urls` bundle.. same fix could apply there later. Though I think Darryl is working on it heavily so might be better to wait for him.

Fixes https://github.com/walletbeat/walletbeat/issues/1076


src/pages/wallet/eip/[eip].astro
```diff
@@ -29,16 +33,49 @@ type Props = { eip: Eip }
...
-function isApplicable(wallets: Array<RatedWallet<string>>): boolean {
-	return wallets.some(
-		wallet => ratedWalletEipSupport(wallet, eip.number).overall !== 'NOT_APPLICABLE',
-	)
+/**
+ * Precompute the slim table rows at build time. `EipSupportTable` is a
+ * hydrated island, and Astro serializes all of its props into the page HTML —
+ * passing full `RatedWallet` objects would inline tens of megabytes of data.
+ * Wallets the EIP does not apply to are filtered out here.
+ */
```
Comment (polymutex) [inline src/pages/wallet/eip/[eip].astro]: Temporal comment and slop

src/schema/eip-support-row.ts
```diff
@@ -0,0 +1,44 @@
...
+import { isSupported } from './features/support'
+import type { Variant } from './variants'
+
+/**
+ * Serializable per-wallet row for EIP support tracker tables.
+ *
+ * EIP tracker tables are hydrated islands (for client-side sorting), and
+ * Astro serializes all island props into the page HTML. Pages must therefore
+ * precompute these slim rows at build time instead of passing full
+ * `RatedWallet` objects, which weigh tens of megabytes once serialized.
```
polymutex: Temporal

polymutex (changes requested):

src/schema/eip-support-row.ts
polymutex, line 1: This file should not exist when `eip-support.ts` is right next to it

polymutex (changes requested):

polymutex (approved):

PR #1110, opened 2026-08-17
Render CustomContent details as Markdown on wallet pages

### LINKED ISSUE #550: Render CustomContent details as Markdown for LLM-friendly wallet pages
# Render `CustomContent` details as Markdown for LLM-friendly wallet pages

## Problem

The `index.html.md` pages (LLM-friendly Markdown wallet reviews, per the [llms.txt convention](https://llmstxt.org/)) currently cannot render attribute details that use `CustomContent` (i.e. Svelte components). When an attribute's `evaluation.details` is a component rather than typographic content, the Markdown page falls back to a link like:

```markdown
[See full details for Scam Prevention](https://walletbeat.example/metamask#scam-prevention)
```

This is better than omitting the content entirely, but it means LLMs consuming the Markdown page get no actual detail content for **9 attributes**:

- `AddressCorrelationDetails`
- `PrivateTransfersDetails`
- `ChainVerificationDetails`
- `ScamAlertDetails`
- `SecurityAuditsDetails`
- `TransactionInclusionDetails`
- `FundingDetails`
- `AccountRecoveryDetails`
- `UnratedAttribute`

### Concrete example

For MetaMask's **Scam Prevention** attribute, the HTML page renders a `ScamAlertDetails` Svelte component that produces rich content like:

> - **MetaMask** helps you stay safe when sending funds by warning you when sending funds to an address you have not sent or received funds from in the past.
> - **MetaMask** helps you stay safe when doing onchain transactions by checking the contract or transaction data against a database of known scams. However, in doing so, it leaks your IP, your Ethereum address, the contract address to an external provider which can correlate them ahead of the transaction being submitted.
> - **MetaMask** helps you stay safe when connecting to onchain apps by checking its URL against a set of known scam apps. However, in doing so, it leaks your IP and the domain name of the app to an external provider.

But the Markdown page only shows:

```markdown
### Scam Prevention: Partial

MetaMask partially implements scam prevention features.

[See full details for Scam Prevention](https://walletbeat.example/metamask#scam-prevention)
```

All of the per-feature breakdown is lost.

## Proposed solution

Pre-render each `CustomContent` component to HTML at build time, then convert the resulting HTML to Markdown using an HTML-to-Markdown library such as [`turndown`](https://github.com/mixmark-io/turndown).

At a high level:

1. In `walletPageMarkdown()` (or a helper it calls), when `evaluation.details` is a `CustomContent`, render the corresponding Svelte component to an HTML string server-side (Svelte components support SSR via `render()` from `svelte/server`).
2. Pass the HTML through `turndown` (or similar) to produce clean Markdown.
3. Use that Markdown as the details text instead of the fallback link.

### Why this approach

- **No duplication:** Each component's rendering logic lives in one place (the Svelte component). There's no need to maintain a separate Markdown variant.
- **Build-time only:** The site is entirely static, so the SSR + HTML-to-Markdown conversion happens during the build with no runtime cost.
- **Mature tooling:** `turndown` handles lists, bold, links, etc. well — exactly the kind of HTML these detail components produce.

### Considerations

- The Svelte SSR `render()` function would need the same props that `WalletPage.svelte` currently passes to each component (`wallet`, `value`, and sometimes `refs`). The `EvaluationData` type and the existing `component`/`componentProps` system should make this wirable.
- Some components render `ReferenceLinks`, which the Markdown page already handles separately. The conversion should either strip those or deduplicate.
- The `turndown` output may need some post-processing (e.g. collapsing excessive whitespace, stripping decorative HTML attributes).

### Alternative considered

Introduce Markdown-specific rendering variants for each `CustomContent` component (e.g. a `toMarkdown()` method alongside the Svelte component). This would give more control over the Markdown output but requires every component author to maintain a second rendering path, which is easy to forget and increases maintenance burden.

MattPereira:
Closes #550.

Detail components are server-rendered with `render()` from `svelte/server` and converted to Markdown with turndown, so wallet `index.html.md` pages carry their content instead of linking back to the HTML page. Build time only; the site is static.

`src/utils/details-markdown.ts` exposes one function, `renderDetailsMarkdown(details, context): string`, which owns the whole details block: typographic content is rendered as text, a detail component is server-rendered and converted, and a component with no Markdown rendering becomes a link to the HTML page. Its `switch` deliberately mirrors the dispatch in `WalletPage.svelte` rather than sharing a registry with it, since that page is hydrated `client:idle` and a shared module risks pulling turndown into the client bundle; `assertNever` makes forgetting a component a compile error.

## Implementation decisions

**1. `UnratedAttribute` is not rendered.** It is 593 of the 645 detail components, and its output restates the attribute's `shortExplanation` — printed directly above it — plus a GitHub call to action aimed at a human. It keeps today's `[See full details for X](url)` link instead, so no page loses anything relative to `beta`.

**2. Component copy changes.** Running component prose through the Markdown page put it through the Harper grammar check that already covers generated pages. These are wallet-facing, so each is listed separately — happy to drop any of them and suppress the rule instead:

- `ScamAlertDetails` concatenated the "However, in doing so, it leaks…" clause onto a bullet list, so it rendered *inside* the last bullet with no sentence break. Visible on the live site today, in MetaMask's contract warnings. It now gets its own paragraph.
- `ScamAlertDetails` concatenated `...transactions by` onto `checking the contract...`, rendering as **"bychecking"** on the site today.
- The same component built a 49-word sentence when a wallet has both new-recipient warnings and a user whitelist. Split in two.
- `AccountRecoveryDetails` and `AccountUnruggabilityDetails` said "all of the tested scenarios"; now "all the tested scenarios".
- `Code4rena` and `Immunefi` added to `.cspell.json`, plus a pattern for audit finding titles. The titles are quoted verbatim from third-party reports, so their wording is not ours to correct.

**3. Details do not repeat what the page already prints.** The page emits each attribute's `shortExplanation` above the details and a flat `#### References` list below, so `renderDetailsMarkdown` drops both when a component would duplicate them:

- `ScamAlertDetails` renders `outcome.shortExplanation` itself, and some attributes have typographic details identical to their short explanation — "Base App does not support hardware wallets." prints twice on `beta` today, across 17 wallet pages. Neither repeats on the HTML page, so the website is unaffected.
- Inline `ReferenceLinks` sections are stripped. Cost: components calling it more than once lose per-claim attribution — `ScamAlertDetails` (one per feature), `SecurityAuditsDetails` (one per audit), 15 instances. Every URL is still in the page-level list, just detached from the claim it supports.

**4. Two `no-unsafe-type-assertion` disables in `details-markdown.ts`.** `Evaluation<_Metadata>.details` is a `Content<WalletNameStrings>`, not parameterised by `_Metadata`, so nothing ties an attribute's details to its outcome type. The renderer therefore only ever sees `Outcome<OutcomeMetadata>` — i.e. `object | null` — and getting back to `ScamPreventionMetadata` and friends is an unchecked downcast. TypeScript can't help here: `switch` on the component name narrows the value, not the type parameter. `AGENTS.md` forbids `eslint-disable`, and `00-global.mdc`'s carve-out for functions "that ensure type safety from beginning to end" doesn't really cover these, since what guarantees the shape is the attribute that produced the `CustomContent`, off-module.

- MattPereira/walletbeat#1 removes both by baking `metadata` into `componentProps` at evaluation time instead, and also drops the three `metadata={outcome.metadata!}` casts in `WalletPage.svelte`. It has its own cost (metadata then exists in two places that must be kept coherent) so it is a separate PR against this branch — merge it and this PR never introduces the disables, or close it and they stand as a deliberate choice.

## Cosmetic warts in output

- `‼️ High Severity (3)⚠️ Medium Severity (2)` — adjacent badges flatten with no separator.
- `- **High**: RBY-01-001 …: Mnemonic recoverable via process dump Not fixed` — the status badge runs into the finding title.
- Heading depth is mixed (`####` and `#####` on one page), because some components emit Markdown `###` and others `<h4>`.


MattPereira:
Flagging a reviewable choice in this PR, split out so it can be accepted or rejected on its own: 
- https://github.com/MattPereira/walletbeat/pull/1

`src/utils/details-markdown.ts` only ever sees `Outcome<OutcomeMetadata>` — `object | null` — because `Evaluation<_Metadata>.details` isn't tied to `_Metadata`. Getting each component's metadata type back needs an unchecked downcast, so the module carries two `@typescript-eslint/no-unsafe-type-assertion` disables. `AGENTS.md` prohibits those outright; the `00-global.mdc` carve-out for "small functions ... that ensure type safety from beginning to end" is a stretch here, since correctness rests on the attribute that produced the `CustomContent`, off-module.

That linked PR removes both by baking `metadata` into `componentProps` at evaluation time, where the type is still known, instead of narrowing an erased outcome at render time. It also drops the three `metadata={outcome.metadata!}` casts from the `WalletPage.svelte` dispatch.

It has a real cost, which is why it's separate rather than folded in here: the metadata then exists in two places that must be kept coherent. `accountRecovery` merges metadata *after* both sub-evaluations are built, so its details have to be rebuilt from the merged value — the first attempt silently dropped the whole drills section from `AccountRecoveryDetails`, caught by the snapshot test.

Either outcome is fine by me:

- **Merge it** into this branch and this PR never introduces the disables.
- **Close it** and the disables stay as a documented, deliberate choice.

Happy to go either way — I didn't want to bundle a schema-layer change into a Markdown-rendering PR without giving you the option to say no.

tests/details-markdown.test.ts
polymutex, line 1: This is not a very useful test and looks pretty brittle. Most of the failure modes this would catch are already caught by the main build test.

src/utils/details-markdown.ts
```diff
@@ -0,0 +1,318 @@
...
+				? `${Number(parent.getAttribute('start') ?? 1) + [...parent.children].indexOf(node)}. `
+				: '- '
+		const body = content
+			.replace(/^\n+/, '')
+			.replace(/\n+$/, '\n')
+			.replace(/\n/gm, `\n${' '.repeat(prefix.length)}`)
+
+		return prefix + body + (node.nextSibling !== null && !/\n$/.test(body) ? '\n' : '')
+	},
+})
```
polymutex: Please do such initialization lazily, see `grammar.ts` for an example.

src/utils/details-markdown.ts
```diff
@@ -0,0 +1,318 @@
...
+ * that open by restating it read as a repeat. Both kinds of details do this:
+ * `ScamAlertDetails` renders `outcome.shortExplanation` itself, and some
+ * attributes give typographic details identical to their short explanation.
+ * Neither repeats on the HTML page, which does not print the short explanation
+ * separately.
+ *
+ * @param markdown The rendered details block.
+ * @param context The context it was rendered with.
+ * @returns The details block without the leading repeat.
+ */
```
polymutex: Overfitting. Seems to me like that's worth making consistent (don't make the page markdown include the short description, and instead have `renderDetailsMarkdown` always do it) rather than having this function effectively do a post-fix.

polymutex (changes requested):
Thanks, this is a great step to making the LLM pages more useful

src/utils/details-markdown.ts
MattPereira: Updated: `renderDetailsMarkdown` now owns the complete block, always emitting the short explanation before additional details. Removed caller-side emission and post-render prefix stripping; ScamAlert suppresses its internal copy for Markdown rendering.

src/utils/details-markdown.ts
MattPereira: done

tests/details-markdown.test.ts
MattPereira: Agree removed the test

PR #1122, opened 2026-08-19
Scam alerts/unlimited approval warning levels

### LINKED ISSUE #1118: Different levels for unlimited token approvals
During my test, it occurred to me that some wallets do not behave consistently when flagging unlimited token approvals. MetaMask only flags untrusted contracts (which I think is coming from Blockaid's implementation). Our current evaluation system only checks if they do support it in general.

Walletbeat should update it that it's possible to know if a wallet only flags on certain scenarios (untrusted contracts, untrusted domains, etc.). The other question is: How should we rate it?

maykelxyz:
Attribute: Add warning levels for unlimited token approval. As per my recent wallet comparison, We've realized that a number of wallets only warns on certain situations: blacklisted domains, unknown contracts, etc. Our data should support that level of complexity regarding unlimited token approval warnings.

## Changes
- `warnsOnUnlimitedApproval` is now `'ALWAYS' | NonEmptyArray<UnlimitedApprovalWarningCondition>` instead of `boolean`
```ts
export enum UnlimitedApprovalWarningCondition {
	EOA = 'EOA',
	UNKNOWN_CONTRACTS = 'UNKNOWN_CONTRACTS',
	BLACKLISTED_CONTRACTS = 'BLACKLISTED_CONTRACTS',
	NEW_CONTRACTS = 'NEW_CONTRACTS',
	CONTRACTS_NOT_INTERACTED_BEFORE = 'CONTRACTS_NOT_INTERACTED_BEFORE',
}
```
- `rateUnlimitedApprovalWarning` now tracks a `conditionalOnly` flag on `ScamAlertSupport`
- `evaluateScamAlerts` returns a new `PARTIAL` outcome (`conditional_unlimited_approval_warning`) when the wallet only warns in specific scenarios rather than unconditionally, with a `howToImprove` explaining why trusted spenders aren't necessarily safe
- Added an `exampleRating` covering the conditional-warning case
- `ScamAlertDetails.svelte` now renders a "but only when X, not unconditionally" clause when the warning is conditional
- Updated `completed.tmpl.ts` and `features.md` for the new type

## Other changes
- Added `blocklist` to cspell

Closes #1118


polymutex (commented):
I fear this may be difficult to test for. What set of conditions would we look for?
Perhaps it is worth enumerating all these conditions, and having individual tests for each?

src/schema/features/security/scam-alerts.ts
```diff
@@ -90,13 +90,82 @@ export type SendTransactionWarning = WithRef<
...
+ */
+export type UnlimitedApprovalWarningBenchmarks = Record<
+	UnlimitedApprovalWarningBenchmarkSpenders,
+	Support | null
+>
+
+/**
+ * Shorthand for a wallet that warns on every unlimited-approval benchmark.
+ */
+export const allUnlimitedApprovalBenchmarksSupported: UnlimitedApprovalWarningBenchmarks = {
```
polymutex: I don't think we should have this shorthand, for the same reason we don't do this for transaction legibility: If we add a benchmark case, then we would implicitly assume every wallet where we previously measured this would also pass the new benchmark, which is not a safe assumption.

It's OK to have such helpers for:
- The all-negative case (e.g. `noDataExtraction`), since it's safe to assume set expansion means every negative-previously-benchmarked wallet would continue being negative on everything
- The positive case, for sets of things that we can be confident will never expand (e.g. `displaysFullCallData`), since the Ethereum transaction ABI is very stable at this point
- The positive case, for sets of things where it *is* safe to assume that expansion would automatically apply to previously-benchmarked wallets, because they implement a common standard where expanding the benchmark set itself would follow the expansion of the standard (e.g. `erc8213SoftwareWalletFullySupported`).

But neither of these things apply here.

It is OK to define this helper in the *attribute* file itself though, as a non-`export`ed `const`, because that is useful to use in the `ratingScale` part of the attribute definition. And since it's not `export`ed, we can be sure that no wallet data file can use it.

polymutex (changes requested):

src/schema/features/security/scam-alerts.ts
maykelxyz: Okay yeah that makes sense, updated to a non-`export` const in the attribute file

PR #1123, opened 2026-08-20
Zerion: Fill in app isolation for both variants

ren2140eth:
Fills `privacy.appIsolation` 

Browser extension values come from source permalinks; mobile values come from screenshots taken in its in-app browser. The two variants agree on all four fields.


data/software-wallets/zerion.ts
```diff
@@ -220,7 +224,138 @@ export const zerion: SoftwareWallet = {
...
+								explanation:
+									'The account selector in the connection dialog is rendered from the list of accounts that already exist.',
+								label: 'Connection dialog account selector',
+								url: 'https://github.com/zeriontech/zerion-wallet-extension/blob/482c0a5f57cee79b618147c804a92a98240c559a/src/ui/pages/RequestAccounts/RequestAccounts.tsx#L145-L152',
+							},
+							{
+								explanation:
+									'When none of the existing accounts match the requested ecosystem, the dialog tells the user to leave it and create an account under Settings.',
+								label: 'Connection dialog empty state',
+								url: 'https://github.com/zeriontech/zerion-wallet-extension/blob/482c0a5f57cee79b618147c804a92a98240c559a/src/ui/pages/RequestAccounts/RequestAccounts.tsx#L459-L466',
```
polymutex: This isn't quite what the feature being rated is about. This is about the wallet's behavior when connecting to an app it hasn't connected to before, not a different chain.

data/software-wallets/zerion.ts
```diff
@@ -220,7 +224,138 @@ export const zerion: SoftwareWallet = {
...
+							},
+						],
+					}),
+					erc7846WalletConnect: notSupportedWithRef({
+						ref: [
+							{
+								explanation:
+									'An incoming request is served only when the public controller defines a method of the same name, and that controller defines no `wallet_connect` method.',
```
polymutex: Overly confusing explanation that only makes sense when looking at the code, not in the context of the page the `ref` is displayed on. The explanation can just say that `wallet_connect` isn't defined in Zerion's wallet/app integration connector code.

data/software-wallets/zerion.ts
```diff
@@ -220,7 +224,138 @@ export const zerion: SoftwareWallet = {
...
+								file: 'public/references/wallets/zerion/screenshots/2026-08-19-zerion-app-isolation-account-picker.png',
+								label: 'Zerion mobile connection sheet with the account list expanded',
+								lastRetrieved: '2026-08-19',
+							},
+						],
+					}),
+					erc7846WalletConnect: notSupportedWithRef({
+						ref: {
+							explanation:
+								'Calling `wallet_connect` from the Walletbeat test page in the Zerion in-app browser returns "the method wallet_connect does not exist/is not available".',
```
polymutex: (This is a better explanation that the one above for the `browser` version of this field)

data/software-wallets/zerion.ts
```diff
@@ -220,7 +224,138 @@ export const zerion: SoftwareWallet = {
...
+							},
+							{
+								explanation:
+									'After disconnecting and reconnecting to the same site, the connection sheet offers the first account rather than the account that site had been connected with.',
+								file: 'public/references/wallets/zerion/screenshots/2026-08-19-zerion-app-isolation-uniswap-reconnect-default.png',
+								label:
+									'Zerion mobile connection sheet for app.uniswap.org offering the first account',
+								lastRetrieved: '2026-08-19',
+							},
+						],
```
polymutex, line 344: Great proof here 👍

polymutex (changes requested):

data/software-wallets/zerion.ts
ren2140eth: this is gone in the latest went with a screenshot ref instead here and a simple explanation

data/software-wallets/zerion.ts
ren2140eth: yep cleaner, updated

polymutex (approved):

PR #1136, opened 2026-08-22
MetaMask: Fill in reproducible builds by variant

ren2140eth:
Fills MetaMask's `transparency.releaseTransparency.reproducibleBuilds` as `notSupported`.

MetaMask does reproduce its Firefox builds — the release workflow rebuilds each published artifact and compares the build trees by SHA-256 digest, and a source package goes to Mozilla's reviewers — but neither is something an independent party can run. The comparison harness is a private repository, the reviewer package goes to a private S3 bucket, and the public build recipe derives `SOURCE_DATE_EPOCH` from a checkout timestamp while the release build falls back to the latest commit timestamp, so a clone-and-build is not byte-identical.

`releaseProcess` stays PARTIAL.

Also adds `ren2140` as a contributor and updates `lastUpdated`.

### DIFF
--- a/data/software-wallets/metamask.ts
+++ b/data/software-wallets/metamask.ts
@@ -87,7 +87,7 @@ export const metamask: SoftwareWallet = {
 		tableName: 'MetaMask',
 		contributors: [polymutex, nconsigny, mattmatt, ren2140],
 		iconExtension: 'svg',
-		lastUpdated: '2026-08-21',
+		lastUpdated: '2026-08-22',
 		urls: {
 			androidManifestXml:
 				'https://raw.githubusercontent.com/MetaMask/metamask-mobile/main/android/app/src/main/AndroidManifest.xml',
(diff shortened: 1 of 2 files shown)


data/software-wallets/metamask.ts
```diff
@@ -798,7 +799,49 @@ export const metamask: SoftwareWallet = {
...
+									'Release zip entry modification times are resolved from `SOURCE_DATE_EPOCH` when set and from the latest commit timestamp otherwise, so the archive does not vary with the time of the build.',
+								url: 'https://github.com/MetaMask/metamask-extension/blob/36fd2b1e50ae2ee5b4de2b146929c61d07ab2f80/development/webpack/utils/plugins/ManifestPlugin/zip-mtime.ts#L43-L53',
+							},
+							{
+								explanation:
+									'Each release publishes the built zips alongside a `SHA256SUMS` file, giving an independent party a published digest to compare a rebuild against.',
+								url: 'https://github.com/MetaMask/metamask-extension/releases/tag/v13.45.0',
+							},
+						],
+					}),
```
polymutex: This is promising but the attribute specifically requires that someone can _independently_ reproduce the build, i.e. you should be able to clone the MetaMask repository and build it yourself and verify that this matches the version you can download on the Chrome/Firefox extension site.

polymutex (changes requested):

data/software-wallets/metamask.ts
ren2140eth: Yep updated so now this is no longer null. Might make sense to refine TSDoc to specifically state the independence piece

polymutex (approved):
Yep, feel free to open another PR to make this clear in the feature field documentation.

PR #1141, opened 2026-08-22
Rabby: Assess mobile app security best practices

ren2140eth:
## Summary

Rabby declares a mobile variant, but `securityBestPractices.mobile` was `'NOT_A_MOBILE_APP'` 

The first commit adds two manifest entries the parser did not model (`DETECT_SCREEN_CAPTURE`, `NSUserNotificationsUsageDescription`), both rated as passes. Split out because it affects every wallet.

RNG — `@scure/bip39` drawing from `crypto.getRandomValues`, installed at startup by `react-native-quick-crypto`: [generation](https://github.com/RabbyHub/rabby-mobile/blob/20a6d0af7c459691084aa470e04f09432f0ce1c7/packages/service-keyring/src/keyringService.ts#L2587-L2589), [provider](https://github.com/RabbyHub/rabby-mobile/blob/20a6d0af7c459691084aa470e04f09432f0ce1c7/apps/mobile/global.ts). The repo's [`@scure/bip39` patch](https://github.com/RabbyHub/rabby-mobile/blob/20a6d0af7c459691084aa470e04f09432f0ce1c7/.yarn/patches/%40scure-bip39-npm-1.3.0-1d74c5c469.patch) changes only seed derivation, not `generateMnemonic`.

## Open question

5,000 PBKDF2 iterations is low — Rabby's own extension does 900,000. Recorded as `ENCRYPTED_WITH_USER_SECRET_STANDARDIZED_KDF`, since the schema defines `WEAK_KDF` as "non-standard or ad-hoc" and PBKDF2 is standardized. Should weak parameters should be penalized?

## Notes

- The collector also refreshed the extension manifest, 0.93.83 → 0.94.4. The only rated change is `externally_connectable` now declared with an empty `ids` list instead of omitted. Does it make sense to update these periodically?


data/software-wallets/rabby.ts
```diff
@@ -681,7 +692,47 @@ export const rabby: SoftwareWallet = {
 					secureRng: SecureRngSource.OS_CSPRNG,
 				},
 				desktop: 'NOT_A_DESKTOP_APP',
-				mobile: 'NOT_A_MOBILE_APP',
+				mobile: {
+					ref: [
+						{
+							explanation:
+								'The mobile app builds its keyring service with its own React Native encryption module rather than the `@metamask/browser-passworder` used by the extension. It derives the vault key from the user password with PBKDF2-HMAC-SHA256 at 5,000 iterations to a 256-bit key, and encrypts the vault with AES-256-CBC, both via `react-native-aes-crypto`.',
```
ren2140eth: refine only need to talk about the mobile version here i think this can be shortened to The mobile app builds its key ring service with its own react native encryption module. It derives the vault key from the user password with PBKDF2-HMAC-SHA256 at 5,000 iterations to a 256-bit key, and encrypts the vault with AES-256-CBC.

data/software-wallets/rabby.ts
```diff
@@ -681,7 +692,47 @@ export const rabby: SoftwareWallet = {
...
+								},
+								{
+									label: 'PBKDF2 parameters and AES mode',
+									url: 'https://github.com/RabbyHub/rabby-mobile/blob/20a6d0af7c459691084aa470e04f09432f0ce1c7/apps/mobile/src/core/services/encryptor.ts#L12-L23',
+								},
+							],
+						},
+						{
+							explanation:
+								'Recovery phrases are generated on device by `@scure/bip39`, which draws entropy from `@noble/hashes` `randomBytes` and therefore from `crypto.getRandomValues`. The app installs that provider at startup via `react-native-quick-crypto`, which is backed by the platform CSPRNG. The repository patches `@scure/bip39` 1.3.0, but the patch only replaces the seed-derivation PBKDF2 with a native implementation and adds `Uint8Array` mnemonic handling; it does not alter `generateMnemonic` or its entropy source.',
```
ren2140eth: shorten to 'Recovery phrases are generated on device by `@scure/bip39` and then what the entropy source is

data/software-wallets/rabby.ts
ren2140eth: Agreed — the extension comparison and `react-native-aes-crypto` are both out of scope for a mobile ref. Trimmed to your wording in 5a6461e1.

data/software-wallets/rabby.ts
ren2140eth: Agreed, trimmed to generation plus entropy source in 5a6461e1. I moved the `@scure/bip39` patch note into a code comment rather than dropping it — it records why the entropy claim still holds despite the repo patching bip39, which is a maintenance detail rather than something a visitor needs. Also removed the patch link from the rendered refs, since it no longer supports anything in the prose.

data/software-wallets/rabby.ts
```diff
@@ -681,7 +692,51 @@ export const rabby: SoftwareWallet = {
...
+						// (.yarn/patches/@scure-bip39-npm-1.3.0-1d74c5c469.patch). It leaves
+						// `generateMnemonic` and its entropy source untouched, so the reference
+						// below holds. It does replace seed derivation (phrase to seed) with a
+						// native implementation, at BIP-39's parameters: PBKDF2-SHA512, 2048
+						// iterations, 64 bytes. It also drops upstream's NFKD normalization of the
+						// phrase, which is inert here because Rabby both generates and validates
+						// phrases against the English wordlist only, and those words are ASCII.
+						{
+							explanation:
+								'Rabby creates your recovery phrase on your own device rather than on a server, and takes the randomness it needs from the operating system rather than from its own code. It generates the phrase with `@scure/bip39`, drawing on `crypto.getRandomValues` as provided by `react-native-quick-crypto`.',
```
ren2140eth: kill the rather than pieces.

data/software-wallets/rabby.ts
```diff
@@ -681,7 +692,51 @@ export const rabby: SoftwareWallet = {
 					secureRng: SecureRngSource.OS_CSPRNG,
 				},
 				desktop: 'NOT_A_DESKTOP_APP',
-				mobile: 'NOT_A_MOBILE_APP',
+				mobile: {
+					ref: [
+						{
+							explanation:
+								'Rabby keeps your recovery phrase on the device, encrypted with a key derived from your app password, so a copy of the stored file cannot be read without that password. The number of derivation rounds sets how slow each password guess is: the mobile app uses PBKDF2-HMAC-SHA256 at 5,000 rounds to a 256-bit key, and encrypts with AES-256-CBC.',
```
ren2140eth: kill so a copy of the stored.... and then kill "the number of derivation rounds sets how slow each password guess is"

ren2140eth: Both cut in 861b88ed. Agreed on the round-count gloss especially — telling the reader what to make of 5,000 is the rating logic's job, not the reference's. Now reads: "Rabby keeps your recovery phrase on the device, encrypted with a key derived from your app password. The mobile app uses PBKDF2-HMAC-SHA256 at 5,000 rounds to a 256-bit key, and encrypts with AES-256-CBC."

data/software-wallets/rabby.ts
ren2140eth: Done in 861b88ed. Defining it by what it is not was the weaker construction anyway. Now reads: "Rabby creates your recovery phrase on your own device, and takes the randomness it needs from the operating system. It generates the phrase with `@scure/bip39`, drawing on `crypto.getRandomValues` as provided by `react-native-quick-crypto`."

data/software-wallets/rabby.ts
```diff
@@ -681,7 +692,73 @@ export const rabby: SoftwareWallet = {
...
+								},
+								{
+									label: '`@noble/hashes` 1.4.0 sources that entropy from `crypto.getRandomValues`',
+									url: 'https://github.com/paulmillr/noble-hashes/blob/531daab72e8cef0dbaf2db134260c758a89a39ed/src/utils.ts#L249-L256',
+								},
+								{
+									label: 'Crypto provider installed at app startup',
+									url: 'https://github.com/RabbyHub/rabby-mobile/blob/20a6d0af7c459691084aa470e04f09432f0ce1c7/apps/mobile/global.ts',
+								},
+							],
```
polymutex, line 755: Excellent find here.

polymutex (approved):
Great PR

PR #1143, opened 2026-08-22
Rainbow: Correct and source the app unlock mechanisms

ren2140eth:
`security.duressResistance` recorded Rainbow as password-only with `ref: refTodo`, which is the browser extension's mechanism, not the mobile app's. 

Split per variant, each with source references:

- **Mobile** — biometric and PIN. `isAuthenticated()` gates the app on device authentication; `getPrivateAccessControlOptions()` sets `ACCESS_CONTROL.USER_PRESENCE` on iOS and `BIOMETRY_CURRENT_SET_OR_DEVICE_PASSCODE` on Android; `shouldAuthenticateWithPIN()` routes to the app's own PIN screen on Android devices with no passcode set.
- **Browser** — password, unchanged in value, now referenced to the extension's unlock screen.

Separately this attribute exempts browser/desktop/embedded wallets. However, there are many wrench attacks that are home invasions in that case I think all of those wallets would be in play

### DIFF
--- a/data/software-wallets/rainbow.ts
+++ b/data/software-wallets/rainbow.ts
@@ -82,7 +82,7 @@ export const rainbow: SoftwareWallet = {
 		tableName: 'Rainbow',
 		contributors: [polymutex, mattmatt, ren2140],
 		iconExtension: 'svg',
-		lastUpdated: '2026-08-20',
+		lastUpdated: '2026-08-22',
 		urls: {
 			androidManifestXml:
 				'https://raw.githubusercontent.com/rainbow-me/rainbow/develop/android/app/src/main/AndroidManifest.xml',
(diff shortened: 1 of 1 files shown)


data/software-wallets/rainbow.ts
```diff
@@ -733,18 +733,55 @@ export const rainbow: SoftwareWallet = {
...
+							},
+							{
+								explanation:
+									'That authentication is Face ID or Touch ID on iOS and fingerprint or face unlock on Android, with the phone passcode accepted in their place.',
+								label: 'Biometric unlock settings in the mobile app source code',
+								url: 'https://github.com/rainbow-me/rainbow/blob/e3df13be2e139357770c4dd20573fc96836bd1ee/src/features/local-auth/keychain.ts#L399-L414',
+							},
+							{
+								explanation:
+									'On an Android phone with no screen lock set, the app asks for a PIN of its own.',
```
ren2140eth: drop "of its own"

ren2140eth: Done in 9d0d13ff.

polymutex (approved):

PR #1145, opened 2026-08-24
Extract attribute details from Svelte into shared model

### LINKED ISSUE #550: Render CustomContent details as Markdown for LLM-friendly wallet pages

MattPereira:
Fixes #550 

Rather than converting Svelte-rendered HTML into Markdown, this PR moves attribute-detail meaning into typed, format-neutral models and renders those models through dedicated web, Markdown, and JSON adapters.

 ```
    Wallet features → Evaluation → Structured details
                                          ├→ Web (Svelte/HTML)
                                          ├→ Markdown
                                          └→ JSON
 ```

### Summary
- Wallet Markdown pages now include full structured attribute details instead of fallback links.
- Keeps the `UnratedAttribute` web-only because it’s a UI action, not evaluation data.
- Replaces `CustomContent` and `WalletPage`’s hard-coded component switch with format-neutral structured evaluation details.
- Adds typed structuredDetails to JSON exports and the public schema
- Keeps references attached to their claims while deduplicating page-level reference lists.
- Fixes pre-existing detail errors exposed by migration: 
  - incorrect aggregation, duplicate entries, lost references, and timezone-shifted audit dates.


tests/wallet-json-export.test.ts
MattPereira, line 1: The tests are brittle just temporary to help prove spike works. Will remove

src/types/content/scam-alert-details.ts
```diff
@@ -1,17 +1,41 @@
...
+	return componentAndProps.component === 'ScamAlertDetails'
+}
+
+/** Extract the stored model without recovering it from erased outcome metadata. */
+export function extractScamPreventionDetails<S extends Strings>(
+	content: Content<S>,
+): ScamPreventionDetails | null {
+	if (!isCustomContent(content) || !isScamAlertDetailsContent(content.component)) {
+		return null
+	}
```
polymutex: Under what circumstances would this ever happen?
If the answer is "never", then this should throw, not paper over the problem.

src/utils/attribute-details-markdown.ts
```diff
@@ -0,0 +1,25 @@
+import type { ScamPreventionDetails } from '@/schema/attributes/security/scam-prevention'
+import { renderStrings } from '@/types/utils/text'
+
+/**
+ * Markdown adapters for structured attribute-detail models.
+ * Add attribute-specific renderers here as more custom detail views adopt shared models.
+ */
+
+/** Markdown adapter for the canonical scam-prevention detail model. */
+export function renderScamPreventionDetailsMarkdown(
```
polymutex: If we're going to have a function like this for each type of custom content, I think it may be warranted to use a text templating engine. Something like jinja or something, maybe there's a more canonical template engine for TypeScript-ish projects?

polymutex (changes requested):
I like this better than #1110, seems like it would have better flexibility to ensure each type can be rendered as optimally as possible in all 3 contexts.
Are you planning to drop #1110?

src/types/content/scam-alert-details.ts
MattPereira: just a scaffolding quirk of the spike implementation for scam-prevention attribute only

i'll ping you for review once the proper adapter registry is setup

src/utils/attribute-details-markdown.ts
MattPereira: Addressed in 50eaa5a4.

Instead of a template engine, each detail type now returns a small list of layout pieces: `heading, paragraph, list, references`

One function turns those into Markdown. Bullet markers, indents and blank lines are decided in that one place instead of in nine renderers.

Skipped a template engine because a text template only helps Markdown; the Svelte views and JSON export would still be hand-written. These pieces serve all three, and TypeScript checks them.

src/types/content/inline.ts
polymutex, line 1: This is an interesting idea, but I think it is not expressive enough to cover all the cases we'd need to cover. Specifically it likely needs bulleted list support, quotation support, multi-line codeblocks, paragraphs, and forced newlines. At least that's what I am fairly confident we already use for markdown. I see some of this is in `src/types/content/detail-block.ts` but they seem non-interoperable.

But even if that wasn't the case, this still feels quite redundant with markdown itself. Maybe it is time to use a real markdown parser and have real Markdown AST types? Consider the utility functions we already have in `src/utils/markdown-utils.ts` which could also benefit from this.

I understand Markdown is just one aspect to "render" to out of three, but it is also one that sufficiently captures what these types are already aiming to capture. Perhaps if using real Markdown AST types, at least we could use a strict subset of it (that we know how to render to HTML and to JSON), and enforce that this type of typed text doesn't use anything beyond that subset.

Thoughts? (Please don't update the whole PR in response to this, we should try to nail this down before any huge further changes.)

polymutex (changes requested):

src/types/content/inline.ts
MattPereira: > Please don't update the whole PR in response to this...

ya my bad shoulda kept to vert slice of single attribute. lesson learned :melting_face:

MattPereira: i agree `detail-block.ts` is incomplete solution and that real markdown parser + AST types would be better

another problem with this PR is the [weird string exports](https://github.com/MattPereira/walletbeat/blob/50eaa5a475c0fbdf8df86a480de9656c0024440b/src/types/content/account-recovery-details.ts#L90-L114) from `types/content/*-details.ts` files which seem ripe for drift as Svelte and Markdown consumers both reconstruct ordering, conditions, headings, lists, etc.

maybe something like this could be possible?

```
  Evaluation
    ↓
  AccountRecoveryDetails       ← facts only
    ├─ JSON DTO
    └─ present(details, context)
            ↓
       restricted MDAST        ← wording + document structure, once
         ├─ Markdown serializer
         └─ generic web renderer
```

<details><summary><b>See codex elaboration</b></summary>

image

</details>

But then the generic svelte renderer would require exception for at least `SecurityAuditsDetails.svelte`

<details><summary><b>See feasability exploration<b></summary>

image


</details>

MattPereira: After more thought, perhaps #550 proposal is best approach after all because it solves the drift issue between svelte and markdown without large refactor

My lean is to close this pr and re-open #1110 

> Pre-render each CustomContent component to HTML at build time, then convert the resulting HTML to Markdown using an HTML-to-Markdown library such as [turndown](https://github.com/mixmark-io/turndown).

MattPereira: @polymutex wdyt?

src/utils/wallet-json-export.ts
```diff
@@ -37,8 +33,16 @@ import {
...
+	type ReferenceJsonExport,
+	serializeReferences,
+	serializeStructuredDetails,
+	type StructuredDetailsJsonExport,
+} from '@/utils/structured-details/json'
+import { renderStructuredDetailsMarkdown } from '@/utils/structured-details/markdown'
 
-const DETAILS_FALLBACK = 'See full details on the wallet page.'
+/** Attributes without details (e.g. unrated ones) export an empty legacy `details` string. */
+const DETAILS_FALLBACK = ''
```
polymutex: Seems this should be replaced now.

And if an attribute is unrated, it should say something to that effect, not nothing.

PR #1147, opened 2026-08-24
Prevent redundant Check workflow runs

MattPereira:
## Issue

- Pushes to open PR branches ran CI twice: once for `push`, once for `pull_request.synchronize`.
- PR metadata and status changes (`edited`, `ready_for_review`, and `review_requested`) reran CI without changing the code under test.
- Superseded PR runs continued consuming CI resources.

Example runs triggered by editing PR title / body:
ci-redundant-runs

## Suggested Changes

- Run `push` checks only for `beta`; PR branches remain covered by `pull_request`.
- Trigger PR checks only for `opened`, `reopened`, and `synchronize`.
- Cancel superseded runs for the same PR.
- Give each `beta` commit its own concurrency group so every trunk commit receives a result.

Removing metadata and status triggers aligns CI with its purpose: run checks when the tested state changes, not when PR metadata or review status changes. Rebases, force-pushes, and new commits still trigger `synchronize`.

## Caveat: base-branch retargeting

- Manually changing the PR base branch no longer triggers CI automatically.
- GitHub exposes base changes through the broad `edited` event, which also fires for title and body edits. Supporting only base changes would require extra job filtering plus special concurrency and check-name handling so metadata edits cannot cancel or mask real checks.
- If  changing the PR’s target branch, close and reopen the PR to run CI against the new target.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -10,35 +10,20 @@ on:
     types:
       - opened
       - reopened
-      - edited # Filtered down to base-branch changes; see `check-quick`.
       - synchronize
       - ready_for_review
       - review_requested
 
 # Supersede in-flight runs for the same PR (or the same branch, for pushes),
-# but never cancel a run on the trunk. Runs that `check-quick` is going to
-# skip are placed in a separate group, so that they cannot evict a run that
-# is doing real work and leave only a skipped -- and therefore passing --
-# result behind.
+# but never cancel a run on the trunk.
 concurrency:
-  group: >-
-    check-${{ github.event.pull_request.number || github.ref }}-${{
-    (github.event.action == 'edited' && github.event.changes.base == null)
-    && 'noop' || 'work' }}
+  group: check-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


MattPereira:
> At 8 matrix jobs plus check-full each, that is ~54 jobs queued for zero information.

Damn claude is so rude / sassy these days might have to fully switch to codex :sweat_smile:

.github/workflows/check.yaml
```diff
@@ -1,15 +1,22 @@
...
+    branches:
+      - beta
   pull_request:
     types:
       - opened
       - reopened
-      - edited
       - synchronize
-      - ready_for_review
-      - review_requested
```
polymutex, line 12: Any reason to not keep these?

MattPereira: These events trigger the CI checks without the underlying code changing:

- `edited` fires for each title and body edit of a PR
- `ready_for_review` fires when PR flipped out of draft mode
- `review_requested` fires on review request

If the commit hasn’t changed, re-running produces the same result while:
- wasting CI time and resources
- delaying checks for newer commits
- cluttering the Actions tab of the repo with duplicate workflow runs

polymutex (approved):

Issue #1115, opened 2026-08-18
Enforce stronger lints and tests on `.svelte` files


polymutex:
> @polymutex `svelte-check` would have caught it. Nothing in the toolchain typechecks `.svelte`
> files today:
> 
> 1. `@astrojs/check` only diagnoses `.astro`/`.ts`/`.js` — it never opens `.svelte`, and reports 0
>    errors on `beta` with all four broken reads present. The same typo in a `.ts` file is caught:
>    `ts(2551): Property 'repository' does not exist on type 'WalletUrls'. Did you mean 'repositories'?`
> 2. typescript-eslint via `svelte-eslint-parser` doesn't cover it — `eslint.config.mjs` `ignores`
>    lists `WalletPage.svelte`, `WalletTable.svelte`, `Filters.svelte` and `Table.svelte`. In the
>    files it does parse, `check:lint` emits `Unsafe member access .metadata on a type that cannot be
>    resolved`, so `wallet` is an error type and the typo wouldn't flag there anyway.
> 
> With the fix reverted, `npx svelte-check@4 --tsconfig ./tsconfig.json` reports that ts2551 on all
> four occurrences.
> 
> The catch: it also reports 93 errors and 28 warnings across 29 files on `beta`, all pre-existing
> and unrelated (Snippet type variance, `undefined` not assignable to `ladders`). So gating CI on it
> means fixing those first. Cheapest option first: gate on `TS2339`/`TS2551` only, the class of bug
> here (haven't measured that subset); land it non-blocking with a baseline file and tighten over
> time; or fix all 93 and gate on the full check.
>  

 _Originally posted by @MattPereira in [#906](https://github.com/walletbeat/walletbeat/issues/906#issuecomment-5311080067)_

cc @darrylyeo

PR #1148, opened 2026-08-24
Add initial Svelte type checking

MattPereira:
Partial progress on #1115

## Summary

- Add `svelte-check` pkg and `check:svelte` script
  - `92 errors and 28 warnings in 29 files`
- Fix generated declarations shadowing `svelte/elements`
  - `65 errors and 28 warnings in 28 files`
- Fix additional diagnostics in `WalletPage.svelte` and `WalletTable.svelte`
  - `58 errors and 28 warnings in 28 files`

## Follow up issue

- #1178


polymutex (approved):
Thanks, this was overdue :)

src/styles/css-attributes.d.ts
```diff
@@ -1,1160 +1,1164 @@
...
-declare namespace astroHTML.JSX {
-	interface HTMLAttributes extends CssAttributes {}
+	declare namespace astroHTML.JSX {
+		interface HTMLAttributes extends CssAttributes {}
+	}
 }
 
 declare module 'svelte/elements' {
-	export interface HTMLAttributes<_T extends EventTarget> extends CssAttributes {}
+	export interface HTMLAttributes<T extends EventTarget> extends CssAttributes {}
```
polymutex, line 1164: `'T' is defined but never used. Allowed unused vars must match /^_/u`

polymutex (changes requested):

src/styles/css-attributes.d.ts
MattPereira: the change to `T` was to fix a bunch of `svelte-check` errors because

> TypeScript only merges interface declarations when their type parameter names match verbatim, so _T silently drops the augmentation and `CssAttributes` never reaches Svelte's `HTMLAttributes` so every `data-*` attribute becomes unknown.

MattPereira: lmk if okay with  `// esling-disable-next-line` or prefer some other solution https://github.com/walletbeat/walletbeat/pull/1148/commits/7e2b7f34d0066f9b420a2c039c88909a89fbcd4e

svelte-check cannot be made to accept `_T` because:


> 1. TS2428 is a compiler invariant, not a svelte-check lint. svelte-check just forwards tsc diagnostics — there's no flag, and no // @ts-expect-error placement helps, since the damage is the mangled 2-param signature, not the diagnostic.
> 2. Every augmentation target Svelte exposes already uses T:
>   - svelte/elements.d.ts:751 — HTMLAttributes<T extends EventTarget>
>   - svelte2tsx/svelte-jsx-v4.d.ts:31 — interface HTMLAttributes<T extends EventTarget = any> {} (the empty interface explicitly provided for augmentation — still T)
>3. Avoiding the merge isn't viable — the only alternative is augmenting SvelteHTMLElements entry-by-entry, ~130 tags enumerated by hand.

MattPereira: Fixed via a file-specific ESLint exception. `T` must match Svelte’s `HTMLAttributes<T>` declaration.

Going back to `_T` causes `TS2428` which results in 100+ `svelte-check` errors

eslint.config.mjs
```diff
@@ -254,4 +254,19 @@ export default [
 			'@typescript-eslint/no-unsafe-member-access': 'warn',
 		},
 	},
+	{
+		files: ['src/styles/css-attributes.d.ts'],
+		rules: {
+			// Must match Svelte's merged `HTMLAttributes<T>` declaration.
+			'unused-imports/no-unused-vars': [
```
polymutex: Does a `eslint-disable-next-line` rule in the file itself not work?

polymutex (changes requested):

eslint.config.mjs
MattPereira: i first tried to put the disable comment in `css-attributes.d.ts` but CI failed cus that is a generated file

but looks like we can actually put the disable in `src/styles/generate/css-attributes-codegen-lib.ts`

MattPereira: also claude and codex both think there is dead code that could be removed
- https://github.com/walletbeat/walletbeat/pull/1182

polymutex (approved):

PR #1149, opened 2026-08-25
Coinspect import

### LINKED ISSUE #736: Import data from Coinspect: Automated data import (2/4)

johngrantuk:
Closes #736 

Vendors Coinspect’s public `current-reports/` tree into `data/coinspect/current-reports/` and adds a weekly bot pipeline (modelled on Helios checkpoint refresh) that keeps that snapshot in lockstep with [coinspect/wallet-security-ranking](https://github.com/coinspect/wallet-security-ranking) `main`.

- `deploy/coinspect/coinspect-update.sh` — `ls-remote` fast-path, sparse shallow clone, rsync (dropping `images/` and `images.json`), then write `data/coinspect/upstream-commit`.
- `deploy/coinspect/coinspect-push.sh` — porcelain gate over `data/coinspect/`, signed commit, push to `beta`.
- `.github/workflows/coinspect-refresh.yaml` — Sunday 00:00 UTC + `workflow_dispatch`. Reuses `deploy/setup-git.sh` and the existing `walletbeat-bot` SSH secrets.
- initial `current-reports/` tree and `upstream-commit` pin are committed here so the first bot run is not a huge unreviewed import.
- Added LICENSE (matching Coinspect) and README 
- Tests: mapping checks now read `current-reports/` directly (`reports-snapshot.json` removed). A new suite asserts `upstream-commit` is a full SHA-1 and matches upstream `main`.

Local refresh: `pnpm coinspect:update` (no push alias; push stays bot-only).

## Pipeline

Solid arrows are control flow inside a script. Dashed arrows are cross-script invocations, always initiated by the workflow file.

```mermaid
flowchart TD
    subgraph workflow [".github/workflows/coinspect-refresh.yaml"]
        cron["cron (weekly, Sun 00:00 UTC) or manual dispatch"]
        checkout["step: checkout beta + install debian-packages.txt"]
        stepUpdate["step: run ./deploy/coinspect/coinspect-update.sh"]
        stepPush["step: run ./deploy/setup-git.sh && ./deploy/coinspect/coinspect-push.sh"]
        cron --> checkout --> stepUpdate --> stepPush
    end

    subgraph updateScript ["deploy/coinspect/coinspect-update.sh"]
        u1["git ls-remote coinspect main -> remote_sha"]
        u2{"remote_sha == cat data/coinspect/upstream-commit ?"}
        uFast["exit 0 (fast-path, no clone)"]
        uClone["sparse shallow clone at remote_sha into tempdir"]
        uGuard{"current-reports/ present and non-empty?"}
        uAbort["exit 1 (refuse to wipe local tree)"]
        uRsync["rsync -a --delete --exclude=images/ --exclude=images.json into data/coinspect/current-reports/"]
        uWrite["overwrite data/coinspect/upstream-commit with remote_sha"]
        uDone["exit 0"]
        u1 --> u2
        u2 -- yes --> uFast
        u2 -- no --> uClone --> uGuard
        uGuard -- no --> uAbort
        uGuard -- yes --> uRsync --> uWrite --> uDone
    end

    subgraph setupScript ["deploy/setup-git.sh (reused, no changes)"]
        s1["configure walletbeat-bot git identity + SSH auth + signing keys"]
    end

    subgraph pushScript ["deploy/coinspect/coinspect-push.sh"]
        p1{"git status --porcelain data/coinspect/ dirty?"}
        pNoop["exit 0"]
        pCommit["git add data/coinspect/ + signed commit"]
        pPush["git push origin beta"]
        p1 -- no --> pNoop
        p1 -- yes --> pCommit --> pPush
    end

    stepUpdate -. "invokes" .-> u1
    stepPush -. "invokes first" .-> s1
    stepPush -. "then invokes" .-> p1
```

A fast-path exit from `coinspect-update.sh` still continues to the push step; `coinspect-push.sh` then no-ops on a clean `data/coinspect/` tree. The sha file is written only after a successful clone + rsync, so `upstream-commit` always names a commit whose tree we actually vendored. Upstream commits that only touch excluded image bundles still produce a bot commit, because `upstream-commit` itself changes.


.github/workflows/coinspect-refresh.yaml
```diff
@@ -0,0 +1,35 @@
+name: Update Coinspect snapshot
+on:
+  workflow_dispatch: {} # On demand
+  schedule: # On schedule
+    - cron: '0 0 * * 0' # Weekly, Sunday 00:00 UTC
```
polymutex: Let's keep it on demand for now until it can successfully run a few times manually.

data/coinspect/README.md
```diff
@@ -0,0 +1,14 @@
...
+[Wallet Security Ranking](https://github.com/coinspect/wallet-security-ranking)
+data repository.
+
+- Upstream source: <https://github.com/coinspect/wallet-security-ranking>
+- Local tree: `current-reports/` (mirrored from upstream `current-reports/`;
+  image bundles are omitted)
+- Upstream commit pin: `upstream-commit`
+- License: see [`LICENSE`](./LICENSE) (CC BY 4.0)
+
+Refreshed by `deploy/coinspect/coinspect-update.sh` (see also
```
polymutex: Use `/`-prefixed URLs to refer to the repository root.

polymutex: Use `/`-prefixed URLs to refer to the repository root.

tests/coinspect-integrity.test.ts
```diff
@@ -139,3 +199,35 @@ describe('coinspect mapping', () => {
...
+
+		expect(localSha, `${UPSTREAM_COMMIT_FILE} must be a full 40-character SHA-1`).toMatch(FULL_SHA1)
+	})
+
+	it('pins the same commit as upstream main', () => {
+		const pinPath = path.join(getRepositoryRoot(), UPSTREAM_COMMIT_FILE)
+		const localSha = fs.readFileSync(pinPath, { encoding: 'utf-8' }).trim()
+		const lsRemote = execSync(`git ls-remote ${UPSTREAM_REPO} ${UPSTREAM_REF}`, {
+			encoding: 'utf-8',
+		})
```
polymutex: Let's not have any network-requiring tests in the default test suite. This would be inherently flaky and rely on GitHub.

In addition, a failure here isn't really the PR author's job to fix, it is up to the automation to catch up. So this would be a non-actionable failure.

I think this part of the test should be either put in its own GitHub CI step, or dropped entirely, or replaced with a very unlikely-to-fail "stale for over a year" type threshold that indicates a dead pipeline.

polymutex (changes requested):
Looking good.

How about removing all the `data/coinspect` files from this PR? They should be auto-fetched on first pipeline run, correct?
Maybe leave 1 or 2 just to keep the tests happy but then let the pipeline do the rest.

tests/coinspect-integrity.test.ts
johngrantuk: Makes sense. I've dropped it. (Kept the local SHA-format check).

deploy/coinspect/coinspect-update.sh
```diff
@@ -0,0 +1,64 @@
...
+# Refresh the vendored Coinspect current-reports/ snapshot.
+
+set -euo pipefail
+set +x
+
+if ! hash git; then
+	echo 'git not installed.' >&2
+	exit 1
+fi
+if ! hash rsync; then
```
polymutex, line 12: Don't we need to add `rsync` to  `deploy/debian-packages.txt`?

deploy/coinspect/coinspect-update.sh
```diff
@@ -0,0 +1,64 @@
...
+	echo 'rsync not installed.' >&2
+	exit 1
+fi
+
+UPSTREAM_REPO='https://github.com/coinspect/wallet-security-ranking'
+UPSTREAM_REF='main'
+LOCAL_COMMIT_FILE='data/coinspect/upstream-commit'
+LOCAL_REPORTS_DIR='data/coinspect/current-reports'
+
+remote_sha=$(git ls-remote "$UPSTREAM_REPO" "$UPSTREAM_REF" | cut -f1)
```
polymutex: Wrap the `$()` in quotes

deploy/coinspect/coinspect-update.sh
```diff
@@ -0,0 +1,64 @@
...
+else
+	local_sha=''
+fi
+
+if [[ "$local_sha" == "$remote_sha" ]]; then
+	echo 'Coinspect upstream is unchanged.' >&2
+	exit 0
+fi
+
+tmp=$(mktemp -d)
```
polymutex: Same

polymutex (changes requested):

polymutex (approved):

PR #1150, opened 2026-08-25
References/inline code snippets

### LINKED ISSUE #1124: Wallet `ref`s: Support showing inline code snippets
Much like we support image-based references (for screenshots), we also tend to have code-based snippet links.
When on GitHub, these are currently enforced to be a commit-specific hash, possibly with line numbers in their anchor part. Example: `https://github.com/zeriontech/zerion-wallet-extension/blob/482c0a5f57cee79b618147c804a92a98240c559a/src/background/Wallet/Wallet.ts#L2232-L2242`

We should have:
- A `data/` subdirectory that holds the actual code text that this corresponds to, e.g. `/data/code-snippets/github.com/zeriontech/zerion-wallet-extension/482c0a5f57cee79b618147c804a92a98240c559a/src/background/Wallet/Wallet.ts.L2232-L2242.snippet`. We can't have the whole file verbatim, as this would be potentially a license violation depending on the remote repository's license; however snippets should be OK under fair use, hence storing only the snippet.
- A tool to automatically fetch them given a set of snippet URLs to the subdirectory
- A test to ensure they are in sync
- Code in the UI to automatically render a snippet of code when a code reference matching this format is used

maykelxyz:
# Feat: Show inline code snippets for reference links

Reference links pointing at commit-pinned, line-anchored GitHub blob URLs (`https://github.com/<org>/<repo>/blob/<40-char-hash>/<path>#L10-L20`) now render the referenced lines inline as a syntax-highlighted code block, with the link itself as the caption. Only the referenced lines are stored locally, never the whole file, to stay within fair use regardless of the source repository's license.

## Changes

- Added `src/schema/code-snippets.ts`: parsing/naming scheme shared by the collector, the UI, and tests:
```ts
export interface CodeSnippetSource {
	org: string
	repo: string
	/** Full 40-character commit hash the URL is pinned to. */
	commit: string
	/** Repository-relative file path, using forward slashes, percent-decoded. */
	path: string
	/** 1-based first line of the snippet, inclusive. */
	firstLine: number
	/** 1-based last line of the snippet, inclusive. */
	lastLine: number
}
```
- Snippets are stored as flat files under `public/references/wallets/<wallet-id>/code/`, named `<org>--<repo>--<commit>--<path with '/' as '--'>.L<first>[-L<last>].<ext>.snippet`. The source extension sits right before `.snippet` so the language stays identifiable.
- Added `pnpm collect:snippets` (`src/tools/code-snippet-collector/`):
  - `--all` scans `data/*-wallets/*.ts` for snippet URLs and fetches missing snippets from `raw.githubusercontent.com`. Already-stored snippets are skipped (URLs are commit-pinned, so content never changes).
  - `--url <blob-url>` fetches a single snippet, `--check` verifies sync offline, `--prune` deletes orphaned snippet files.
- Added `vite-plugin-code-snippet-highlight.mjs`: `.snippet` imports become arrays of HTML strings, one per line, syntax-highlighted with Shiki at build time. Keeps Shiki and its grammars out of the client bundle while SSR and hydration output stay identical. Unknown extensions fall back to plain escaped text.
- `src/utils/code-snippet-index.ts` bundles all stored snippets via `import.meta.glob` and resolves a reference URL to its snippet, keyed by filename only, since the filename fully encodes the source location, so resolution doesn't need to know the wallet.
- `ReferenceLinks.svelte` renders snippet-backed URLs as `<figure>` code blocks with line numbers and a caption link, and drops them from the plain link list.
- Added `tests/code-snippets-integrity.test.ts`: fails when stored snippets are missing, orphaned, or have a line count that doesn't match their filename, with instructions on which `collect:snippets` mode fixes it. Content drift is impossible since URLs are commit-pinned.
- Collected snippets for all currently referenced URLs (79 files).

## Other changes

- Exported `fullCommitHashRegExp` and `lineFragmentRegExp` from `src/schema/url.ts`
- Excluded `*.snippet` files from prettier, cspell, and codebase integrity tests
- Added Shiki to cspell

Closes #1124


tests/codebase-integrity.test.ts
```diff
@@ -59,6 +59,8 @@ const ALLOWED_EXTENSIONS = new Set([
 	'.tsv',
 	'.yaml',
 	'.toml',
+	// Stored code snippets from external repositories,
```
polymutex: No need for this comment.

.cspell.json
```diff
@@ -693,6 +694,7 @@
 		"node_modules/**",
 		"pnpm-lock.yaml",
 		"public/**/*.svg",
+		"public/references/wallets/*/code/**",
```
polymutex: Restrict to files ending in `.snippet`

vite-plugin-code-snippet-highlight.mjs
```diff
@@ -0,0 +1,154 @@
...
+ * build time. Highlighting at build time keeps Shiki (and its grammars) out
+ * of the client bundle while producing identical output for server-side
+ * rendering and client hydration.
+ *
+ * The language comes from the source file's extension, which the snippet
+ * naming scheme keeps right before the `.snippet` suffix
+ * (e.g. `...--Wallet.L2232-L2242.ts.snippet` → `ts`).
+ */
+
+/** Shiki theme used for all snippets. The site is dark-only. */
```
polymutex: No it's not

vite-plugin-code-snippet-highlight.mjs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/** @type {Set<string>} */
+const loadedLanguages = new Set()
+
+/**
+ * @param {string} text
+ * @returns {string}
+ */
+function escapeHtml(text) {
```
polymutex, line 78: Surely there's a better way to do this

vite-plugin-code-snippet-highlight.mjs
```diff
@@ -0,0 +1,154 @@
...
+
+	if (parts.length < 2) {
+		return undefined
+	}
+
+	const extension = parts[parts.length - 2]
+
+	return Object.hasOwn(extensionToLanguage, extension)
+		? extensionToLanguage[/** @type {keyof typeof extensionToLanguage} */ (extension)]
+		: undefined
```
polymutex: Should error if not found rather than `undefined`.

If we ever encounter an extension we shouldn't highlight, then we should explicitly encode it here as known-unhighlighted.

tests/code-snippets-integrity.test.ts
```diff
@@ -0,0 +1,41 @@
...
+			const needsFetch = problems.some(
+				p =>
+					p.kind === SnippetProblemKind.MISSING_SNIPPET ||
+					p.kind === SnippetProblemKind.LINE_COUNT_MISMATCH,
+			)
+			const needsPrune = problems.some(p => p.kind === SnippetProblemKind.ORPHAN_SNIPPET)
+			const fixInstructions = [
+				needsFetch
+					? 'Run `pnpm collect:snippets -- --all` to fetch missing snippets ' +
+						'(delete the file first for LINE_COUNT_MISMATCH).'
```
polymutex: Either have the collect tool auto-delete and remove this message, or show this message only when there is at least one `LINE_COUNT_MISMATCH` problem

polymutex (changes requested):
Looks really nice and now I'm convinced this is a great idea.  
(Yes, it's also my idea, but I'd have said the same if it wasn't lol)

Some UI feedback:

- I think the snippets would look a lot better if we add surrounding lines of code (like +3 or +4 lines either before/after, as well as the surrounding indentation context similar to how `diff` does it (basically, find the last line that is at least one indentation level earlier than the first line of the snippet, which is usually a function name or class name; repeat for each smaller indentation level). Then, the actual referenced lines can be highlighted (different background color).
- The link above the snippet should still include the filename/commit/line numbers as it did previously (it should also show the label as it does today, it's just that it has the space to show both IMO)
- When hovering over the link, the underline extends until the `<>` icon but doesn't go over it. IMO it should either stop before the last space character before the `<>`, or it should go over the whole `<>` icon. The "external link" icon has the same issue btw, so maybe this is a different PR's problem to solve.

Didn't look at the collect tool yet.

polymutex (changes requested):
A few UI notes (still haven't reviewed the pipeline code):

- Scrollbars are ugly and should only be shown when hovering IMO
- The middle of the highlighted portion of the code should be vertically centered in the snippet box initially
- The highlighted and non-highlighted portions of the code are almost indistinguishable from one another (background color is pretty much the same)

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+		.map(entry => entry.name)
+
+	for (const walletDir of walletDirs) {
+		const dirPath = path.join(dataDir, walletDir)
+		const walletFiles = fs
+			.readdirSync(dirPath, { withFileTypes: true })
+			.filter(
+				entry => entry.isFile() && entry.name.endsWith('.ts') && !entry.name.endsWith('.tmpl.ts'),
+			)
+			.map(entry => entry.name)
```
polymutex, line 57: Please no. Use `collectAllRefs`.

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+		}
+	}
+
+	return occurrences
+}
+
+/** Length of a line's leading whitespace, used as its indentation. */
+function indentationOf(line: string): number {
+	return /^[ \t]*/.exec(line)?.[0].length ?? 0
+}
```
polymutex, line 92: This should use the specific whitespace prefix, not its length. See `trimWhitespacePrefix` for logic that determines common whitespace across lines.

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+		!Array.isArray(candidate.segments) ||
+		!candidate.segments.every(
+			(segment): segment is StoredSnippetSegment =>
+				typeof segment === 'object' &&
+				segment !== null &&
+				typeof segment.startLine === 'number' &&
+				Array.isArray(segment.lines) &&
+				segment.lines.every((line: unknown) => typeof line === 'string'),
+		)
+	) {
```
polymutex, line 285: wat

A bit difficult to follow...

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+	kind: SnippetProblemKind
+	/** Repository-relative path of the snippet file the problem is about. */
+	snippetPath: string
+	issue: string
+}
+
+const walletsReferencesDir = path.join('public', 'references', 'wallets')
+
+/** Recursively list all repository-relative file paths under `dir`. */
+function listFilesRecursively(repoRoot: string, dir: string): string[] {
```
polymutex, line 335: Use `crawlCodebase` instead of this function.

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+		} else if (entry.isFile()) {
+			files.push(entryPath)
+		}
+	}
+
+	return files
+}
+
+/** Repository-relative paths of all files under wallets' code snippet directories. */
+export function listStoredSnippetFiles(repoRoot: string): string[] {
```
polymutex, line 358: Same, this shouldn't be needed

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,450 @@
...
+	 * to: either the reference was removed or changed, or the file does not
+	 * follow the snippet naming scheme.
+	 */
+	ORPHAN_SNIPPET = 'ORPHAN_SNIPPET',
+	/**
+	 * A stored snippet that isn't valid `StoredSnippetContent` JSON (including
+	 * the old flat-text format), or whose highlighted range doesn't match the
+	 * range in its filename.
+	 */
+	STALE_CONTENT = 'STALE_CONTENT',
```
polymutex, line 322: When would this ever happen?

src/tools/code-snippet-collector/code-snippet-collector.ts
```diff
@@ -0,0 +1,242 @@
...
+The referenced lines are stored along with a small bounded window of
+surrounding context and enclosing scope-header lines (e.g. the containing
+function/class signature) — never the whole file — to stay within fair use
+regardless of the source repository's license.
+
+Modes (pick one):
+  --all                  Scan data/*-wallets/*.ts for snippet URLs and fetch
+                         every snippet that is not stored yet. Already-stored
+                         snippets are skipped: URLs are commit-pinned, so
+                         their content never changes.
```
polymutex, line 38: Not useful

src/tools/code-snippet-collector/code-snippet-collector.ts
```diff
@@ -0,0 +1,242 @@
...
+	process.stderr.write(`
+Usage: pnpm collect:snippets [options]
+
+Stores local copies of the code snippets that wallet data files reference via
+commit-pinned, line-anchored GitHub blob URLs
+(https://github.com/<org>/<repo>/blob/<40-char-hash>/<path>#L<first>-L<last>).
+The referenced lines are stored along with a small bounded window of
+surrounding context and enclosing scope-header lines (e.g. the containing
+function/class signature) — never the whole file — to stay within fair use
+regardless of the source repository's license.
```
polymutex, line 32: Not useful

src/tools/code-snippet-collector/code-snippet-collector.ts
```diff
@@ -0,0 +1,242 @@
...
+
+if (hasFlag('help')) {
+	usage()
+}
+
+const allMode = hasFlag('all')
+const checkMode = hasFlag('check')
+const pruneMode = hasFlag('prune')
+const urlArg = getArg('url')
+const walletIdArg = getArg('id')
```
polymutex, line 83: Use `cac`

package.json
polymutex, line 1: Shouldn't `pnpm fix` automatically fix orphan/missing snippets?

vite-plugin-code-snippet-highlight.mjs
polymutex, line 1: Can this not be in the repo root?

polymutex (changes requested):

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
maykelxyz: Got i t. Resolved to:

```ts
/**
 * Find every line-anchored, commit-pinned GitHub blob URL among wallet data
 * refs, paired with the wallet ID it was found under.
 */
export function findSnippetOccurrences(_repoRoot: string): SnippetOccurrence[] {
	const occurrences: SnippetOccurrence[] = []

	for (const collected of collectAllRefs(allWallets)) {
		// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- collected.walletName always comes from Object.entries(allWallets)'s own keys.
		const walletId = allWallets[collected.walletName as keyof typeof allWallets].metadata.id

		for (const fq of collected.fullyQualifiedRefs) {
			for (const urlEntry of fq.urls) {
				const source = parseGitHubBlobUrl(urlEntry.url)

				if (typeof source === 'string') {
					continue
				}

				occurrences.push({
					fieldPath: collected.fieldPath,
					snippetPath: snippetRelativePath(walletId, source),
					source,
					url: urlEntry.url,
					walletId,
				})
			}
		}
	}

	return occurrences
}
```

package.json
maykelxyz: you're right, added.

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
maykelxyz: If someone edits the .snippet files directly (either making an invalid JSON, or changing the range). Updated it to SNIPPET_CONTENT_MISMATCH, probably better naming convention

polymutex:
Is this ready for review? (Also please rebase)

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,409 @@
...
+/**
+ * Find every line-anchored, commit-pinned GitHub blob URL among wallet data
+ * refs, paired with the wallet ID it was found under.
+ */
+export function findSnippetOccurrences(_repoRoot: string): SnippetOccurrence[] {
+	const occurrences: SnippetOccurrence[] = []
+
+	for (const collected of collectAllRefs(allWallets)) {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- collected.walletName always comes from Object.entries(allWallets)'s own keys.
+		const walletId = allWallets[collected.walletName as keyof typeof allWallets].metadata.id
```
polymutex: Use `assertValidWalletName` instead of disabling eslint.

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,409 @@
...
+					`but its filename declares ${occurrence.source.firstLine}-${occurrence.source.lastLine}.`,
+				kind: SnippetProblemKind.SNIPPET_CONTENT_MISMATCH,
+				snippetPath,
+			})
+		}
+	}
+
+	const walletsDir = path.join(repoRoot, walletsReferencesDir)
+
+	if (fs.existsSync(walletsDir)) {
```
polymutex: This should never be `false`, and if it is, something is wrong. So this `if` can be removed, and if something is wrong, then it's better for the error to be thrown.

src/tools/code-snippet-collector/code-snippet-collector.ts
```diff
@@ -0,0 +1,224 @@
...
+				process.stderr.write(
+					'Error: not a commit-pinned, line-anchored GitHub blob URL.\n' +
+						'Expected: https://github.com/<org>/<repo>/blob/<40-char-hash>/<path>#L<first>[-L<last>]\n',
+				)
+				process.exit(1)
+			}
+
+			source = parsed
+		} catch (error) {
+			process.stderr.write(`Error: ${error instanceof Error ? error.message : String(error)}\n`)
```
polymutex: Use `getErrorMessage`

src/tools/code-snippet-collector/code-snippet-collector-lib.ts
```diff
@@ -0,0 +1,409 @@
...
+
+	for (const collected of collectAllRefs(allWallets)) {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion -- collected.walletName always comes from Object.entries(allWallets)'s own keys.
+		const walletId = allWallets[collected.walletName as keyof typeof allWallets].metadata.id
+
+		for (const fq of collected.fullyQualifiedRefs) {
+			for (const urlEntry of fq.urls) {
+				const source = parseGitHubBlobUrl(urlEntry.url)
+
+				if (typeof source === 'string') {
```
polymutex: Over-broad, as this would likely unintendedly expand if the set of sentinel strings `parseGitHubBlobUrl` returns were to grow. Need to verify each of the specialized strings.

src/utils/code-snippet-index.ts
```diff
@@ -0,0 +1,100 @@
...
+export function codeSnippetForUrl(url: string): ResolvedCodeSnippet | null {
+	let source: ReturnType<typeof parseGitHubBlobUrl>
+
+	try {
+		source = parseGitHubBlobUrl(url)
+	} catch {
+		return null
+	}
+
+	if (typeof source === 'string') {
```
polymutex: This pattern repeats, which says something. e.g. maybe it's time for a helper, or a custom exception type.

src/tools/code-snippet-collector/code-snippet-collector.ts
```diff
@@ -0,0 +1,224 @@
...
+
+		process.exit(0)
+	})
+
+cli.help()
+
+try {
+	cli.parse(process.argv, { run: true })
+} catch (error) {
+	process.stderr.write(`Error: ${error instanceof Error ? error.message : String(error)}\n`)
```
polymutex: `getErrorMessage`

tests/code-snippets-integrity.test.ts
```diff
@@ -0,0 +1,40 @@
...
+			const needsPrune = problems.some(p => p.kind === SnippetProblemKind.ORPHAN_SNIPPET)
+			const fixInstructions = [
+				needsFetch
+					? 'Run `pnpm collect:snippets -- all` to fetch missing snippets ' +
+						'and refetch any with stale or mismatched content.'
+					: null,
+				needsPrune
+					? 'Run `pnpm collect:snippets -- prune` to delete orphaned snippet files.'
+					: null,
+			]
```
polymutex: Just `pnpm fix`

package.json
```diff
@@ -46,11 +46,12 @@
 		"dev:background:restart": "deploy/dev/background-dev.sh --restart",
 		"lint": "pnpm run check:syntax --write && pnpm run check:lint --fix && pnpm run check:spelling:reorder",
 		"preview": "pnpm run astro preview",
-		"fix": "pnpm run lint && pnpm run check:treasury:markdown && pnpm run wallet-data-collection lint-fix && pnpm run check:features:markdown && pnpm run generate:css-attributes && pnpm run generate:wbicons",
+		"fix": "pnpm run lint && pnpm run check:treasury:markdown && pnpm run wallet-data-collection lint-fix && pnpm run check:features:markdown && pnpm run generate:css-attributes && pnpm run generate:wbicons && pnpm collect:snippets all",
```
polymutex: This should automatically fix everything about snippets, i.e. fetch missing **and** prune old ones. Should be its own subcommand.

polymutex (changes requested):

PR #1157, opened 2026-08-27
Deduplicate `Check` workflow and isolate dist checks

MattPereira:
The `check-full` job repeated all the checks already run by `check-quick`

This PR changes the second job to run only the dist-related checks, reducing total CI time.


polymutex (commented):
I don't think we need this... I see `check:all` as performing two things:
- Be a very easy way for agents to check their work across all dimensions we may want to care about that they may have forgotten to consider
- Run as part of CI to double check everything
Having a dumb "just test everything" test is easy to maintain, hard to forget to update, and fulfills both of these goals at once.
By contrast, having something specific to each means that we need to maintain two separate ways of running everything, which introduces the possibility of drift between the two.
Meanwhile, CI takes ~15 minutes and it's a non-interactive process where quick feedback isn't always useful; and when it is, that's why we already have the quick individual CI steps. Yes it's redundant, yes the end-to-end CI time is not as fast as it could be, but: why does that matter?
Maybe if we were paying for GitHub Actions time that would be a different conversation, but we are well within the free compute :)

PR #1159, opened 2026-08-28
Data: Add Uniswap wallet extension

maykelxyz:
# Uniswap data
- Filled in most of Uniswap's previously-null feature data:
  - `accountSupport`: EOA (BIP32/BIP44/BIP39, exportable seed) + EIP-7702 via the `uniswapCalibur` smart wallet contract
  - `addressResolution`: non-chain-specific ENS resolution via the Uniswap Unitags API (`OFFCHAIN`, `VERIFIABLE`)
  - `chainAbstraction`: built-in bridging with `AGGREGATED` fee display by default, `COMPREHENSIVE` after expanding, plus cross-chain balance views for ETH/USDC
  - `chainConfigurability`: no chain/RPC management in Settings, so `notSupported`
  - `ecosystem.delegation`: bundled EIP-7702 delegation on first operation
  - `integration.browser`: EIP-1193, EIP-2700, EIP-6963 all supported
  - `privacy.appIsolation`: `APP_SPECIFIC_ACCOUNT` behavior for `eth_accounts`
  - `privacy.transactionPrivacy`: no private transfer techniques supported, `defaultFungibleTokenTransferMode: 'PUBLIC'`
  - `security.accountRecovery`: no guardian recovery or drills
  - `security.duressResistance`: browser (password) and mobile (biometric) unlock mechanisms, no duress mode
  - `security.keysHandling`: `FULLY_ON_USER_DEVICE` key generation, `NON_MULTIPARTY`
  - `security.publicSecurityAudits`: two Trail of Bits audits (mobile and browser extension), each with their unpatched flaws
  - `security.securityBestPractices`: browser (PBKDF2 + AES-GCM encrypted mnemonic) and mobile (Android `EncryptedSharedPreferences`/Keystore, iOS Keychain/Secure Enclave), parsed from collected manifests
  - `selfSovereignty.permissionsManagement`: `notSupported`
  - `selfSovereignty.transactionSubmission.l2`: `SUPPORTED_BUT_NO_FORCE_INCLUSION` for both Arbitrum and OP Stack
  - `walletCall`: `wallet_sendCalls` supported, no atomicity guarantee
- Added `data/wallet-contracts/uniswap-calibur.ts` for the EIP-7702 delegate contract (`isValidSignature`, `validateUserOp`, ERC-7821 batching)
- Added `data/entities/trail-of-bits.ts` as a `SecurityAuditor` entity
- Collected Android/iOS mobile manifests and the browser extension manifest under `data/software-wallets/manifests/uniswapWallet/`, parsed via `parseMobileManifestJson`/`parseBrowserExtensionManifest`
- Added supporting screenshots under `public/references/wallets/uniswap/screenshots/`

## Attributes
- Added `AndroidPermission.USE_BIOMETRIC`, `USE_FINGERPRINT`, `VIBRATE`, `READ_MEDIA_IMAGES` (all `PASS`) and `AD_ID` (`FAIL`, ad tracking)
- Added `IosUsageDescription.LOCATION_ALWAYS_AND_WHEN_IN_USE` (`FAIL`, background tracking not needed by a wallet)

## Other changes
- Added `RnEthersRs`, `UNIEXT`, `UNIMOB`, `Unitags` to cspell


pr.md
polymutex, line 1: ?

data/software-wallets/uniswap-wallet.ts
```diff
@@ -35,26 +121,139 @@ export const uniswapWallet: SoftwareWallet = {
...
+			],
 			chainSpecificAddressing: {
-				erc7828: null,
-				erc7831: null,
+				erc7828: notSupported,
+				erc7831: notSupported,
+			},
+			nonChainSpecificEnsResolution: supported<AddressResolutionData>({
+				medium: 'OFFCHAIN',
+				offchainDataVerifiability: 'VERIFIABLE',
```
polymutex: How so?

data/software-wallets/uniswap-wallet.ts
```diff
@@ -35,26 +121,139 @@ export const uniswapWallet: SoftwareWallet = {
...
+							{
+								file: 'public/references/wallets/uniswap/screenshots/2026-08-31-swap-info-comprehensive.png',
+								label:
+									'Uniswap Wallet swap review screen, expanded, showing rate, network cost, slippage, and route',
+							},
+						],
+						afterSingleAction: FeeDisplayLevel.COMPREHENSIVE,
+						byDefault: FeeDisplayLevel.AGGREGATED,
+						fullySponsored: false,
+						walletServiceFeeDisplayUnits: nonEmptySet(WalletServiceFeeDisplayUnit.PERCENTAGE),
```
polymutex: I don't see this in the screenshot.

polymutex (changes requested):

data/software-wallets/uniswap-wallet.ts
maykelxyz: Yep, you're right. Updated it

data/software-wallets/uniswap-wallet.ts
maykelxyz: Originally, the idea was because they were using an endpoint to query names, I set the medium to 'OFFCHAIN'.. they were still doing onchain calls though. I guess CHAIN_CLIENT is more fitting.

data/software-wallets/uniswap-wallet.ts
```diff
@@ -35,26 +117,142 @@ export const uniswapWallet: SoftwareWallet = {
...
+				{
+					explanation:
+						'Entering an ENS name like vitalik.eth in the Send flow resolves it to the corresponding address, regardless of the destination chain.',
+					file: 'public/references/wallets/uniswap/screenshots/non-chain-specific-ens-resolution.png',
+					label: 'Uniswap Wallet Send flow resolving vitalik.eth to an address',
+				},
+				{
+					explanation:
+						'Username/ENS resolution is fetched from the Uniswap Unitags API, an offchain service, rather than resolved directly onchain.',
+					url: 'https://github.com/Uniswap/interface/blob/da6d36f71c4d2fd665b0aae1a052a4ffda917b31/packages/uniswap/src/data/apiClients/unitagsApi/useUnitagsUsernameQuery.ts',
```
polymutex, line 156: Seems to contradict the `CHAIN_CLIENT` below. But the `getOnChainEnsFetch` thing does look like it's using the chain client. So which is it? Need to dig into why it uses one or the other.

polymutex (changes requested):

data/software-wallets/uniswap-wallet.ts
```diff
@@ -35,26 +117,142 @@ export const uniswapWallet: SoftwareWallet = {
...
+							{
+								file: 'public/references/wallets/uniswap/screenshots/2026-08-31-swap-info-comprehensive.png',
+								label:
+									'Uniswap Wallet swap review screen, expanded, showing rate, network cost, slippage, and route',
+							},
+						],
+						afterSingleAction: FeeDisplayLevel.COMPREHENSIVE,
+						byDefault: FeeDisplayLevel.AGGREGATED,
+						fullySponsored: false,
+						walletServiceFeeDisplayUnits: 'NOT_APPLICABLE' as const,
```
polymutex, line 199: Does it charge no fee?

polymutex (changes requested):

data/software-wallets/uniswap-wallet.ts
maykelxyz: It seems like it. Also did two more txs in mainnet and showed no trace to other addresses aside from the swap contract

image
image

data/software-wallets/uniswap-wallet.ts
maykelxyz: Yep, just saw it. They use both. They use both lookups in parallel. Username resolution is being fetched through Unitags API, but if an ENS matches, they use the ENS resolved address

https://github.com/Uniswap/interface/blob/da6d36f71c4d2fd665b0aae1a052a4ffda917b31/apps/web/src/pages/Swap/Send/state/hooks.tsx#L58-L86

polymutex: OK. Then unfortunately I think we need to extend the feature types to properly encode this.
For this PR's purposes, I suggest either doing the feature type extension in another PR (and leaving this PR on hold until then), or to leave this PR's field as `null` and then doing the feature type extension in another PR.

maykelxyz: Originally I also thought that we need to update the feature types in this PR but I think the PR can stand as is for now for these reasons:
-> The question we're asking for `nonChainSpecificEnsResolution` is does it resolve ENS domains? Uniswap wallet answers that question without confusion, no? They have another feature "unitags username" but I don't think that's an issue
```ts
export interface AddressResolution<ARS = Support<AddressResolutionData> | null> {
	/**
	 * Support for basic ENS lookups (ENS domain to non-chain-specific raw hex
	 * address).
	 * To test: type `donations.walletbeat.eth` in the send address field. If it resolves, it is supported.
	 */
	nonChainSpecificEnsResolution: ARS
```

-> I don't think the feature type extension would change the value of nonChainSpecificEnsResolution. We'd likely just add another feature to capture wallet-specific username resolution, something along the lines of walletUsernameResolution: "Does the wallet have its own username resolution service?"

polymutex: OK, but there's still the problem that the presence of this `ref` in the attribute that explains that ENS resolution is done with the chain client still looks contradictory when rendered on the site.

One way to resolve this may be to delete it, but to preserve this good piece of information, I think it may be useful to move this into a `note` in the not-often-used `overrides.attributes.[...].note` field (see `daimo.ts` for example usage).

polymutex (changes requested):

data/software-wallets/uniswap-wallet.ts
maykelxyz: Ok yeah, moved it. Looks like the `note` property is not actually being shown in the UI. Intended? if not, then I'll create an issue

polymutex (approved):
Need to file separate issue for the note display on the UI

PR #1162, opened 2026-08-30
MetaMask: Cite the source of the public revenue breakdown

ren2140eth:
Adds references for MetaMask's monetization data. Revenue breakdown was true previously with no ref profided

### DIFF
--- a/data/software-wallets/metamask.ts
+++ b/data/software-wallets/metamask.ts
@@ -278,7 +278,7 @@ export const metamask: SoftwareWallet = {
 			ref: [
 				{
 					explanation:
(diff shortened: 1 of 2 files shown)


data/software-wallets/metamask.ts
```diff
@@ -278,9 +278,22 @@ export const metamask: SoftwareWallet = {
 			ref: [
 				{
 					explanation:
-						'MetaMask is funded through transparent swap fees and venture capital with publicly disclosed funding rounds.',
+						"DeFiLlama publishes MetaMask's fee and revenue figures, broken down by product line (wallet swaps, perpetuals, predictions and MetaMask USD) and by chain, with 24-hour, 30-day and all-time totals.",
```
ren2140eth: cut it off after chain.

ren2140eth: Trimmed in 48c42e56.

maykelxyz (approved):

PR #1163, opened 2026-08-30
Zerion: Split mobile and browser licensing

ren2140eth:
Records Zerion's licensing per variant: and splits it by browser and mobile (previously was only browser)

**Schema limitation.** `LicensingType` is declared once per wallet, but the license values under it are per-variant. Zerion needs `SINGLE_WALLET_REPO_AND_LICENSE` for the browser and `SEPARATE_CORE_CODE_LICENSE_VS_WALLET_CODE_LICENSE` for mobile, which the union can't express, so this PR uses `SEPARATE` wallet-wide and sets the browser's `coreLicense` to the extension's own GPL-3.0. Both values being identical, they collapse to one license and the browser still rates PASS, but the data records a core/app boundary in the extension that does not exist. Fixing it properly means moving `type` into the per-variant resolution in `license.ts` and updating the consumers of `ResolvedWalletLicensing` can open an issue on this


ren2140eth:
Noting a schema limitation this change runs into, for whoever wants to pick it up later.

`LicensingType` is declared once for the whole wallet, but the license values underneath it are per-variant (`VariantFeature<LicenseWithRef>`). That assumes a wallet splitting core from app code splits that way on every variant.

Zerion doesn't. Its browser extension is genuinely one repo under one license (`SINGLE_WALLET_REPO_AND_LICENSE`); its mobile app is genuinely a public Apache-2.0 core plus a closed app (`SEPARATE_CORE_CODE_LICENSE_VS_WALLET_CODE_LICENSE`). No single value of `type` describes both, and the union in `WalletLicensing` (`src/schema/features/transparency/license.ts:249`) can't express one shape per variant.

This PR takes `SEPARATE` wallet-wide and sets the browser's `coreLicense` to the extension's own GPL-3.0, the same license already on its `walletAppLicense`. Both values being identical, the evaluator collapses them to one license and the browser still rates PASS (GPL-3.0) with a single merged reference. The cost is that the data records a core/app boundary in the extension that does not exist.

One latent consequence worth knowing: that only avoids `Unimplemented case; please implement.` at `src/schema/attributes/transparency/open-source.ts:247` because both browser values are the same enum. A wallet in the same position whose two licenses differed would hit that throw.

The proper fix is to move `type` inside the per-variant resolution in `license.ts` and update the consumers of `ResolvedWalletLicensing`. That's schema work rather than a data correction, so it's out of scope here.

data/software-wallets/zerion.ts
```diff
@@ -221,10 +225,55 @@ export const zerion: SoftwareWallet = {
...
+				},
+				[Variant.MOBILE]: {
+					ref: {
+						explanation:
+							'Zerion states only that the core code of its iOS and Android apps is open for review.',
+						label: 'Zerion security page, "Open source"',
+						lastRetrieved: '2026-08-30',
+						url: 'https://zerion.io/security',
+					},
+					license: SourceNotAvailableLicense.PROPRIETARY,
```
maykelxyz, line 275: This seems to be incorrect.. It says "The core code for iOS, Android, and Chrome Extension apps is open for anyone to review." I dont think that implicitly says "Only the core code of the ios and andoid apps are the ones open for review"

Image

maykelxyz (changes requested):

data/software-wallets/zerion.ts
ren2140eth: yep that's fair but we do also know that the android and ios apps are not fully open source given there is nothing in their github so thought this was the best evidence given it's stated by zerion themselves that only the core portion of the ios and android apps are open source (so they are implying the full app is not). They do fully open source the browser variant tho so its their copy that is also slightly incorrect. An alternative is to just point to the repo and simply state that the ios and android apps are closed source (i was going back and forth between the two ultimately chose this since it seems a bit stronger given its a statement from the team but lmk what you think)

maykelxyz: > An alternative is to just point to the repo and simply state that the ios and android apps are closed source  

i think this is better

polymutex: I think the wording you have is fine other than this reordering:

> Zerion states _only_ that the core code of its iOS and Android apps is open for review

to:

> Zerion states that _only_ the core code of its iOS and Android apps is open for review

Or to be more explicit, you can say:

> While Zerion states that the core code of its iOS and Android apps is open for review, the broader mobile application code is not.

ren2140eth: thanks perfect will update with that

maykelxyz (approved):

maykelxyz (changes requested):
Looks good now, but has merge conflicts to be resolved

maykelxyz (approved):

PR #1185, opened 2026-09-08
Attribute: Add built-in swap approval rating to permissions management

### LINKED ISSUE #1168: Silent unlimited token approval through in-wallet swaps
Wallets might've been silently approving unlimited tokens w/o letting users know about the approval itself.

Did a normal swap transaction:

Image

Turns out, an unlimited approval is inside the transaction without letting the user know what they're actually signing.

Image

Probably same thing here: https://x.com/m477lander/status/2093667168248324394

maykelxyz:
Attribute: Add built-in swap approval rating to permissions management

## Changes

`permissions-management` now also rates a wallet's own built-in swap/bridge feature on
whether it requests exact-amount token approvals by default, not just whether existing
approvals can be inspected/revoked.

- Split `PermissionsManagement` into `approvalsManagement` (the old inspect/revoke fields,
  renamed to `ApprovalsManagement`) and a new `builtInSwapApprovals` field:
```ts
export interface PermissionsManagement {
	approvalsManagement: Support<ApprovalsManagement>
	builtInSwapApprovals: BuiltInSwapDefaultApprovalBehavior | 'NO_BUILT_IN_SWAP'
}
```
- Added `BuiltInSwapDefaultApprovalBehavior`: `EXACT_AMOUNT`, `UNLIMITED_BUT_EDITABLE`,
  `UNLIMITED_BUT_DISCLOSED`, `UNLIMITED_AND_UNDISCLOSED`.
- `permissionsManagement` is no longer `Support<PermissionsManagementSupport>`, it's always
  present now since `approvalsManagement` carries its own `Support` internally.

Overall rating is the worse of the approvals-management rating and the swap-approval rating:

| Rating      | Condition                                                                 |
| ----------- | -------------------------------------------------------------------------- |
| **Pass**    | Can inspect and revoke approvals, and built-in swaps (if any) default to exact-amount |
| **Partial** | Can inspect but not revoke, OR built-in swaps default to unlimited-but-editable |
| **Fail**    | Cannot inspect approvals, OR built-in swaps default to unlimited (disclosed or not) |

Only an exact-amount default passes, unlimited fails regardless of disclosure.

## Ambire, Base App, MetaMask, Phantom, Rabby, Rainbow, Zerion data
- Added `builtInSwapApprovals` for each wallet, verified by performing a small swap and
  inspecting either the approve calldata or the on-chain `Approval` event
  - Ambire, Base App, MetaMask, Rabby, Rainbow, Zerion: `EXACT_AMOUNT`
  - Phantom: `UNLIMITED_AND_UNDISCLOSED`, confirmed via decoded approve calldata showing max
    approval despite the review screen showing only the swap amount
- Rabby's `approvalsManagement` collapsed from per-variant (browser/mobile) to a single
  `notSupported`/ref block alongside the new swap rating

Closes #1168


src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -39,76 +44,165 @@ function worstControl(...controls: SpendingApprovalsControl[]): SpendingApproval
...
+	}
+}
+
+/** Only an exact-amount-by-default approval passes; any unlimited default fails, whether disclosed or not. */
+function ratingForSwapApprovals(behavior: BuiltInSwapDefaultApprovalBehavior): ExplicitRating {
+	switch (behavior) {
+		case BuiltInSwapDefaultApprovalBehavior.EXACT_AMOUNT:
+			return Rating.PASS
+		case BuiltInSwapDefaultApprovalBehavior.UNLIMITED_BUT_EDITABLE:
+			return Rating.PARTIAL
```
polymutex: Inconsistent with docstring and methodology text below. I think this case should FAIL too, fwiw.

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -39,76 +44,165 @@ function worstControl(...controls: SpendingApprovalsControl[]): SpendingApproval
...
 	}
 
-	if (worst === SpendingApprovalsControl.CAN_INSPECT_BUT_NOT_REVOKE) {
+	if (overallRating === Rating.PARTIAL) {
+		const isSwapDriven = approvalsRating === Rating.PASS && swapRating === Rating.PARTIAL
+
 		return ctx.build({
 			outcome: {
-				id: 'can_inspect_not_revoke',
+				id: isSwapDriven ? 'editable_unlimited_swap_approval' : 'can_inspect_not_revoke',
```
polymutex: If changing the above to FAIL, then `isSwapDriven` should become unreachable and can be simplified away here.

polymutex (changes requested):
Great research here, and feature types look clean. Just an inconsistency to clarify on methodology.

src/schema/attributes/self-sovereignty/permissions-management.ts
maykelxyz: hmm yeah, that makes sense. It was a design choice by the wallet team

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -39,59 +44,139 @@ function worstControl(...controls: SpendingApprovalsControl[]): SpendingApproval
...
+		case BuiltInSwapDefaultApprovalBehavior.EXACT_AMOUNT:
+			return Rating.PASS
+		case BuiltInSwapDefaultApprovalBehavior.UNLIMITED_BUT_EDITABLE:
+		case BuiltInSwapDefaultApprovalBehavior.UNLIMITED_BUT_DISCLOSED:
+		case BuiltInSwapDefaultApprovalBehavior.UNLIMITED_AND_UNDISCLOSED:
+			return Rating.FAIL
+	}
+}
+
+function evaluate(ctx: EvaluationContext, control: PermissionsManagementSupport): Evaluation {
```
polymutex: Control flow has become confusing here. Matching on `overallRating` for the final branches hides the actual decision logic.

In other words, compared to:

```
const hasFoo = ...
const hasBar = ...
const hasMoreStuff = ...
const overallRating = ... long set of boolean conditions based on the above...
if (overallRating === good) {
  if(!hasBar) { return ... }
  return ...
}
if (overallRating === meh) {
  return ...
}
return ...
```

It's much easier to follow:

```
const hasFoo = ...
const hasBar = ...
if (hasFoo && !hasBar) {
  return good evaluation ...
}
if (hasFoo && hasBar) {
  return ...
}
```

In other words, make the rating enum be a part of the evaluation object verbatim, not part of the pivoting condition. The net effect is that the `if` conditions and the overall control flow are much easier to understand. The previous control flow followed this pattern more closely.

I think the temptation is coming from having refactored some of the logic out into `approvalsManagementRating` and `ratingForSwapApprovals` helpers that return `ExplicitRating`s. Nothing wrong with helpers, but you can have the cake and eat it too by having these helpers return a full `Evaluation` object instead. That also has the advantage of keeping the other non-rating fields (especially `howToImprove`) more closely relevant to the thing that's actually causing the rating to be dinged, rather than generic text explaining everything that _could_ be wrong. Then in the higher-level `evaluate` function, you can simply call `pickWorstRating` over what the helpers returned.

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -156,48 +241,77 @@ export const permissionsManagement: Attribute = {
...
 		result across all token standards determines the overall rating.
-		
-		As Account Abstraction becomes more prevalent, this methodology 
+
+		Wallets that offer a built-in swap or bridge feature are also evaluated on
+		whether that feature requests an exact-amount approval by default. Only an
+		exact-amount default passes. A default of unlimited fails this attribute,
+		regardless of disclosure or whether the user can edit the amount down
+		before signing, even if the wallet otherwise supports inspecting and
+		revoking approvals well.
```
polymutex: This should be more conciliatory to allow for minor over-approvals, needed to cover possible slippage in swaps.

src/schema/features/self-sovereignty/permissions-management.ts
```diff
@@ -29,4 +31,66 @@ export interface PermissionsManagement {
...
 
+/**
+ * How a wallet's own built-in swap/bridge feature requests token approvals
+ * on the user's behalf by default.
+ */
+export enum BuiltInSwapDefaultApprovalBehavior {
+	/**
+	 * The wallet requests only the amount needed for the swap/bridge before signing.
+	 */
+	EXACT_AMOUNT,
```
polymutex: I think this should be renamed to `MINIMAL_AMOUNT`, to reflect the case that in practice, it's OK to request a larger-than-exact-but-still-minimal amount to account for swap slippage.

polymutex (changes requested):

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -55,169 +54,201 @@ function approvalsManagementRating(control: SpendingApprovalsControl): ExplicitR
...
+ * Evaluates the wallet's built-in swap/bridge approval defaults on their own.
+ * Returns `null` when there is nothing to complain about (no built-in swap,
+ * or a minimal-amount-by-default swap that covers roughly what the swap
+ * needs, including a reasonable buffer for slippage): any unlimited default
+ * fails, whether disclosed or not.
+ */
+function swapApprovalsEvaluation(
+	ctx: EvaluationContext,
+	builtInSwapApprovals: BuiltInSwapDefaultApprovalBehavior | 'NO_BUILT_IN_SWAP',
+): Evaluation | null {
```
polymutex, line 67: Ambiguity in return type: "unrated"/not-applicable case could now be either encoded inside `Evaluation` or in the `null` case. Pick one, either by constraining `Evaluation` to `ExplicitRating`, or getting rid of `null`.

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -55,169 +54,201 @@ function approvalsManagementRating(control: SpendingApprovalsControl): ExplicitR
...
-	)
+	const undisclosed =
+		builtInSwapApprovals === BuiltInSwapDefaultApprovalBehavior.UNLIMITED_AND_UNDISCLOSED
 
-	if (walletHasBuiltInSwap && swapRating === Rating.FAIL) {
-		const undisclosed =
-			builtInSwapApprovals === BuiltInSwapDefaultApprovalBehavior.UNLIMITED_AND_UNDISCLOSED
+	return ctx.build({
+		outcome: {
+			id: undisclosed ? 'undisclosed_unlimited_swap_approval' : 'disclosed_unlimited_swap_approval',
```
polymutex, line 80: Too many fields depending on `undisclosed`. Use two separate `return` statements with `if (undisclosed)`

src/schema/attributes/self-sovereignty/permissions-management.ts
```diff
@@ -55,169 +54,201 @@ function approvalsManagementRating(control: SpendingApprovalsControl): ExplicitR
...
+			})
+	}
+}
+
+function evaluate(ctx: EvaluationContext, control: PermissionsManagementSupport): Evaluation {
+	const { approvalsManagement, builtInSwapApprovals } = control
+
+	const swapEvaluation = swapApprovalsEvaluation(ctx, builtInSwapApprovals)
+	const passingSwapDetails =
+		swapEvaluation === null && hasBuiltInSwap(builtInSwapApprovals)
```
polymutex, line 238: Dead branch

polymutex (changes requested):

polymutex (approved):

PR #1190, opened 2026-09-09
Wallet page: replace attribute petals with rating-colored icons

### LINKED ISSUE #1026: Wallet page: Replace use of petals for attributes on table of contents and body
The petals are nice but visually heavy. Replace them with the attribute icon itself, perhaps colorized to a rating-specific color, and/or with a rating-specific glow/drop-shadow.

This applies to petals in:
- The right sidebar
- The main body's attribute box headers
- The stage details, for stage criteria that map to attributes (petals aren't used there, but whatever they end up using, it should be the same as what we use to represent attributes in the aforementioned 2 places)

lucilapastore:
## Summary
- Replace the teardrop petals in the wallet-page table of contents and attribute box headers with the attribute icon itself.
- Color each icon with the rating/score accent and add a matching glow so stage and rating remain visible without the heavy petal shape.
- Leave the flower-chart petals in the right-sidebar pie navigation unchanged.

Closes #1026

### DIFF
--- a/src/views/WalletPage.svelte
+++ b/src/views/WalletPage.svelte
@@ -303,19 +303,6 @@
 		new Map(pieRotationSteps.map(step => [step.href, step.timeline]))
 	)
 
-	const attributeSliceStyles = $derived(
-		new Map(
-			pieNavigationItems.flatMap(group => (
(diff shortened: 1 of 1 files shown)


polymutex (approved):
Looks pretty awesome IMO:

Image

PR #1193, opened 2026-09-09
Attribute/duress resistance conditions

### LINKED ISSUE #1161: Duress Resistance: Should basic unlock be required?
I've been testing few wallets namely Phantom, Uniswap, and one common behavior they do is to not required basic unlocks like password, pin, etc. They __DO NOT HAVE__ basic text passwords, only biometrics (Face ID). Meaning if you leave your phone open, one can send funds if you do not have biometrics security turned on

maykelxyz:
Attribute: Require non-biometric unlock for duress resistance

## Changes

- Added `BasicUnlockMechanismSupport` enum (`OPTIONAL` / `REQUIRED`) to replace the old boolean `mechanisms` map, so a wallet can distinguish "supports PIN" from "requires PIN":

```ts
export enum BasicUnlockMechanismSupport {
	OPTIONAL = 'OPTIONAL',
	REQUIRED = 'REQUIRED',
}

export interface BasicUnlock {
	mechanisms: Record<BasicUnlockMechanism, BasicUnlockMechanismSupport | 'NOT_SUPPORTED'>
}
```

- A biometric-only or merely-optional lock is no longer duress-resistant. An attacker can force a fingerprint/face scan, and devices without biometric hardware can end up with no lock at all. `duress-resistance.ts` now requires at least one of PIN, password, or pattern to be `REQUIRED`, otherwise it fails with a new `weak_lock_only` outcome.

| Rating | Condition |
| --- | --- |
| **Pass** | Has a required PIN, password, or pattern lock, plus a working duress mode |
| **Partial** | Has a required PIN, password, or pattern lock, no duress mode |
| **Fail** | No lock screen, or only biometric/optional mechanisms with no required PIN/password/pattern |

## Other changes

- Updated wallets' duress resistance
- Regenerated `features.md`

Closes #1161

### DIFF
--- a/src/schema/attributes/security/duress-resistance.ts
+++ b/src/schema/attributes/security/duress-resistance.ts
@@ -10,6 +10,7 @@ import {
 	type BasicUnlock,
 	BasicUnlockMechanism,
 	basicUnlockMechanismName,
+	BasicUnlockMechanismSupport,
 	DuressAction,
 	duressActionDescription,
 	duressActionName,
@@ -87,7 +88,7 @@ function weakLockOnly(ctx: EvaluationContext, basicUnlock: WithRef<BasicUnlock>)
 function basicLockOnly(ctx: EvaluationContext, basicUnlock: WithRef<BasicUnlock>): Evaluation {
 	const mechNames = commaListFormat(
 		Object.keys(basicUnlock.mechanisms)
-			.filter(m => basicUnlock.mechanisms[m])
+			.filter(m => basicUnlock.mechanisms[m] !== 'NOT_SUPPORTED')
 			.map(basicUnlockMechanismName),
 	)
 
@@ -124,7 +125,7 @@ function hasDuressMode(
 ): Evaluation {
 	const mechNames = commaListFormat(
 		Object.keys(basicUnlock.mechanisms)
-			.filter(m => basicUnlock.mechanisms[m])
+			.filter(m => basicUnlock.mechanisms[m] !== 'NOT_SUPPORTED')
 			.map(basicUnlockMechanismName),
 	)
 	const activeActions = Object.keys(duressMode.actions).filter(a => duressMode.actions[a])
@@ -173,9 +174,10 @@ export const duressResistance: Attribute = {
 
 		Wallets can mitigate this threat by:
 
-		1. **Lock screen (basic)**: Requiring a PIN, password, or biometric before granting access.
-		   This protects against opportunistic thieves and buys time, but a determined coercer can
-		   watch you unlock it.
+		1. **Lock screen (basic)**: Requiring a PIN, password, or pattern before granting access.
+		   Biometrics (Face ID, fingerprint) may be offered as an optional convenience, but must
+		   not be the only option: an attacker can force a fingerprint or face scan, and devices
+		   without biometric hardware would otherwise be left with no lock at all.
 
 		2. **Duress mode (stronger)**: A separate duress PIN or passphrase that, when entered,
 		   either opens a decoy wallet (providing plausible deniability) or wipes the device and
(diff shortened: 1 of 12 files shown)


src/schema/features/security/duress-resistance.ts
```diff
@@ -34,15 +34,27 @@ export function basicUnlockMechanismName(m: BasicUnlockMechanism): string {
...
  */
 export interface BasicUnlock {
 	/**
-	 * Which unlock mechanisms the wallet supports.
-	 * Set each mechanism to `true` if supported, `false` if not.
+	 * Which unlock mechanisms the wallet supports, and whether each one is
+	 * required or merely optional.
 	 */
-	mechanisms: Record<BasicUnlockMechanism, boolean>
+	mechanisms: Record<BasicUnlockMechanism, BasicUnlockMechanismSupport | 'NOT_SUPPORTED'>
```
polymutex: This is an improvement over the previous type, but still seems inconsistent with how we treat "either unsupported, or supported with additional metadata" for other feature types. I think this should be `Support<{ type: 'OPTIONAL' | 'REQUIRED' }>` or similar (e.g. the `'OPTIONAL' | 'REQUIRED'` can remain a real `enum`).

This would also add the possibility of adding a `ref` here (`Support<WithRef<...>>`), might be useful to be able to attach a screenshot ref of the option in case it's optional.

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -41,10 +42,53 @@ function noLockScreen(ctx: EvaluationContext): Evaluation {
 	})
 }
 
+/**
+ * A biometric-only (or merely optional) lock is not duress-resistant: an
+ * attacker can force a fingerprint/face scan, and devices without biometric
+ * hardware may end up with no lock at all. At least one of PIN, password, or
+ * pattern must be REQUIRED for the lock screen to count.
```
polymutex: It's not so much that an attacker can force a scan, because an attacker can force a PIN/password/pattern entry too. What matters is that biometrics don't leave room for plausible deniability.
When an attacker forces a PIN/password/pattern entry, they cannot prevent the user from using a decoy code that unlocks an alternative wallet or self-destructs the wallet. They also cannot distinguish between a user legitimately having forgotten their PIN/password/pattern, vs a user that only pretends to have forgotten.

I'm not saying that the attribute should specifically require decoy unlock support; that's for stage 2 duress resistance. But what's important for lower-stage wallets is that they still allow non-decoy PIN/password/pattern unlock, because that ensures such entry methods remain normalized across wallets, rather than being exclusively associated with decoy-code-supporting wallets alone. Because if that association ever gets too strong, then plausible deniability starts to vanish for the wallets that do support decoy codes.

It's probably worth adding some of the above in the rationale for the attribute.

polymutex (changes requested):

src/schema/attributes/security/duress-resistance.ts
```diff
@@ -130,9 +180,14 @@ export const duressResistance: Attribute = {
...
-		   This protects against opportunistic thieves and buys time, but a determined coercer can
-		   watch you unlock it.
+		1. **Lock screen (basic)**: Requiring a PIN, password, or pattern before granting access.
+		   Biometrics (Face ID, fingerprint) may be offered as an optional convenience, but must
+		   not be the only option, since they offer no plausible deniability. A forced
+		   PIN/password/pattern entry leaves room for a decoy code that unlocks a decoy wallet or
+		   wipes the device, indistinguishable to the attacker from a genuine entry, and an
+		   attacker cannot tell a user who has genuinely forgotten their credential from one who is
+		   only pretending to. Devices without biometric hardware would otherwise be left with no
+		   lock at all.
```
polymutex, line 190: I don't understand what this last sentence is saying, but either way I don't think it's necessary.

polymutex (changes requested):
Please get the grammar check passing

polymutex (approved):

Issue #737, opened 2026-05-16
Import data from Coinspect: Credit Coinspect for its data in the UI (3/4)


polymutex:
Per #215, Coinspect is now sharing its data under a permissive license: https://github.com/coinspect/wallet-security-ranking

A lot of what Coinspect looks for in a wallet is also what Walletbeat looks for (and that's not surprising, since we both share the objective of pushing wallets on security). To that end, it would great to import Coinspect's data to enrich Walletbeat's own dataset. (Sharing data in the reverse direction is already implemented in #167).

There are four steps to making this work:

1. Creating a mapping from Coinspect wallet IDs to Walletbeat's wallet IDs (#735)
2. Importing Coinspect's raw data and ensuring automatic syncs (#736)
3. Adding UI code to credit Coinspect (and link back to it), on attribute boxes where the data used to evaluate the attribute comes from Coinspect's dataset (this issue)
4. Creating data transformation functions to import Coinspect's data into Walletbeat's own data structures (#738) (We should not implement this until the UI code for crediting Coinspect exists.)

This issue tracks (3): Crediting Coinspect for its data in the Walletbeat UI.

This includes:
- Having a way for Walletbeat's wallet data to have subsets of it attributed to Coinspect somehow. (Or better, having a generic data attribution mechanism, of which Coinspect would be one possible instantiation of.) This could hook into the `ref` system, perhaps expanding `WithRef` to have an optional `attribution` enum field, or perhaps having its `ref` field take on a special value for Coinspect data.
- Adding code to *all attributes* that consume data possibly attributed to Coinspect that tracks whether the data they rely on *does* come from Coinspect, and output that as part of their evaluation output. This should be relatively easy with the `EvaluationContext` object, and integration in the `WithRef` type makes this straightforward because we already have plumbed all such "feature data derivedness" through that system already.
- Adding code to the per-wallet page on the per-attribute boxes, which links to Coinspect if the attribute's evaluation process relied on any data that was imported from Coinspect.

PR #1195, opened 2026-09-09
Preserve citation evidence on unsupported wallet features

johngrantuk:
## Summary
- Keep `notSupportedWithRef` evidence on displayed PASS/PARTIAL/FAIL claims instead of dropping it during evaluation. This is required for #737.
- Leave supported-ref collection, variant aggregation, and unrelated nested refs unchanged; skip unsupported refs on generic UNRATED/EXEMPT outcomes and on fee-transparency capability absences that do not evidence fee display.
- Render scam-alert citations inline on the wallet page whenever a warning has refs, independent of support status, and keep those refs in `evaluation.references` for markdown/JSON consumers.

## Rainbow visual check

Manual acceptance fixture for unsupported-feature citations. Open Rainbow and confirm absence evidence is shown on the selected FAIL/PARTIAL claims, and that existing supported citations still appear.

- [ ] **Browser** — [`/rainbow?variant=BROWSER`](http://localhost:4321/rainbow?variant=BROWSER)
  - **App isolation** cites the connect dialog with no create-account option: `2026-06-18-app-isolation-connect-wallet-picker.png`
  - Existing **supported** citations still render (e.g. `eth_accounts` / the same picker screenshots)
- [ ] **Mobile** — [`/rainbow?variant=MOBILE`](http://localhost:4321/rainbow?variant=MOBILE)
  - **Scam prevention** shows absence citations **inline** under the relevant warning rows:
    - `2026-07-26-scam-alerts-flagged-app-send-no-warning.png` (no send warning)
    - `2026-07-18-scam-alerts-unlimited-approval-neutral-label.png` (unlimited approval not treated as a warning)
  - **Transaction batching** cites `2026-08-20-wallet-call-mobile-batch-unsupported.png`


tests/evaluation-context-references.test.ts
polymutex, line 1: I don't think this test is useful or scalable, and suggest deleting it. It catches that the current batch of attributes is doing the right thing, which is useful to catch regressions, but doesn't catch what this PR might have already missed, and it also won't catch further changes if we re-jigger the feature types that these attributes look at.

That said, a more generic test that automatically detects which feature fields are used by which attributes, and systematically verifies that they are passed through (without encoding attribute-specific logic) would be very useful and catch all three classes of issues simultaneously.

For example, it could iterate through all wallets' attribute ratings, and systematically change all `ref` fields to some well-defined URL pattern that it can recognize in the evaluation results. This lets it figure out which `ref`s are accessed by that attribute for this wallet. Then, for that attribute and for all other wallets, it can then verify that if that other wallets' data has the same `ref` field populated, and the rating is non-`EXEMPT` and non-`UNRATED`, then the same `ref` field must also be preserved.

Maybe this is a good idea for a separate PR.

tests/unsupported-reference-attributes.test.ts
polymutex, line 1: Same comment here. (And if we did want to keep a test on this, it doesn't make much sense to split it in two files along these lines)

polymutex (changes requested):
Good idea but I don't think the current testing is the right approach. OK to submit as-is whether you plan to address the testing in a separate PR or this one.

polymutex (approved):

tests/evaluation-context-references.test.ts
johngrantuk: Makes sense. I have opened #1199 to be added as a follow up.

polymutex (approved):
Still got merge conflict, perhaps you merged an older commit of `beta`?

Issue #1178, opened 2026-09-06
Get `svelte-check` clean and gate CI on it


MattPereira:
Progress tracker to fulfill #1115.

## Progress

- [x] Cluster 1: Svelte generic declarations — 13 errors (#1186)
- [x] Cluster 2: TypeScript inference edges — 10 errors
- [x] Cluster 3: Control-flow narrowing and nullability — 14 errors
- [x] Cluster 4: UI component contract drift — 7 errors
- [ ] Cluster 5: Wallet and evaluation model mismatches — 10 errors
- [ ] Cluster 6: Web-platform and tooling compatibility — 4 errors
- [ ] Decide whether 26 warnings must also be resolved
- [ ] Add `svelte-check` to CI

The baseline is 58 errors: 13 addressed by #1186 plus 45 remaining across clusters 2–6.

## Suggested sequencing

| Order | Cluster | Errors | Effort |
|---|---|---:|---|
| 1 | Svelte generic declarations | 13 | Complete in #1186 |
| 2 | TypeScript inference edges | 10 | Low–medium; four patterns |
| 3 | Control-flow narrowing and nullability | 14 | Low; mostly localized |
| 4 | UI component contract drift | 7 | Medium; confirm intended APIs |
| 5 | Wallet and evaluation model mismatches | 10 | High; needs type-model decisions |
| 6 | Web-platform and tooling compatibility | 4 | Mixed; two errors blocked upstream |

`WalletTable.svelte` and `WalletPage.svelte` account for most errors. Cluster 5 carries most design risk; cluster 6 cannot fully close without an upstream fix or suppression.

<details>
<summary><b>Detailed cluster breakdown</b></summary>

### 1. Svelte generic declarations — 13 errors

#1186 addresses all 13 Svelte `generics=` identity errors.

### 2. TypeScript inference edges — 10 errors

- `ConcreteWalletEvalStrings` rejected by `Typography` generic inference: 6
- `StageCriterionRating` rejected when indexing `stageCriterionRatings`: 2
- `BlockTransition` transition function/parameter tuple incompatibility: 1
- Generic `WalletTable` value rejected as an image `alt`: 1

### 3. Control-flow narrowing and nullability — 14 errors

- `false | tag | undefined` is not narrowed in the `WalletTable.svelte` tag loop: 10
- `slice.children` and `slice.gradient` remain possibly undefined after a guard in `Pie.svelte`: 2
- `walletConnectState.result` is inferred as only `null`: 1
[...]

PR #1196, opened 2026-09-09
Fix clusters 2 & 3 for `svelte-check`

MattPereira:
Part of #1178.

- Fix TypeScript inference edges
- Fix control-flow narrowing and nullability
- Reduce `check:svelte` errors from 45 to 21

### DIFF
--- a/src/components/BlockTransition.svelte
+++ b/src/components/BlockTransition.svelte
@@ -7,8 +7,8 @@
 	import type { TransitionConfig } from 'svelte/transition'
 
 	type TransitionFn = {
-		bivarianceHack(node: Element, params?: unknown): TransitionConfig
-	}['bivarianceHack']
+		transition(node: Element, params?: unknown): TransitionConfig
+	}['transition']
 
 	type TransitionFnAndParams = [TransitionFn] | [TransitionFn, unknown]
 
--- a/src/components/BlockTransition.svelte
+++ b/src/components/BlockTransition.svelte
@@ -4,6 +4,13 @@
 ">
 	// Types
 	import type { Snippet } from 'svelte'
+	import type { TransitionConfig } from 'svelte/transition'
+
+	type TransitionFn = {
+		bivarianceHack(node: Element, params?: unknown): TransitionConfig
+	}['bivarianceHack']
+
+	type TransitionFnAndParams = [TransitionFn] | [TransitionFn, unknown]
 
 
 	// Props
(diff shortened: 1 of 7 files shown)


src/schema/attributes.ts
```diff
@@ -23,7 +23,7 @@ export type WalletNameAndPseudonymStrings =
  * Used as return type of getWalletEvalStrings so it's assignable to Typography's
  * strings prop when content uses WalletNameAndPseudonymStrings (which is a union including null).
  */
-export interface ConcreteWalletEvalStrings {
+export type ConcreteWalletEvalStrings = {
```
MattPereira, line 26: Typography requires a string index signature, which TypeScript infers for closed type aliases but not declaration-mergeable interfaces.

src/components/BlockTransition.svelte
```diff
@@ -6,13 +6,11 @@
...
-	> = (
-		Fn extends (node: Element, _?: infer Params) => TransitionConfig
-			? [Fn] | [Fn, Params | undefined]
-			: never
-	)
+	type TransitionFn = {
+		bivarianceHack(node: Element, params?: unknown): TransitionConfig
+	}['bivarianceHack']
+
+	type TransitionFnAndParams = [TransitionFn] | [TransitionFn, unknown]
```
MattPereira, line 13: Transition functions accept transition-specific parameter types, so a bivariant callback type preserves configurability without requiring each function to accept arbitrary unknown parameters.

src/schema/stages.ts
```diff
@@ -95,7 +95,7 @@ export type StageCriterionEvaluation = {
 } & (
 	| {
 			// If not unrated, the explanation must be provided.
-			rating: Omit<StageCriterionRating, StageCriterionRating.UNRATED>
+			rating: Exclude<StageCriterionRating, StageCriterionRating.UNRATED>
```
MattPereira, line 98: Exclude removes the UNRATED enum member from the union, while Omit operates on object properties and loses the usable enum-key type.

src/views/WalletTable.svelte
```diff
@@ -825,7 +825,7 @@
 						</div>
 					{/if}
 				{:else if column.id === 'displayName'}
-					{@const displayName = value}
+					{@const displayName = wallet.metadata.displayName}
```
MattPereira, line 828: Reading displayName directly from wallet metadata preserves its string type, while generic column value remains a union and cannot safely populate image alt.

src/schema/stages.ts
polymutex: lol whoops. How did this even work...

polymutex (approved):

PR #1200, opened 2026-09-10
737 coinspect credit

### LINKED ISSUE #737: Import data from Coinspect: Credit Coinspect for its data in the UI (3/4)

johngrantuk:
## Summary 

Closes [#737](https://github.com/walletbeat/walletbeat/issues/737). #1195 is related and should be merged to ensure all credits shown.

- Credit Coinspect by tagging the **reference**, not the wallet or the attribute.
- `coinspectRef()` builds a normal citation (URL, note, date) and marks it as coming from Coinspect. Existing reference plumbing stays the same, so attributes do not need new code.
- The wallet page and JSON export look for tagged references and show attribution. Ordinary Walletbeat citations stay untagged and look the same as before.
- Adds credit line in the markdown export. Extends `walletPageMarkdown` (and `tests/wallet-page-markdown.test.ts`).

The Coinspect importer itself stays in #738; this PR does not add any production `/data`.


src/schema/reference.ts
polymutex, line 442: Doesn't at least this function need awareness of the `source` field?

data/entities/coinspect.ts
```diff
@@ -0,0 +1,27 @@
...
+	name: 'Coinspect',
+	legalName: { name: 'Coinspect Security LLC', soundsDifferent: false },
+	type: {
+		chainDataProvider: false,
+		corporate: true,
+		dataBroker: false,
+		exchange: false,
+		infrastructureProvider: false,
+		offchainDataProvider: false,
+		securityAuditor: false,
```
polymutex: They are

src/utils/data-source-credits.ts
polymutex, line 1: How about moving this into `reference.ts`?

tests/coinspect-ref.test.ts
```diff
@@ -0,0 +1,72 @@
...
+				label: `Coinspect ${CHECK}`,
+			},
+		])
+		expect(ref.explanation).toBe(NOTE)
+		expect(ref.lastRetrieved).toBe('2026-01-12')
+		expect(ref.source).toBe(coinspectDataSource)
+		expect(COMMIT.endsWith('\n')).toBe(false)
+		expect(COMMIT).toHaveLength(40)
+	})
+})
```
polymutex: Change-detector test, not useful

tests/coinspect-ref.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+describe('coinspectRef through rateWallet', () => {
```
polymutex: Not convinced this test is very useful either

src/schema/data-sources/index.ts
polymutex, line 1: I don't think we'll have that many data sources that we need a whole multi-file package for it :)
(This file and `coinspect.ts` can just be a single `src/schema/data-sources.ts` file)

polymutex (changes requested):
Can you screenshot what this looks like for a sample wallet?

src/schema/reference.ts
johngrantuk: Made mergeRefs source-aware:

- isRefEqual now also compares a.source?.entity.id !== b.source?.entity.id, so the pre-merge dedup no longer collapses a stamped ref into an unstamped twin.
- Explanation-merge map is now keyed on `${ref.source?.entity.id ?? ''}\u0000${ref.explanation}` instead of just the explanation. Refs sharing a note but coming from different data sources (or stamped vs. unstamped) stay separate.
- Merged object preserves source: the rebuilt ref now spreads in source: existing.source ?? ref.source when either side has one. Since the key already partitions by source identity, this can never fuse two different datasets.

data/entities/coinspect.ts
```diff
@@ -0,0 +1,27 @@
+import type { CorporateEntity } from '@/schema/entity'
+
+export const coinspect: CorporateEntity = {
```
polymutex: `CorporateEntity & SecurityAuditor`

src/schema/reference.ts
```diff
@@ -450,7 +465,10 @@ export function mergeRefs(
 		}
 	}
 
-	const byExplanation = new Map<string, FullyQualifiedReference>()
+	// Merge refs that share both an explanation and a data-source identity.
+	// Refs with the same explanation but different sources (including stamped
+	// vs. unstamped) stay separate so that per-source attribution survives.
+	const byExplanationAndSource = new Map<string, FullyQualifiedReference>()
```
polymutex: IMO, if explanations match but data source doesn't, and one leg doesn't have a data source, then they should still get merged but the merged version has the data source.
If both legs have a data source and they are the same data source, that works too.
If both legs have a data source and they are not the same, then throw an error because we probably have a bigger problem somewhere.

src/views/DataSourceCredits.svelte
polymutex, line 1: I think this looks a bit too wordy for something we probably expect to be repeated in many attributes across many pages. It makes sense in the markdown version of the page because there's not that many ways to represent this, but for the full-HTML version of the site I think we can do better.

I'm thinking this section could appear exactly once at the bottom of the page, not per-attribute, i.e. the data credits are merged and deduplicated across all references of all attributes, and if there's anything, then the section appears at the bottom. Then, for any reference link in the attribute sections, if there is a data source, then add a `[Credit: Coinspect]` link to the explanation link, which scrolls to the data credits section of the page. That'll cut down on the visual weight of the page while actually putting the attribution closer to the reference itself (rather than the current way where it's a bit separated from the attribute references section altogether).

Maybe we could do the same on the markdown version too, using `#anchor` links as a way to indicate "what you're looking for is in a different section"...

src/views/DataSourceCredits.svelte
```diff
@@ -0,0 +1,102 @@
...
+						<a
+							href={getUrl(credit.source.license.url)}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<cite>{credit.source.license.name}</cite>
+							<span>{@html ExternalLinkIcon}</span>
+						</a>
+					</p>
+
```
polymutex: Per screenshot, this feels a bit too visually sparse. It's got 5 separate lines:
- "Data credit" section header
- Data source header
- Reference
- Data license
- Attribution text

Feels like we could merge this down to two:

- "Data credit" section header
- Data source: Reference, data license, attribution text

src/views/DataSourceCredits.svelte
```diff
@@ -0,0 +1,102 @@
...
+			{#each credits as credit (credit.source.entity.id)}
+				<li data-list-item="gap-2">
+					{#if isUrl(credit.source.entity.url)}
+						<a
+							href={getUrl(credit.source.entity.url)}
+							target="_blank"
+							rel="noopener noreferrer"
+						>
+							<cite>{credit.source.entity.name}</cite>
+						</a>
```
polymutex: Perhaps we could have the a small (icon-sized) logo too here?

polymutex (changes requested):

src/views/DataSourceCredits.svelte
johngrantuk: Makes sense. 

Source2
DataCredit2

src/utils/wallet-page-markdown.ts
```diff
@@ -272,6 +278,16 @@ export function walletPageMarkdown<_AttributeGroupId extends string>(
 						parts.push('#### References', '')
 
 						for (const ref of qualifiedRefs) {
+							// Inline data-credit link points to the consolidated
+							// "Data credits" section at the bottom of the page.
+							// Stamped references share this suffix so attribution
+							// stays close to the citation without repeating the
+							// full credit block per attribute.
```
polymutex, line 285: Temporal comment, please remove

src/views/DataSourceCredits.svelte
polymutex: Missing a space between the comma and before "CC", and another between "." and "Adapted". I think it may need `&nbsp;` or something.

src/views/ReferenceLinks.svelte
```diff
@@ -165,9 +165,30 @@
...
+									Last retrieved <time datetime={ref.lastRetrieved}>{ref.lastRetrieved}</time>
+								</span>
+							{/if}
+							{#if ref.source}
+								{@const source = ref.source}
+								<a
+									class="ref-credit"
+									href={`#${dataCreditAnchorId(source.entity)}`}
+									title={`Data source: ${source.entity.name} — see credits at bottom of page`}
+									aria-label={`Data source: ${source.entity.name} — see credits at bottom of page`}
```
polymutex, line 181: Sorry but Imma fight against these emdashes for the foreseeable future. They scream "this website is vibe-coded"

src/views/WalletPage.svelte
```diff
@@ -337,6 +337,33 @@
 		calculateOverallScore(attributeTree, wallet.overall, () => true),
 	)
 
+	// All fully-qualified references cited across every non-exempt attribute on
+	// this page. Fed into a single `<DataSourceCredits>` at the bottom of the
+	// page so stamped-source attribution is shown once (deduplicated by
+	// source) rather than repeated per attribute; inline `[Credit: …]`
+	// markers on each reference link scroll to the corresponding entry.
```
polymutex, line 344: Temporal

tests/wallet-page-markdown.test.ts
```diff
@@ -148,34 +153,75 @@ describe('walletPageMarkdown', () => {
...
 					const section = attributeSection(md, heading)
-					const credits = computeDataSourceCredits(toFullyQualified(evalAttr.evaluation.references))
 
-					if (credits.length === 0) {
-						expect(section).not.toMatch(/^#### Data credits?$/m)
-						continue
-					}
+					// Per-attribute Data credit(s) subsections were consolidated
+					// into a single page-level section; attribute sections must
+					// no longer emit their own credit heading.
```
polymutex, line 159: Temporal

In fact, the whole test is temporal and not really useful

polymutex (changes requested):

johngrantuk:
Updated screenshot with correct spacing:

Screenshot 2026-09-16 at 11 18 39

polymutex (approved):

polymutex:
Looks good, please rebase

PR #1202, opened 2026-09-10
Speed up Astro checks by simplifying CalendarDate

MattPereira:
## Issue

astro and ts checks were so slow because:

> Template literal types made of finite unions get eagerly expanded by TS into an actual union of every literal:
> 
> - MonthAndDay = 31×7 + 30×4 + 29 = 366 members
> - year = Century(2) × Digit(10) × Digit(10) = 200
> - → 73,200 string literal members
> 
> That union is materialized once, then every one of the ~59 CalendarDate references — plus every date literal in wallet/news/audit data files — does assignability checks against it. Each check is a union-membership scan, and the union gets re-instantiated inside generics (schema types, Ref, etc.). Cost is superlinear in practice, which is why it dominated the check phase.

## Summary

- loosen `CalendarDate` type to reduce time required for checks
  - `astro check`: 2m09s → 16.5s 
  - TypeScript check phase: 103s → 5.1s 
- test moves the year check the type gave up from compile time to runtime, validating every date in the data and every date literal in source.

### DIFF
--- a/src/types/date.ts
+++ b/src/types/date.ts
@@ -22,7 +22,7 @@ export type CalendarDate = `${Century}${number}-${MonthAndDay}`
 const CALENDAR_DATE_PATTERN = /^(20|21)\d{2}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])$/
 
 /** Check whether a string is a valid CalendarDate. */
-function isCalendarDate(date: string): date is CalendarDate {
+export function isCalendarDate(date: string): date is CalendarDate {
 	if (!CALENDAR_DATE_PATTERN.test(date)) {
 		return false
 	}
--- a/src/types/date.ts
+++ b/src/types/date.ts
@@ -10,8 +10,14 @@ type MonthAndDay =
 	| `02-${FebruaryDays}`
 type Century = '20' | '21' // We're good from 2000 to 2199.
 
-/** A valid date in YYYY-MM-DD format. */
-export type CalendarDate = `${Century}${Digit}${Digit}-${MonthAndDay}`
+/**
+ * A calendar date with a 20xx/21xx year prefix and valid MM-DD shape.
+ *
+ * Using `${number}` for the year suffix avoids expanding this type into every possible date,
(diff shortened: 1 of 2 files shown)


src/types/date.ts
```diff
@@ -10,8 +10,14 @@ type MonthAndDay =
 	| `02-${FebruaryDays}`
 type Century = '20' | '21' // We're good from 2000 to 2199.
 
-/** A valid date in YYYY-MM-DD format. */
-export type CalendarDate = `${Century}${Digit}${Digit}-${MonthAndDay}`
```
MattPereira: `${Century}${Digit}${Digit}-${MonthAndDay}`

  Expands to:

  2 centuries × 10 digits × 10 digits × 366 month-days
  = 73,200 literal types

  Conceptually:

  '2000-01-01' | '2000-01-02' | ... 73,198 more

src/types/date.ts
```diff
@@ -10,8 +10,14 @@ type MonthAndDay =
...
+/**
+ * A calendar date with a 20xx/21xx year prefix and valid MM-DD shape.
+ *
+ * Using `${number}` for the year suffix avoids expanding this type into every possible date,
+ * which makes TypeScript checking pathologically slow. Use `assertCalendarDate` when exact
+ * YYYY-MM-DD validation is required at runtime.
+ */
+export type CalendarDate = `${Century}${number}-${MonthAndDay}`
```
MattPereira: `${Century}${number}-${MonthAndDay}`

  ${number} is kept as a symbolic “numeric string” pattern—it is not expanded into every possible number. TypeScript handles only roughly:

  2 centuries × 366 month-days = 732 patterns

  This avoids repeatedly comparing objects against a 73,200-member union. In the trace, one such comparison consumed ~90 seconds.

  Astro uses TypeScript for semantic diagnostics, so fixing TypeScript’s workload fixes both checks. Tradeoff: ${number} accepts more year shapes;
  assertCalendarDate() enforces exact YYYY-MM-DD at runtime.

PR #1212, opened 2026-09-12
Fix Svelte wallet evaluation model mismatches

MattPereira:
Closes cluster 5 of #1178.

## Summary

- require wallet pages to consume complete `RatedWallet` values while keeping ladder-definition maps partial
- update stage helpers to handle partial ladder registries explicitly
- separate baked custom-detail component props from evaluation data injected during rendering
- validate erased outcome metadata at the heterogeneous `EvaluationTree` rendering boundary
- remove the unsafe component-props assertion and add focused adapter tests


polymutex (commented):
Any particular reason why this PR is marked as draft?

src/types/content/evaluation-details.ts
```diff
@@ -0,0 +1,154 @@
...
+): EvaluationDetailRenderData {
+	switch (details.component) {
+		case 'ScamAlertDetails':
+			return {
+				...details,
+				outcome: requireOutcomeMetadata(details.component, outcome, isScamPreventionMetadata),
+			}
+		case 'SecurityAuditsDetails':
+			return {
+				...details,
```
MattPereira, line 133: An adapter verifies that each detail component receives compatible outcome data before rendering, but this only catches mismatches at runtime and duplicates information that should be encoded in the type system.


Optional follow up issue to fix :point_down: 
- https://github.com/walletbeat/walletbeat/issues/1215

src/types/content/evaluation-details.ts
```diff
@@ -0,0 +1,154 @@
...
+	| (SecurityAuditsDetailsContent & { outcome: Outcome<SecurityAuditsMetadata> })
+	| (AccountRecoveryDetailsContent & { outcome: Outcome<AccountRecoveryMetadata> })
+	| (AccountUnruggabilityDetailsContent & {
+			outcome: Outcome<AccountUnruggabilityMetadata>
+	  })
+	| UnboundEvaluationDetailRenderData
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+	return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
```
polymutex, line 39: Needs to check that all keys are strings.

src/types/content/evaluation-details.ts
```diff
@@ -0,0 +1,154 @@
...
+
+function isAccountUnruggabilityMetadata(value: unknown): value is AccountUnruggabilityMetadata {
+	return (
+		isRecord(value) &&
+		Object.hasOwn(value, 'minimumGuardianPolicy') &&
+		Object.hasOwn(value, 'outcomes') &&
+		(value.minimumGuardianPolicy === null || isRecord(value.minimumGuardianPolicy)) &&
+		(value.outcomes === null || Array.isArray(value.outcomes))
+	)
+}
```
polymutex, line 93: How about putting these in their `*-details.ts`-specific file?

src/types/content/unrated-attribute.ts
```diff
@@ -1,18 +1,14 @@
...
 
-export interface UnratedAttributeProps<
-	_OutcomeMetadata extends OutcomeMetadata,
-> extends EvaluationData<_OutcomeMetadata> {}
-
-export interface UnratedAttributeContent<_OutcomeMetadata extends OutcomeMetadata> {
+export interface UnratedAttributeContent {
 	component: 'UnratedAttribute'
-	componentProps: UnratedAttributeProps<_OutcomeMetadata>
+	componentProps: Record<never, never>
```
polymutex, line 7: What is this type?

src/utils/stage-attributes.ts
```diff
@@ -171,19 +170,29 @@ export function getAttributeCriteriaForWallet<
...
+	ladders: Ladders<_AttributeGroupId>,
 	attribute: Attribute<_OutcomeMetadata>,
 	wallet: StageEvaluatableWallet<_AttributeGroupId>,
 ) {
 	return Object.entries(ladders)
-		.flatMap(([ladderType, ladder]) =>
-			ladder.stages.map((stage, stageIndex) => ({
+		.flatMap(entry => {
+			if (entry === undefined) {
+				return []
```
polymutex, line 180: Fail loudly

src/utils/stage-attributes.ts
```diff
@@ -171,19 +170,29 @@ export function getAttributeCriteriaForWallet<
...
+			}
+
+			const [ladderType, ladder] = entry
+
+			if (ladder === undefined) {
+				return []
```
polymutex, line 186: Fail loudly

tests/evaluation-details.test.ts
polymutex, line 1: Not convinced of the value of this test, unless it can perhaps be written in such a way that it will catch new types of attributes that were added to the codebase but not added to it.

polymutex (changes requested):

src/types/content/unrated-attribute.ts
MattPereira: A poor attempt at empty object `{}`

Changed to `Record<PropertyKey, never>` which is strictly `{}`

tests/evaluation-details.test.ts
MattPereira: New evaluation-detail types now fail type-checking until the test declares whether they require component-specific metadata validation. Required types need fixtures covering valid, missing, and incompatible metadata. This does not cover every attribute because attributes may reuse a renderer or use plain text.

also don't mind dumping test

PR #1214, opened 2026-09-12
Add referer-based request matchers, and transitive `NOT_WALLET_INITIATED` tagging via referer matching.

polymutex:
This (correctly) detects that the previously-debank-attributed collected data isn't actually collected by the extension during the `INSTALL` flow. This was caused by requests that were made by opening the Rabby download page, not by the extension itself.


polymutex:
Ping @maykelxyz. Will merge in a day or two if not reviewed...

maykelxyz (approved):
Looks good to me, I'll be using this tool again this week, should be helpful

PR #1216, opened 2026-09-13
Chore: Add CLAUDE.md importing AGENTS.md

MattPereira:
anthropic still refusing to adopt the standard :melting_face: 

BEFORE:
image

AFTER:
image


polymutex (changes requested):
Use a symlink instead?

polymutex (approved):

Issue #1199, opened 2026-09-10
Add feature/attribute tests


johngrantuk:
I don't think this test is useful or scalable, and suggest deleting it. It catches that the current batch of attributes is doing the right thing, which is useful to catch regressions, but doesn't catch what this PR might have already missed, and it also won't catch further changes if we re-jigger the feature types that these attributes look at.

That said, a more generic test that automatically detects which feature fields are used by which attributes, and systematically verifies that they are passed through (without encoding attribute-specific logic) would be very useful and catch all three classes of issues simultaneously.

For example, it could iterate through all wallets' attribute ratings, and systematically change all `ref` fields to some well-defined URL pattern that it can recognize in the evaluation results. This lets it figure out which `ref`s are accessed by that attribute for this wallet. Then, for that attribute and for all other wallets, it can then verify that if that other wallets' data has the same `ref` field populated, and the rating is non-`EXEMPT` and non-`UNRATED`, then the same `ref` field must also be preserved.

Maybe this is a good idea for a separate PR.

_Originally posted by @polymutex in https://github.com/walletbeat/walletbeat/pull/1195#discussion_r3975846128_

PR #1217, opened 2026-09-13
Test feature reference propagation

MattPereira:
Stab at #1199.

## Summary
data-driven invariant test for feature-reference propagation across all rated wallet variants.

- clones each variant's resolved features and replaces every populated `ref` with a unique marker URL
- re-evaluates the wallet and verifies marker substitution did not change the outcome ID or rating
- observes which markers reach each attribute evaluation's references
- compares wallets with the same attribute and outcome
- requires a reference path preserved by one eligible wallet to be preserved by every eligible wallet containing that path
- excludes `EXEMPT` and `UNRATED` evaluations
- documents one intentional account-portability exception

This catches inconsistent propagation without maintaining attribute-specific fixtures as wallet data grows.

## Limitations

This proves cross-wallet consistency, not that every feature reference used by an attribute is propagated:

- a reference omitted by every evaluation provides no positive example, so it remains undetected
- comparisons require at least two wallets with the populated path
- evaluations are grouped by outcome ID, so differences across outcome branches are not compared
- array indices are normalized to `[]`; preserving one element marks that path as preserved
- coverage depends on currently populated wallet data and executed branches
- intentional semantic differences still require explicit exceptions

## Potential Follow-up

A stronger guarantee requires tracking or controlling feature access with refactor:

1. Make raw `EvaluationContext.features` unavailable to attributes.
2. Add a typed accessor such as `ctx.use(selector)` for evidence-bearing feature values.
3. Have the accessor automatically add the selected value's references.
4. Enforce through types/lint that attributes cannot bypass the accessor.

A recursive `Proxy` could trace reads in tests, but would need semantics for exploratory reads, null checks, containers, and values inspected but irrelevant to the selected outcome. The accessor design makes reading evidence and propagating its reference one operation.


polymutex (approved):
Excellent :)

PR #1238, opened 2026-09-16
Display wallet-specific attribute notes

### LINKED ISSUE #1226: Display `override.[...].note` field on attributes
Long-forgotten field that the previous UI generation used to support. In #1159 we are using it for MetaMask. We should plumb it through to the UI (and markdown version, and JSON export, etc.)

MattPereira:
Closes #1226

the legacy UI used :point_right: emoji for notes 

image

image


src/views/WalletPage.svelte
```diff
@@ -1276,7 +1282,7 @@
 								strings={getWalletEvalStrings(wallet)}
 							/>
 
-							{#if override}
+							{#if override?.howToImprove !== undefined}
```
MattPereira, line 1287: needed to be more specific with the gate since multiple override fields now

polymutex (approved):
Such fast turnaround 👍

PR #1244, opened 2026-09-18
Implement wallet stage filtering

### LINKED ISSUE #504: Stage system: move not-stage-0 wallets to separate tab

lucilapastore:
## Summary
- Closes #504: wallets that fail Stage 0 (public source availability) cannot be independently verified for many attributes, so they are no longer ranked with Stage 0+ wallets.
- The comparison table’s Stage column becomes **Stage 0+**. **Others** is hidden until chosen from that column’s chevron, same pattern as Expand on Rating. There are no Stage 0+ / Others buttons above the table.
- Hardware and embedded tables are unchanged (no stage ladder).

## Test plan
- [ ] The Stage header reads **Stage 0+** and has a chevron like Rating’s Expand control.
- [ ] Opening the chevron shows Stage 0+ and Others; picking Others lists unranked wallets (Phantom, Zerion, …) and retitles the column to **Others**.
- [ ] On a narrow viewport, the same Stage 0+ / Others choice appears in the stages filters.
- [ ] `/hww/summary/` has no Stage 0+ / Others control.


polymutex:
Not a bad approach. Not easily discoverable, but that's not necessarily a bad thing for this particular feature...

Please delete the `tests/stage.test.ts` test though, it is useless.

@darrylyeo Thoughts?

lucilapastore:
> Not a bad approach. Not easily discoverable, but that's not necessarily a bad thing for this particular feature...
> 
> Please delete the `tests/stage.test.ts` test though, it is useless.
> 
> @darrylyeo Thoughts?

On it!

